PR #41, opened 2023-08-14
Avoid generating unneccessary conditionals

ewoutkramer:
The code generator generated some dead code, where an else branch could never be reached because the "if" was a constant boolean. To avoid compiler warnings (which are now errors), change the codegen to avoid generating these conditionals, replacing them with either the IfTrue or IfFalse expressions, depending on the fixed value.


EvanMachusak (approved):
Approved

PR #43, opened 2023-08-15
Remove dependency on AgileObjects

ewoutkramer:
After reading the C# code generator, it occurred to me that we do not need this dependency anymore by adding a few small tweaks. As I needed something small to get familiar with the C# code generator, I decided to see if I could cut it out.

As I was reviewing the code, I noticed some indentation bugs, and even a real one, which I fixed (walking the path on a nullable base object was not postfixed with the "?." operator).


EvanMachusak (approved):
Tests still pass

ewoutkramer:
Merging anyway, PR was approved, just had to update due to changes in develop.

PR #44, opened 2023-08-15
Avoid inefficient DataRetriever calls

ewoutkramer:
Because of an upstream tooling change, the ELM retrieve statement now always(?) has its 'code' property set, so we generated a lot of `GetProperty()` calls to get to the code property, even for defaults.

The ExpressionBuilder will now detect that the given code property is the default, and call the retriever methods with a null PropertyInfo argument again.

Furthermore, *if* there is a non-default property specified in the elm, we will generate the GetProperty only once, reusing the result in Retrieve calls.

Also included some smaller cleanups which do not change code generation.


Cql/Cql.Compiler/ExpressionBuilder.cs
```diff
@@ -2038,30 +2039,44 @@ protected bool IsInterval(Type t, out Type? elementType)
             return selectManyLambda;
         }
 
+        // Yeah, hardwired to FHIR 4.0.1 for now.
+        private static readonly IDictionary<string, ClassInfo> modelMapping = Models.ClassesById(Models.Fhir401);
```
EvanMachusak, line 2043: We should make the ModelInfo an abstract property on TypeResolver and require all implementations to provide one.  ExpressionBuilder can refer to it through its TypeResolver.  Theoretically, we need 1 TypeResolver per model declared in using statements in the provided library, e.g. to handle
```
using FHIR version '4.0.1'
using QICore version '1.0.0'
```

A single TypeResolver won't be enough for this.

EvanMachusak (changes requested):

Cql/Cql.Compiler/ExpressionBuilder.cs
ewoutkramer: Let's do that.

ewoutkramer: I cannot do this without a reorg of our project structure, since TypeResolver is in Abstractions, which cannot refer to Model, since Model already indirectly references Abstractions, so there's a circular reference.

However, I'd like to pull the PRs before we do a reorg, to avoid merge conflicts, so I have turned this suggestion into an issue to pick up after we've pulled the open PRs.

See https://github.com/FirelyTeam/firely-cql-sdk/issues/66.

PR #67, opened 2023-09-12
Fixes found during cql-to-elm prototype

EvanMachusak:
Some fixes in the ExpressionBuilder for some of the exotic expressions we were not handling.


Cql/Cql.Compiler/ExpressionBuilder.cs
```diff
@@ -145,63 +145,110 @@ public DefinitionDictionary<LambdaExpression> Build()
...
 
                         var newCodingExpression = Expression.New(codeCtor,
                             Expression.Constant(code.id),
                             Expression.Constant(csUrl),
                             Expression.Constant(null, typeof(string)),
                             Expression.Constant(null, typeof(string))!
                         );
-
                         var contextParameter = Expression.Parameter(typeof(CqlContext), "context");
                         var lambda = Expression.Lambda(newCodingExpression, contextParameter);
```
ewoutkramer, line 186: This lambda should create a `CqlConcept` (and include the description!), because that's what this construct is creating. If we are allowed to use that in the RHS of the retrieve, then we should let CqlConcept implement "list of codes".

ewoutkramer (changes requested):

Cql/Cql.Compiler/ExpressionBuilder.cs
EvanMachusak: Implemented & tested

PR #71, opened 2023-09-18
Added a fix in Tuple comparer; Updated to return if any first mismatc…

igajurel23:
Root cause:
The tuple comparer is comparing elements from across tuples, and summing up the differences on each mismatch found. This resulted into a summed up number that erroneously ends up as 0, which is used downstream to tell that there was a complete match between the tuples.

Fix/Update:
Updated to return from the function immediately upon finding a mismatch across elements, instead of traversing all elements, given the intension of TupleComparer is to tell that the tuple are not matching for any or all given elements under comparision.


Cql/Cql.Comparers/TupleBaseTypeComparer.cs
```diff
@@ -46,7 +46,11 @@ select new
                 var compare = MemberComparer.Compare(prop.XValue, prop.YValue, precision);
                 if (compare == null)
                     return null;
-                result += compare.Value;
+                if (compare.Value != 0)
+                {
+                    result += compare.Value;
```
EvanMachusak, line 51: Why not just return false here?

EvanMachusak (changes requested):

EvanMachusak (approved):

Issue #53, opened 2023-08-30
CollapseHelper does not currently handle properly included intervals


fweiand-ncqa:
**Description**
The current implementation of Collapse does not handle cases when one interval properly includes the other.

Currently, after sorting the intervals using their low value, the intervals are combined two at a time if the intervals meet or overlap.
The resulting interval always uses the low value from the first interval and the high value from the second. 
`return new CqlInterval<T?>(x.low, y.high, x.lowClosed, y.highClosed);`

https://github.com/FirelyTeam/firely-cql-sdk/blob/62ee395cecccca95566d56cefdbc72c6c53825e6/Cql/Cql.Operators/CqlOperators.IntervalOperators.cs#L369C21-L369C90

However, in a scenario when the first interval properly includes the second this causes a bug as the first interval has both the lowest and highest value.

**To Reproduce**
1. Create two intervals where one is properly included in the other
define "Interval 1": Interval[22, 28]
define "Interval 2": Interval[23, 25]

2. Collapse the intervals
define "Result": collapse("Interval 1", "Interval 2")

**Expected behavior**
The expected result is an interval from the lowest value to the highest:
Interval[22,28]
The actual result is an interval with the lowest value from the first interval and the highest value from the second.
Interval[22,25]

fweiand-ncqa:
Issue has been resolved with last merge

PR #72, opened 2023-09-19
Collapse fix properly included intervals

fweiand-ncqa:
This PR will update the collapse helper function to handle cases when one interval properly includes the other. 

At the moment, after sorting the intervals using their low value, the intervals are combined two at a time if the intervals meet or overlap.The resulting interval always uses the low value from the first interval and the high value from the second.

With this fix, the function will check if one interval includes the other and return the appropriate interval or merge if the two intervals if not.

Also created the a unit test to test this behavior.


fweiand-ncqa:
Resolves Issue #53

Cql/Cql.Operators/CqlOperators.IntervalOperators.cs
```diff
@@ -364,10 +364,14 @@ internal partial class CqlOperators
                 // In other words, adjacent intervals within a sorted list are merged if they either overlap or meet.
                 if ((meets(x, y, precision) ?? false) || (OverlapsHelper(x, y, precision, toClosed) ?? false))
                 {
-                    var lowClosed = x.lowClosed;
-                    var highClosed = y.highClosed;
-
-                    return new CqlInterval<T?>(x.low, y.high, x.lowClosed, y.highClosed);
+                    if  (IntervalIncludesInterval(x, y, precision!) ?? false)
```
ewoutkramer, line 367: Wouldn't we need to check for `IntervalIncludesInveral(y,x, precision1)` too?  In other words, does the same bug not exist when y properly includes x?

Cql/CoreTests/PrimitiveTests.cs
```diff
@@ -3449,6 +3449,30 @@ public void Sort_Lists_Dates_Containing_Null()
...
+                new CqlInterval<CqlDate>(new CqlDate(2023, 1, 20), new CqlDate(2023, 1, 28), true, true),
+                new CqlInterval<CqlDate>(new CqlDate(2023, 1, 22), new CqlDate(2023, 1, 25), true, true),
+                new CqlInterval<CqlDate>(new CqlDate(2023, 2, 20), new CqlDate(2023, 2, 25), true, true),
+                new CqlInterval<CqlDate>(new CqlDate(2023, 2, 18), new CqlDate(2023, 2, 28), true, true)
+            };
+
+            var collapsed = rtx.Operators.Collapse(intervals, null).ToArray();
+            var result = rtx.Operators.Comparer.Compare(expected!, collapsed!, null);
+            if (result != 0)
+                throw new AssertFailedException($"Expected {expected}; actual {collapsed}");
```
ewoutkramer, line 3473: What are the advantages of throwing over just `Assert.AreEqual(0, result)` ?

Cql/Cql.Operators/CqlOperators.IntervalOperators.cs
fweiand-ncqa: No, because the intervals are first sorted by their low value and added to the queue:

`var queue = ListSortBy(intervals, i => i == null ? null! : i.low!, ListSortDirection.Ascending)?.ToList();`

**This means that interval x will always have a low value that is less than or equal to interval y low value**. In the event that the low values are the same, and the y interval properly includes the x interval, the outcome will be correct when we return:

`return new CqlInterval<T?>(x.low, y.high, x.lowClosed, y.highClosed);`

Example: 
```
interval1 = [20,29]
interval2 = [20,28]
```

After sorting the queue will be either
`queue = { [20, 28], [20, 29]}`
or 
`queue = { [20, 29], [20, 28]}`

In the first case, interval y properly include interval x and will return [x.low, y.high] ( [20,29] )

In the second case interval x properly includes includes y which triggers the conditional and return x ( [20,29] )

Cql/CoreTests/PrimitiveTests.cs
fweiand-ncqa: No advantages that I know of. I used Collapse_Containing_Null as a template for building this test to follow convention and that uses AssertFailedException as well.

Cql/Cql.Operators/CqlOperators.IntervalOperators.cs
ewoutkramer: Perfect, just double checking, thanks!

Cql/CoreTests/PrimitiveTests.cs
ewoutkramer: This looks weird to me, but if it was a standing practice, I guess we should not bother about this case eithere.

ewoutkramer (approved):

PR #77, opened 2023-09-27
Enable the ignoreCase option for the enum parsing

DamienM419:
Resolve the issue of parsing {http://hl7.org/fhir}id into the proper type for "Id" in the FHIRAllTypes enum

### DIFF
--- a/Cql/Cql.Packaging/CqlCrosswalk.cs
+++ b/Cql/Cql.Packaging/CqlCrosswalk.cs
@@ -210,10 +210,10 @@ namespace Hl7.Cql.Packaging
                     int prefixLength = "{http://hl7.org/fhir}".Length;
                     var remainder = name.Substring(prefixLength);
                     var split = remainder.Split('.');
-                    if (Enum.TryParse<FHIRAllTypes>(split[0], true, out var fhirType))
-                        return TypeEntryFor(fhirType);
-                    else
-                        return null;
+                    if (ModelInfo.FhirTypeNameToFhirType(split[0]) is FHIRAllTypes primitiveType)
+                        return TypeEntryFor(primitiveType);
+
+                    return null;
                 }
                 else
                     return null;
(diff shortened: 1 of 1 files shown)


Cql/Cql.Packaging/CqlCrosswalk.cs
```diff
@@ -210,7 +210,7 @@ public CqlCrosswalk(TypeResolver typeResolver)
                     int prefixLength = "{http://hl7.org/fhir}".Length;
                     var remainder = name.Substring(prefixLength);
                     var split = remainder.Split('.');
-                    if (Enum.TryParse<FHIRAllTypes>(split[0], out var fhirType))
+                    if (Enum.TryParse<FHIRAllTypes>(split[0], true, out var fhirType))
```
ewoutkramer, line 213: This relies on the name of the enum field being the same as the type, which is not guaranteed. So you should use

`FHIRAllTypes? ModelInfo.FhirTypeNameToFhirType(string typeName)`

This is a static method, so you can do:

```
if( ModelInfo.FhirTypeNameToFhirType(remainder) is FhirAllTypes primitiveType)
{
   return TypeEntryFor(primitiveType);
}
```

ewoutkramer (changes requested):

Cql/Cql.Packaging/CqlCrosswalk.cs
DamienM419: This won't work when the name being passed in is "http://hl7.org/fhir}Claim.Item".  The remainder becomes "Claim.Item" and when you pass that into the ModelInfo.FhirTypeNameToFhirType(string typeName), the result is null. Passing in split works and generates the proper C# file using the PackagerCLI.

```
int prefixLength = "{http://hl7.org/fhir}".Length;
var remainder = name.Substring(prefixLength);
var split = remainder.Split('.');
if (ModelInfo.FhirTypeNameToFhirType(split[0]) is FHIRAllTypes primitiveType)
    return TypeEntryFor(primitiveType);

return null;
```

ewoutkramer (approved):

PR #80, opened 2023-10-06
feat: expose TypeResolver and TypeConverts

patburke234:
At NCQA, our existing DCS software utilizes the TypeResolver and TypeConverter classes. We use these to convert between the CQL and FHIR data models in various parts of our application. This PR is to expose the minimum set of classes/methods that we need publicly.


EvanMachusak (approved):

Cql/Cql.Firely/FhirTypeConverter.cs
```diff
@@ -18,8 +18,10 @@
 
 namespace Hl7.Cql.Fhir
 {
-    internal static class FhirTypeConverter
+    public static class FhirTypeConverter
```
ewoutkramer, line 21: Can you remove the 

`#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member` 

from the top of the file and document this class?

Cql/Cql.Packaging/CqlCrosswalk.cs
```diff
@@ -13,9 +13,9 @@
 using Hl7.Fhir.Model;
 namespace Hl7.Cql.Packaging
 {
-    internal class CqlCrosswalk
+    public class CqlCrosswalk
```
ewoutkramer, line 16: Can you remove the

`#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member`

from the top of the file and document this class?

Cql/Cql.Packaging/CqlCrosswalk.cs
```diff
@@ -247,7 +247,7 @@ public CqlCrosswalk(TypeResolver typeResolver)
         }
     }
 
-    internal class TypeEntry
+    public class TypeEntry
```
ewoutkramer, line 250: This feels like a record. We should remove the setters for the properties.  Maybe we can even make this a `record` class?

Cql/Cql.Packaging/CqlCrosswalk.cs
ewoutkramer, line 16: These classes do feel like internal utility classes to me.  What are they? A kind of mapping from a FHIR type to a CQL type?  Should this also work for other models?

ewoutkramer (changes requested):

Cql/Cql.Packaging/CqlCrosswalk.cs
patburke234: Yes, it's a mapping between FHIR and CQL types. We use it to convert our generated outputs to the data model our customers want (FHIR or CQL). I could see it expanding to other data models in the future, but I would think we'd need create a parent and rename this "FhirCqlCrosswalk", which is what it used to be named.

patburke234: Sure,  I will update it in a subsequent PR.

Cql/Cql.Firely/FhirTypeConverter.cs
patburke234: Sure,  I will update it in a subsequent PR.

ewoutkramer:
Oops, I didn't mean to merge it yet ;-)  Now we need another PR to fix my remarks. Don't know what happened there...

patburke234:
> Oops, I didn't mean to merge it yet ;-) Now we need another PR to fix my remarks. Don't know what happened there...
> 
> Interesting, I didn't. It's an automated message, I requested changes, but by doing so, I fulfilled some requirement and it got merged automatically. Mmmmm.....

Very strange. I'll work on a PR soon to get these items fixed.

PR #84, opened 2023-10-12
feat: Expose TypeResolver, CqlCrosswalk, and TypeConverter publicly

patburke234:
At NCQA, our existing DCS software utilizes the TypeResolver and TypeConverter classes. We use these to convert between the CQL and FHIR data models in various parts of our application. This PR is to expose the minimum set of classes/methods that we need publicly.

@ewoutkramer - seems my last change reverted my first change accidentally. I messed something up with Git. Here are all the same changes from #80  but with the changes you requested in the original PR.


Cql/Cql.Packaging/CqlCrosswalk.cs
```diff
@@ -247,18 +348,15 @@ public CqlCrosswalk(TypeResolver typeResolver)
...
-}
-#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
\ No newline at end of file
+    /// <summary>
+    /// Contains information about type mappings between CQL and FHIR type
+    /// </summary>
+    /// <param name="FhirType">the FHIR type</param>
+    /// <param name="CqlType">the CQL type</param>
+    /// <param name="ElementType">the element type, if the base type is a collection</param>
+    public record TypeEntry(
```
ewoutkramer, line 357: Is this then a `CqlTypeToFhirMapping` ?

ewoutkramer (changes requested):
@patburke234 - just a few more questions, before we make this type public for ever ;-)

Cql/Cql.Packaging/CqlCrosswalk.cs
```diff
@@ -1,27 +1,51 @@
...
 using Hl7.Fhir.Model;
+using System.Reflection;
+
 namespace Hl7.Cql.Packaging
 {
-    internal class CqlCrosswalk
+    /// <summary>
+    /// Defines the type mapping between Cql, ELM, and FHIR types
+    /// </summary>
+    public class CqlCrosswalk
```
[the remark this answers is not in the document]
patburke234: Sure, that's a good summary. It can map between those types in both directions.

Cql/Cql.Packaging/CqlCrosswalk.cs
patburke234: Yes, that's accurate. Can update these.

ewoutkramer (approved):

PR #88, opened 2023-10-16
Change to generated code caused by changes in rc2

ewoutkramer:
One of changes we made in the PRs since rc1 caused the generated code to change slightly.  This PRs contains these changes.


ewoutkramer:
Probably caused by #67, but I am not sure.

ewoutkramer:
@EvanMachusak - can you check whether these make sense? They don't look too weird to me, but I'd like to double check....

PR #97, opened 2023-10-27
chore: expose constants as public

patburke234:
At NCQA we need to be able to identify extension values that are useful for our measure processing. Specifically in this case, we need to be able to extract the CQL Default Value, CQL Default Value Type, and Type of the parameters using these constants.

### DIFF
--- a/Cql/Cql.Packaging/Constants.cs
+++ b/Cql/Cql.Packaging/Constants.cs
@@ -8,15 +8,42 @@
 
 namespace Hl7.Cql.Packaging
 {
-    internal static class Constants
+    /// <summary>
+    /// Constant values used when packaging measures
+    /// </summary>
+    public static class Constants
     {
+        /// <summary>
+        /// Measure group code system
+        /// </summary>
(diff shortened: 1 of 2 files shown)


ewoutkramer:
I'd rename this class from Constants to `NcqaExtensionUrls` or something, since that's what it is.

This:
`public const string MeasureGroupCodeSystem = "https://ncqa.org/fhir/CodeSystem/measure-group";`  sounds NCQA specific, so I wonder whether we should hard-code that into the exporter at all. @EvanMachusak what do you think?

I think that before we make these extension urls definitive (or even that codesystem url) - is this something that is simply missing form the current spec and we should urge HL7 to define these extensions in the HL7 space?

EvanMachusak:
The group code (https://hl7.org/fhir/r4/measure-definitions.html#Measure.group.code) doesn't have a recommended binding so we invented a CodeSystem in our IG for this.

I agree with you that anything NCQA specific should be removed from the SDK.  This is a custom step that we would inject into the packaging process to decorate the Measure resource with information.

EvanMachusak:
I am going to close this as #100 supercedes it.

PR #105, opened 2023-11-03
feat: refactor packager CLI for extensibility

patburke234:
At NCQA we need to extend the functionality of the current packager logic for our specific needs. This PR refactors the CLI to promote extensibility.


Cql/Cql.Packaging/ResourceWriters/FhirResourceWriter.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        /// <summary>
+        /// Writes the specified resources to the output directory in the FHIR format
+        /// </summary>
+        /// <param name="resources">the resources to write</param>
+        public override void WriteResources(IEnumerable<Resource> resources)
+        {
+            var options = new JsonSerializerOptions()
+                .ForFhir(typeof(Resource).Assembly)
```
ewoutkramer, line 29: This is incorrect, see https://docs.fire.ly/projects/Firely-NET-SDK/en/latest/parsing/system-text-json-serialization.html.

This is the way to do it:

```
Patient p = new() {  };
var options = new JsonSerializerOptions().ForFhir(ModelInfo.ModelInspector).Pretty();
string patientJson = JsonSerializer.Serialize(p, options);
```

ewoutkramer: Also, the Json.Text framework uses the options object to cache the reflection metadata, so it's better to keep it around if you do multiple `WriteResources()`.

ewoutkramer (changes requested):

Cql/Cql.Packaging/ResourceWriters/FhirResourceWriter.cs
patburke234: Updated.

ewoutkramer (approved):

PR #106, opened 2023-11-07
remove newline & whitespace from C# tags

richfirely:
Fixing annotations writing to the C# with newline and causing compiler to error. I considered doing this at the ELM stage, but these characters might be used for display purposes someday and only appears to affect annotations.


Cql/Cql.Runtime/DefinitionDictionary.cs
```diff
@@ -485,20 +486,49 @@ orderby g.Key
...
+            private static string CleanValue(string value)
+            {
+                if (string.IsNullOrEmpty(value)) return value;
+
+                char[] result = new char[value.Length];
+                int resultIndex = 0;
+                char space = (char)32;
+                char prevChar = space; 
+
+                foreach (char currentChar in value)
```
ewoutkramer, line 513: This algorithm is hard to read for me. It looks like it tries to convert (consecutive) whitespace into a single space, but if that is the case, I think there is a bug in it (inserting a space when seeing a whitespace char other than space does not change prevChar to "space", so we'll still insert whitespace the next time?).  In any case it's good to document what to "clean the field" means!

ewoutkramer (changes requested):

Cql/Cql.Runtime/DefinitionDictionary.cs
richfirely: Thanks good eye on the bug. I fixed it and also made the setup a little more clear by using trim (which is still fast and also cleans up entirely empty annotations of whitespace characters). I opted for this manual process because of performance - a replace over and over was pretty slow by like X times its used, because it goes back through the string over and over.

ewoutkramer: To be perfectly honest, because this code is just run at compile time (so once, when you are preparing the measure), it is not time critical as it is not run every time the measure runs.  So I would just do this:

```
var input = "Hello \r\n there, how are  you\tnowadays?";
var output = Regex.Replace(x, @"\s+", " ");
```

What do you think about that?

ewoutkramer (changes requested):

richfirely:
My pleasure to simplify. I overthought it because of speed.

ewoutkramer (approved):

PR #123, opened 2023-11-21
Manually upgraded packages to 8.0 to silence dependabot.

ewoutkramer:


marcovisserFurore (approved):
Maybe for another moment, but those package dependencies would work better in a shared property file. Then you have to update only 1 file.

PR #130, opened 2023-12-11
Add missing conversions

### LINKED ISSUE #129: CQLOperatorsBinding.BindConvert() returning Argument Exception for MedicationRequest.status
Hello,

I'm not sure if this is a bug, but would like to get some feedback on what I'm seeing. 

Attached is a CQL function that returns a MedicationRequest resource. After converting it into ELM and running it through the PackagerCLI, I receive the following error:

image

The error occurs in CQLOperatorsBinding.BindConvert(), approximately line 752. 

Tracing through the code, it looks like it attempts to determine if it can perform a conversion on MedicationRequest.status prior to the thrown exception. The check returns the conversion type as 'incompatible' because there doesn't appear to be a matching type in the TypeConverter dictionary. 

image

I've attached a CSV export of any references to 'MedicationRequest' found in the TypeConverter dictionary.

[TypeConverter.csv](https://github.com/FirelyTeam/firely-cql-sdk/files/13504913/TypeConverter.csv)

Just wanted to see if there is a way to resolve the conversion error. Are there additional type references that need to be added to the TypeConverter in CQL.Conversion? Please let me know if any additional information is needed.

Any help would be appreciated!

CQL and Elm files:
[cql_and_elm_example.zip](https://github.com/FirelyTeam/firely-cql-sdk/files/13507807/cql_and_elm_example.zip)

Call stack: 
[call_stack.txt](https://github.com/FirelyTeam/firely-cql-sdk/files/13504916/call_stack.txt)

EvanMachusak:
We haven't robustly tested the use case in #129 which requires round-tripping our conversions from FHIR model through CQL model back to FHIR model, which occurs when you attempt to create a FHIR resource in CQL, as is the case in the issue.

This PR:

- Adds conversions from `Code<T>` to `T?`
  - There may be a better way to do this
- Adds some FHIR type-to-string conversions

### DIFF
--- a/Cql/CoreTests/ExpressionBuilderTests.cs
+++ b/Cql/CoreTests/ExpressionBuilderTests.cs
@@ -9,6 +9,7 @@ using Hl7.Cql.Conversion;
 using Hl7.Fhir.Introspection;
 using Hl7.Fhir.Model;
 using Hl7.Cql.Abstractions;
+using System;
 
 namespace CoreTests
 {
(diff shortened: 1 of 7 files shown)


Cql/CoreTests/ExpressionBuilderTests.cs
```diff
@@ -45,5 +46,26 @@ public void QueriesTest_1_0_0()
...
+            var logger = CreateLogger();
+            var eb = new ExpressionBuilder(binding, typeManager, elmPackage, logger);
+
+            var fdt = new FhirDateTime(2023, 12, 11, 9, 41, 30, TimeSpan.FromHours(-5));
+            var fdts = fdt.ToString();
+            var fs = new FhirDateTime(fdts);
+            Assert.AreEqual(fdt, fs);
+
+            
+            var expressions = eb.Build();
```
ewoutkramer, line 63: your setting up an ExpressionBuilder, and then call Build(), but shouldn't you check something?

ewoutkramer (changes requested):

Cql/CoreTests/ExpressionBuilderTests.cs
EvanMachusak: Success for this test is that `Build()` doesn't throw any exceptions.

ewoutkramer: I understand, but it's nice for the reader (and those that maintain the test for you, later), that that is explicit. I'll add a comment & null check for you.

ewoutkramer: Weirdly enough, it did not build locally anymore because of localization and null-reference errors. Fixed that too. Maybe the new compiler that came with SDK 8 is a bit stricter than the one used in the CI build?  Anyway, fixed it.

ewoutkramer (approved):

PR #136, opened 2024-01-10
fix:  CqlTypeToFhirTypeMapper bug fixes + tests

patburke234:
2 bugs were identified inside of the CQL -> FHIR type mapper logic.

1) String type was being returned from the TypeResolver as "Net.System.String" which the mapper logic was unable to handle. We had some logic to handle primitive types further down, so I moved it up to handle this case properly.
2) LINQ Result expressions, list CastIterator, etc. were not being handled correctly since they do not implement IEnumerable<T>, List<T>, etc. I added some logic to handle these by checking if the type implements the non-generic IEnumerable interface.  

Added some covering unit tests for these use cases.

### DIFF
--- a/Cql/Cql.Runtime/BaseTypeResolver.cs
+++ b/Cql/Cql.Runtime/BaseTypeResolver.cs
@@ -144,11 +144,11 @@ namespace Hl7.Cql.Runtime
             if (type.IsGenericType)
             {
                 var genericTypeDefinition = type.GetGenericTypeDefinition();
-                if (genericTypeDefinition == typeof(IEnumerable<>))
-                    return type.GetGenericArguments()[0];
-                else if (genericTypeDefinition == typeof(List<>))
-                    return type.GetGenericArguments()[0];
-                else if (genericTypeDefinition == typeof(ICollection<>))
+                if (genericTypeDefinition == typeof(IEnumerable<>) 
+                     || genericTypeDefinition == typeof(List<>) 
+                     || genericTypeDefinition == typeof(ICollection<>)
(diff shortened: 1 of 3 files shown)


Cql/Cql.Runtime/BaseTypeResolver.cs
```diff
@@ -144,11 +144,11 @@ public override bool ImplementsGenericInterface(Type type, Type genericInterface
...
                 var genericTypeDefinition = type.GetGenericTypeDefinition();
-                if (genericTypeDefinition == typeof(IEnumerable<>))
-                    return type.GetGenericArguments()[0];
-                else if (genericTypeDefinition == typeof(List<>))
-                    return type.GetGenericArguments()[0];
-                else if (genericTypeDefinition == typeof(ICollection<>))
+                if (genericTypeDefinition == typeof(IEnumerable<>) 
+                     || genericTypeDefinition == typeof(List<>) 
+                     || genericTypeDefinition == typeof(ICollection<>)
+                     || genericTypeDefinition.GetInterfaces().Contains(typeof(System.Collections.IEnumerable)) //Handle LINQ Iterator types
```
EvanMachusak: This won't work for dictionary types, SortedList, Hash, etc.

Dictionary<TKey,TVale> has two generic arguments and implements IEnumerable.  Its implementation of IEnumerable returns a sequence of KeyValuePair<TKey,TValue> but the type you'll return on line 152 will be typeof(TKey).

EvanMachusak (approved):
I think ultimately we can obviate the need for this check in the future by getting tight about the types we return from CQL definitions (e.g. perhaps CqlList could be used) but this is okay for now.

PR #143, opened 2024-01-26
Use IValueSetFacade everywhere

ewoutkramer:
(this is the first of two PRs to bring in code that bridges the CQL's terminology concepts and the Firely SDK's terminology support. I split them in two to keep reviewing manageable)

Although IValueSetFacade existed already, there was still some code that used the concrete implementation `ValueSetFacade` instead. By using the interface everywhere we can (in a later PR) switch out the implementation by something that uses the existing FHIR ValueSet instead, or even a proxy to ITerminologyService.

It also allowed me to introduce another implementation, InMemoryValueSet, that made it possible to simplify ValueSetUnion and make code inside that class reusable.

This is what happened:
* Use IValueSetFacade instead of the concrete type
* Deduped code for creating ValueSetFacades
* Simplified ValueSetUnion to use the IValueSetDictionary of the facade.
* Added some nullability markers & applied some Resharper(Rider) suggestions.
* Renamed ValueSetFacade to CqlValueSetFacade, and added an additional IValueSetFacade implementation called InMemoryValueSet.
* Code for doing a union in ListOperators overlapped with that in the ValueSetUnion constructor, deduped.
* The union logic in ValueSetUnion itself as been made "lazy" and cached, and implemented using the new InMemoryValueSet. Basically the cache of codes in ValueSetUnion has been moved to InMemoryValueSet, so the ValueSetUnion got more simple, and it is now easier (and more explicit) to create a "cache" (= an InMemoryValueSet).
* Moved the CqlCodeEqualityHasher from ValueSetUnion to an extension method where it can easily be re-used so all components are guaranteed to use the same hasher.

Note: The ValueSet union had a constructor which took an IValueSetDictionary which it will use to look up the codes in the IValueSetFacades to join. That is confusing, since these facades themselves already have a reference to their IValueSetDictionary. It seems a source of errors if these become incompatible, and also it seems to make sense that the ValueSetFacades "know best", so I changed the code so the valueset facades will provide the codes themselves to the union, instead of them being taken from another (possibly inconsistent) reference to IValueSetDictionary.


.idea/.idea.Cql/.idea/.name
baseTwo, line 1: Best to ignore the .idea folder and only include team related settings explicitly. Here is a possible example https://github.com/salarmehr/idea-gitignore

EvanMachusak (approved):

PR #152, opened 2024-02-05
Feature/151 packagecli take rooturl arg

### LINKED ISSUE #151: PackageCLI to take a command line argument to replace the cononical root url '#' with an actual value
# Background
The PackageCLI generates library resources as json files (as described [here](https://firely.atlassian.net/wiki/spaces/CQLSDK/pages/1359085592/How+to+Running+VONK-5284-load-vs-cql+Branch#R4%5CDemo%5CPUT-Create-Library---AdultOutpatientEncountersFHIR4-2.2.000)). These files need to be manually updated before sending them to the Create Library endpoint on Firely Server. The url property starts with a # sign, and this needs to be replaced with the actual root url.

# Solution
Look at specs and update the PackageCLI to take an additional argument for the root uri and replace the # with that value if it is provided.

baseTwo:
#Work done for #151 
* Add support to provide "--canonical-root-url"
* Validate and reject unsupported args
* Removed NoWarns in solution
* Fix launchsettings (was using incorrect arguments, which was detected by adding step above)

### DIFF
--- a/Cql/Cql.Packaging/ResourcePackager.cs
+++ b/Cql/Cql.Packaging/ResourcePackager.cs
@@ -29,9 +29,19 @@ namespace Hl7.Cql.Packaging
         /// </summary>
         /// <param name="elmDir">directory to find the ELM files</param>
         /// <param name="cqlDir">directory to find the CQL files</param>
-        /// <param name="resourceCanonicalRootUrl">root part of the resource canonical; otherwise just a '#'</param>
         /// <param name="afterPackageMutator">optional mutator for the resources prior to writing</param>
-        public void Package(DirectoryInfo elmDir, DirectoryInfo cqlDir, string? resourceCanonicalRootUrl = null, Action<IEnumerable<Resource>>? afterPackageMutator = null)
+        public void Package(DirectoryInfo elmDir, DirectoryInfo cqlDir, 
+            Action<IEnumerable<Resource>>? afterPackageMutator = null) =>
+            PackageCore(elmDir, cqlDir, afterPackageMutator, null);
+
+        /// <summary>
+        /// Package the resources in the given ELM and CQL directories and output them using the writers provided in the constructor 
+        /// </summary>
+        /// <param name="args">A</param>
+        public void Package(PackageArgs args) =>
+            PackageCore(args.ElmDir, args.CqlDir, args.AfterPackageMutator, args.ResourceCanonicalRootUrl);
+
+        private void PackageCore(DirectoryInfo elmDir, DirectoryInfo cqlDir, Action<IEnumerable<Resource>>? afterPackageMutator, string? resourceCanonicalRootUrl)
         {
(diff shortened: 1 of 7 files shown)


Cql/Cql.Packaging/ResourcePackager.cs
```diff
@@ -29,8 +29,9 @@ public class ResourcePackager
         /// </summary>
         /// <param name="elmDir">directory to find the ELM files</param>
         /// <param name="cqlDir">directory to find the CQL files</param>
+        /// <param name="resourceCanonicalRootUrl">root part of the resource canonical; otherwise just a '#'</param>
         /// <param name="afterPackageMutator">optional mutator for the resources prior to writing</param>
-        public void Package(DirectoryInfo elmDir, DirectoryInfo cqlDir, Action<IEnumerable<Resource>>? afterPackageMutator = null)
+        public void Package(DirectoryInfo elmDir, DirectoryInfo cqlDir, string? resourceCanonicalRootUrl = null, Action<IEnumerable<Resource>>? afterPackageMutator = null)
```
ewoutkramer: This change would cause a compiler error for current callers, right?  Could you place the new argument last?

Cql/Cql.Runtime/Cql.Runtime.csproj
```diff
@@ -7,7 +7,6 @@
 		<RootNamespace>Hl7.Cql.Runtime</RootNamespace>
 		<Product>CQL engine runtime</Product>
 		<Description>Support package for Hl7.Cql. Contains the required runtime components for CQL execution.</Description>
-		<NoWarn>CS8601</NoWarn>
```
ewoutkramer, line 10: Never noticed it was here. Good riddance.

cql-sdk.props
```diff
@@ -18,6 +18,7 @@
 		<PackageReadmeFile>README.md</PackageReadmeFile>
 		<PublishRepositoryUrl>true</PublishRepositoryUrl>
 		<EmbedUntrackedSources>true</EmbedUntrackedSources>
+		<NoWarn>CA1311,CS8601,CS8602,CS8604</NoWarn>
```
ewoutkramer: Do we really need to suppress CS8601, CS8602 and CS8604? They are pretty useful, right?  CA1311 used to be quite important to flag internationalization issues (the original codebase did not work with a Dutch locale for example).

ewoutkramer (changes requested):
Nice work. Just a few questions.

cql-sdk.props
baseTwo: Removed all of these and code still compiles OK

Cql/Cql.Packaging/ResourcePackager.cs
baseTwo: I kept the original methods and added an overload to take PackageArgs

ewoutkramer (approved):

Issue #163, opened 2024-02-09
Performance with FhriDateTime and strings


DamienMosier:
We saw a performance issue when converting to Firely SDK that took a deck from 20 minutes to 72 minutes to run. The issue was traced down to FhirDateTime objects.

The FhirTypeConverter.cs contained an FhirdateTime to CqlDateTime converter but never used.

```
add((M.FhirDateTime f) => f.TryToDateTime(out var dt) ?
    new CqlDateTime(
        dt!.Years!.Value, dt.Months,
        dt.Days, dt.Hours, dt.Minutes, dt.Seconds, dt.Millis,
        dt.HasOffset ? dt.Offset!.Value.Hours : null, dt.HasOffset ? dt.Offset!.Value.Minutes : null) : null);
```

In the FhirHelpers cql (https://fhir.org/guides/cqf/common/Library-FHIRHelpers.html) we have ToInterval

```
define function ToInterval(period FHIR.Period):
    if period is null then
        null
    else
        if period."start" is null then
            Interval(period."start".value, period."end".value]
        else
            Interval[period."start".value, period."end".value]

```

The ".value" portion of the CQL creates lambda's like this where the value is referenced.

```
.If ($var8 != null) {
                                (.Extension<Hl7.Cql.Compiler.NullConditionalMemberExpression> {
                                    .Block(Hl7.Fhir.Model.Period $var10) {
                                        $var10 = $period;
                                        .If ($var10 != null) {
                                            $period.StartElement
                                        } .Else {
                                            .Default(Hl7.Fhir.Model.FhirDateTime)
                                        }
                                    }
                                }).Value
```
In the ExpressionBuilder.cs file there is a PropertyHelper the value is accessed as a string and eventually drills down and calls CqlOperators.ConversionFunctionName to get "ConvertStringToFhirDateTime" which parses the date into the DateTimeIso8601.TryParse. This uses regex to parse the date. After profiling this is indicative of a huge performance issue.

Need to implement a way to indicate not to drill down into the "value" property and to use the source object (FhirDateTime) so the converter is used to parse the date.

PR #167, opened 2024-02-12
Enhance performance with FhirDateTime

DamienMosier:
Enhancements for https://github.com/FirelyTeam/firely-cql-sdk/issues/163

Implements a LRU cache for the date parsing to remove the need to do TryToDateTime on strings if we've already converted it once. Default size of the cache to 10k but it can be controlled and adjusted.

Add new function to bypass the need to use ",value" lookup and use the source object for FhirDateTime objects so we can use the converter that already existed.


Cql/Cql.Packaging/LibraryPackager.cs
```diff
@@ -31,12 +31,17 @@ public class LibraryPackager
     {
         internal LibraryPackager()
         {
-            TypeConverter = FhirTypeConverter.Create(ModelInfo.ModelInspector);
+            TypeConverter = FhirTypeConverter.Create(ModelInfo.ModelInspector, 0);
         }
 
-        internal LibraryPackager(TypeConverter? typeConverter)
+        internal LibraryPackager(int cacheSize)
```
ewoutkramer: I think these new constructors move the responsibility of setting the cache size from the TypeConverter (where they belong) into the LibraryPackager. Ultimately, we just need the constructor that takes a TypeConverter. Especially the overload taking a converter + a cache size is confusing, since it allows either the passed in typeconverter or the passed in cache size to be relevant, but for the "outside user" it is unclear what happens what you pass both.  I think we should clean this up if possible.

Cql/Cql.Packaging/PublicAPI.Unshipped.txt
```diff
@@ -23,13 +23,19 @@ Hl7.Cql.Packaging.CqlTypeToFhirMapping.CqlTypeToFhirMapping(Hl7.Cql.Packaging.Cq
 Hl7.Cql.Packaging.CqlTypeToFhirMapping.CqlTypeToFhirMapping(Hl7.Fhir.Model.FHIRAllTypes? FhirType, Hl7.Cql.Primitives.CqlPrimitiveType? CqlType, Hl7.Cql.Packaging.CqlTypeToFhirMapping? ElementType = null) -> void
 Hl7.Cql.Packaging.CqlTypeToFhirMapping.Deconstruct(out Hl7.Fhir.Model.FHIRAllTypes? FhirType, out Hl7.Cql.Primitives.CqlPrimitiveType? CqlType, out Hl7.Cql.Packaging.CqlTypeToFhirMapping? ElementType) -> void
 Hl7.Cql.Packaging.PackageArgs
-Hl7.Cql.Packaging.PackageArgs.PackageArgs(System.IO.DirectoryInfo! elmDir, System.IO.DirectoryInfo! cqlDir, System.Action<System.Collections.Generic.IEnumerable<Hl7.Fhir.Model.Resource!>!>? afterPackageMutator = null, string? resourceCanonicalRootUrl = null) -> void
 Hl7.Cql.Packaging.PackageArgs.ElmDir.get -> System.IO.DirectoryInfo!
 Hl7.Cql.Packaging.PackageArgs.CqlDir.get -> System.IO.DirectoryInfo!
 Hl7.Cql.Packaging.PackageArgs.AfterPackageMutator.init -> void
 Hl7.Cql.Packaging.PackageArgs.AfterPackageMutator.get -> System.Action<System.Collections.Generic.IEnumerable<Hl7.Fhir.Model.Resource!>!>?
+Hl7.Cql.Packaging.PackageArgs.LRUCacheSize.get -> int
```
ewoutkramer: It does not feel right to me that we have to pass in the cache size everywhere, even if the caller has no idea why. E.g. `LoadElm` and `EvaluateExpression` now take a cache size argument, but it feels too much like we're exposing an internal detail.  In other parts of the SDK I have used a static global (which is an anti-pattern by itself) for the cache size.  Any other options?

ewoutkramer (changes requested):
I think we have to find an alternative to passing the cache size everywhere, even in parts of the public surface where it feels like an internal detail.

Cql/Cql.Packaging/LibraryPackager.cs
DamienMosier: Upon further review, the cache size is only needed during the run in a context of a patient, not packaging or any other location. The user can control the cache size in the FhirModelBindingOptions object for this purpose so I am removing all this unnecessary code.

Cql/Cql.Packaging/PublicAPI.Unshipped.txt
DamienMosier: I don't think the cache size needs to be specified during the packaging process as it's only needed during execution so it is being removed.

Cql/Cql.Packaging/LibraryPackager.cs
ewoutkramer: I like that even better.

ewoutkramer (approved):

Issue #169, opened 2024-02-13
Introduce PackageCLI argument for specifying a fixed datetime used in output library resource json files


baseTwo:
When the library resource json files are generated, the current datetime is used for the "date" property in the output e.g.
![image](https://github.com/FirelyTeam/firely-cql-sdk/assets/4639799/fb3f4d96-275a-49bf-80ca-b4ad7922ce7f)

This is all good for production purposes, but during development of the code generator, it would be more beneficial to fix this value so that the file stays the same as new features are added, and won't appear as hundreds of files changing in the source repo.

Propose adding an argument for packageCLI to take a datetime value for this, so that a fixed value can be used during development

PR #171, opened 2024-02-13
Allow date to be overriden in resource library's json

baseTwo:
Refer to issue #169 for details.

A new argument `--override-utc-date-time` which overrides the date used in the resource library e.g. 
image

### DIFF
--- a/Cql/PackagerCLI/Program.cs
+++ b/Cql/PackagerCLI/Program.cs
@@ -1,6 +1,9 @@
 ﻿#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+using System.Diagnostics;
+using System.Globalization;
 using Hl7.Cql.Packaging;
 using Hl7.Cql.Packaging.ResourceWriters;
+using Hl7.FhirPath.Sprache;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Serilog;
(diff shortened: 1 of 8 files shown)


Cql/PackagerCLI/Program.cs
```diff
@@ -139,15 +139,19 @@ private static int ShowHelp()
...
-                    [--canonical-root-url] <url>  The root url used for the resource canonical. If omitted a '#' will be used
+                    --elm       <directory>         Library root directory
+                    --cql       <directory>         CQL root directory
+                    [--fhir]    <file|directory>    Resource location, either file name or directory
+                    [--cs]      <file|directory>    C# output location, either file name or directory
+                    [--d]       <true|false>        Produce as a debug assembly
+                    [--f]       <true|false>        Force an overwrite even if the output file already exists
+                    [--canonical-root-url] <url>    The root url used for the resource canonical. 
+                                                    If omitted a '#' will be used
+                    [--override-utc-date-time] <utc-date-time>
```
EvanMachusak, line 152: Recommend clarifying the expected string format - presumably ISO8601.  An example in the help would be useful.

EvanMachusak:
This is a lot of unexplained rewrite for a relatively simple feature.  This can be done much more surgically.

Cql/PackagerCLI/Program.cs
baseTwo: This PR is not ready for review. I have one in the workings with an example in the help

Issue #166, opened 2024-02-12
Generate C# output files on the fly


baseTwo:
# Current Problem
During the development of the PackageCLI tool, bugs are discovered on new input sets and fixed, however, as part of fixing one bug in one file, another one pops up in another file. Since there is no output generated until all input files are processed successfully in-memory, it is impossible to know if the first one was fixed correctly.

# Solution
Generate files as soon as possible during the processing of input files. Input files need to be ordered deterministically and such that files with no dependencies are run first. Temporary files must be created during the process, after which original output files can be overwritten.

Propose adding another argument for the PackageCLI specifying where input files are generated. Temp files must have a temp file extension, and only after everything is generated, can they replace the orginal files.

EvanMachusak:
>  bugs are discovered on new input sets and fixed, however, as part of fixing one bug in one file, another one pops up in another file. 

Can you clarify what you mean here?  I have spent a lot of time debugging codegen over the past couple of years and I have not had an issue with the way PackagerCLI currently works.

The packager CLI does three things:

1) Translates ELM to Expressions using the ExpressionBuilder
2) Translates Expressions to C# source code and compiles it using Roslyn
3) Assembles FHIR resources from the source files.

Step 1 can be tested independently for single files and errors can be observed immediately.  If you don't want to compile them to Roslyn you can compile the lambda expressions generated by the ExpressionBuilder - there is a CompileAll extension method that does this - and you will see any translation errors there.

You can also opt to write the C# files to a directory that contains a .csproj and then include that .csproj in a solution.  That is how we develop measures.  We compile CQL to ELM => ELM to C# => then use Visual Studio to compile the C#.  It makes errors very easy to see and debug, since we can see all of the source code.  If you haven't set this up for the CQL files you are trying to test, I would recommend doing so.  It will make your life a lot easier.

Step 2 cannot be tested independently because dependencies matter.  If library B depends on library A, library A's C# will have a reference (as a member) to library B, and will obviously not compile unless library B's class will resolve.

Step 3 is already independent in the code (`createLibraryResource` in LibraryPackager.cs) but is private.  You could make it internal to test it independently if you like.

[...]

Issue #148, opened 2024-02-01
Exception when compiling CMS measure HFBetaBlockerTherapyforLVSDFHIR-1.3.000


alexzautke:
Getting ArgumentException while building statement for definition 'Initial Population 'in library 'HFBetaBlockerTherapyforLVSDFHIR-1.3.000'. 

```
System.ArgumentException: Cannot resolve type for expression
   at Hl7.Cql.Compiler.TypeManager.TypeFor(Element element, ExpressionBuilderContext context, Boolean throwIfNotFound)
      in C:\Dev\firely-cql-sdk\Cql\Cql.Compiler\TypeManager.cs:line 130
   at Hl7.Cql.Compiler.TypeManager.TypeFor(Element element, ExpressionBuilderContext context, Boolean throwIfNotFound)
      in C:\Dev\firely-cql-sdk\Cql\Cql.Compiler\TypeManager.cs:line 104
   at Hl7.Cql.Compiler.ExpressionBuilder.Property(Property op, ExpressionBuilderContext ctx)
      in C:\Dev\firely-cql-sdk\Cql\Cql.Compiler\ExpressionBuilder.cs:line 2266
   at Hl7.Cql.Compiler.ExpressionBuilder.TranslateExpression(Element op, ExpressionBuilderContext ctx)
      in C:\Dev\firely-cql-sdk\Cql\Cql.Compiler\ExpressionBuilder.cs:line 803
   at Hl7.Cql.Compiler.ExpressionBuilder.Property(Property op, ExpressionBuilderContext ctx)
      in C:\Dev\firely-cql-sdk\Cql\Cql.Compiler\ExpressionBuilder.cs:line 2247
   at Hl7.Cql.Compiler.ExpressionBuilder.TranslateExpression(Element op, ExpressionBuilderContext ctx)
      in C:\Dev\firely-cql-sdk\Cql\Cql.Compiler\ExpressionBuilder.cs:line 803
   at Hl7.Cql.Compiler.ExpressionBuilder.CalculateAgeAt(CalculateAgeAt e, ExpressionBuilderContext ctx)
      in C:\Dev\firely-cql-sdk\Cql\Cql.Compiler\ExpressionBuilder.ClinicalOperators.cs:line 34
   at Hl7.Cql.Compiler.ExpressionBuilder.TranslateExpression(Element op, ExpressionBuilderContext ctx)
[...]

PR #173, opened 2024-02-14
Fix/cql 148 Refactor Definition Building, Fix build for ExpressionRef

baseTwo:
# Fix
Fix for [issue 148](https://github.com/FirelyTeam/firely-cql-sdk/issues/148) Getting ArgumentException while building statement for definition 'Initial Population 'in library 'HFBetaBlockerTherapyforLVSDFHIR-1.3.000'.

Since there was some refactoring done as well (moving library definition building into a separate class), the fix was in `TypeManager` line 86, handling ExpressionRef and getting the return type from the previous definitions by name. (Side note : Separating the refactoring into another PR would make the relavant issue more difficult to fix) 

# More Exceptions
Note that upon running this code on the cms measures branch, another exception is discovered later on. Since there is no output files generated until after processing all input files in-memory successfully first, it is not possible to check if this PR actually fixed the problem correctly. [A new feature request 166](https://github.com/FirelyTeam/firely-cql-sdk/issues/166) was submitted to enable generating temporary output files as soon as possible. 

# Refactor 
Definitions Building  was refactored out of `ExpressionBuilder` into `ExpressionBuilder.DefinitionsBuilder`. This was done because ExpressionBuilder is already a huge class, definitions building adds a lot of extra state.
* Direct access to the `ExpressionBuilder` and  `ExpressionBuilderContext` were encapsulated by the `ExpressionBuilder.DefinitionsBuilderContext` and `DefinitionsBuilder.ExpressionBuilderContextFacade` classes respectively.
* Prevent leaking out List<> or Dictionary<>
image

# Utilities Added
* `ReflectionUtility` makes it easier to build reflection info, by using expressions.
* `ConstructorInfos` creates and caches ConstructorInfo instances used often

# Other
* `DirectedGraph` has a property to generate a graphviz diagram
* Move code out of the `Library` class, since the generated part is attributed with `[DebuggerStepThrough]` which makes it tedious to debug. The code is now in `LibraryExtensions`
* Found a bug when running the PackageCLI from a directory which contains dashes, so placed the arguments in quotes in launchSettings.json.


Cql/Cql.Compiler/ExpressionBuilderContext.cs
```diff
@@ -236,8 +204,8 @@ internal ExpressionBuilderContext WithImpliedAlias(string aliasName, Expression
         /// </summary>
         internal ExpressionBuilderContext Deeper(elm.Element expression)
         {
+            _ = expression;
```
ewoutkramer: Looks like a no-op, unless we have an overloaded `=` operator with side effects ;-)

Cql/Cql.Compiler/Infrastructure/ReflectionUtiltity.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// <summary>
+    /// Retrieves the underlying type from a nullable type; or just the type itself if it wasn't nullable.
+    /// </summary>
+    /// <param name="type">The type to check.</param>
+    /// <returns>A tuple containing the underlying type (or specified type) and a boolean indicating if the type was nullable.</returns>
+    public static (Type nonNullableType, bool isNullable) GetUnderlyingTypeForNullable(this Type type)
+    {
+        return type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>)
+            ? (type.GetGenericArguments()[0], true)
```
ewoutkramer: Isn't that the same as `Nullable.GetUnderlyingType()`

ewoutkramer (approved):

Cql/Cql.Compiler/Infrastructure/ReflectionUtiltity.cs
baseTwo: The use case is that we need to know if the original type was nullable or not, so that we can map expression types when assigning between T and T?

Cql/Cql.Compiler/ExpressionBuilderContext.cs
baseTwo: It is a no-op, otherwise the compiler complains. Originally the expression was added to a list of predecessors, which in itself wasn't used. I will remove it now, but for the future it would make sense to reintroduce this again so that we can better build exception messages with the expression path that got us to that.

Cql/Cql.Compiler/Infrastructure/ReflectionUtiltity.cs
ewoutkramer: I meant: 

```
   return Nullable.GetUnderlyingType(type) is {} t ?  (t, true) : (type,false);
```

EvanMachusak: Agree with Ewout.  GetUnderlyingType does this and returns null if the type is not nullable.  You don't need to construct a tuple with a boolean - you can just compare the resulting `Type` with `null` to know that.

Cql/Cql.Compiler/CqlOperatorsBinding.cs
```diff
@@ -509,7 +509,7 @@ private Expression SortBy(MemberExpression operators, Expression source, Express
         {
             if (by is LambdaExpression lambda && order is ConstantExpression orderConstant && orderConstant.Type == typeof(ListSortDirection))
             {
-                var elementType = TypeResolver.GetListElementType(source.Type) ?? throw new InvalidOperationException($"{source.Type} was expected to be a list type.");
+                var elementType = TypeResolver.GetListElementType(source.Type).NotNull($"'{source.Type}' was expected to be a list type.");
```
EvanMachusak: This is a meaningless change that reduces performance since method calls are more expensive than null coalescing.  Why do this?

Cql/Cql.Compiler/Extensions.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Hl7.Cql.Elm;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Hl7.Cql.Compiler
+{
+    internal static class Extensions
+    {
+        public static string OrEmptyString(this string? value)
```
EvanMachusak: All of these methods are unnecessary and reduce performance.  Every C# programmer can read `x ?? ""` as "x or empty string" without a method name spelling it out.

Cql/Cql.Compiler/TypeManager.cs
```diff
@@ -62,97 +63,127 @@ public TypeManager(TypeResolver resolver, string assemblyName = "Tuples", string
             AssemblyName = assemblyName;
             var assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(new AssemblyName(AssemblyName), AssemblyBuilderAccess.Run);
             ModuleBuilder = assemblyBuilder.DefineDynamicModule(AssemblyName);
-            Resolver = resolver ?? throw new ArgumentNullException(nameof(resolver));
-
+            Resolver = resolver.ArgNotNull();
```
EvanMachusak: More unnecessary method calls.  The expression you're replacing is the one automatically created by Visual Studio when you Quick Actions => Add Null Check.

Cql/Cql.Packaging/LibraryPackager.cs
```diff
@@ -39,20 +39,14 @@ internal LibraryPackager(TypeConverter? typeConverter)
...
-            {
-                var library = elm.Library.LoadFromJson(file);
-                if (library?.NameAndVersion != null)
-                {
-                    dict.TryAdd(library.NameAndVersion, library);
-                }
-            });
-            return dict;
-        }
+        public static IDictionary<string, elm.Library> LoadLibraries(DirectoryInfo elmDir) =>
```
EvanMachusak: This is an unncessary change that makes the code harder to debug.  You will find throughout this SDK that I write code in a way that you can easily conclude, "But you can do that in just a couple of LINQ statements."  I used to write code like this proposed change, and then stopped doing it because I constantly ended up rewriting my code to look like the original function here because I wanted to put a break point somewhere and couldn't.

EvanMachusak (changes requested):
There are too many unnecessary changes in this pull request that have nothing to do with the issue in question.  Most of them are not useful changes and are just refactoring to taste.

Cql/Cql.Compiler/Extensions.cs
baseTwo: Got rid of this class

Cql/Cql.Packaging/LibraryPackager.cs
baseTwo: The change was necessary because we need to have a deterministic outcome, by loading the input files in order. The original Parallel.For could complete loading elm files in a rondom order. I changed it back to using the original Parallel.For, but added code to keep it deterministic

Cql/Cql.Compiler/ExpressionBuilderContext.cs
baseTwo: I think I'll just remove this from the original code as well

Cql/Cql.Compiler/CqlOperatorsBinding.cs
baseTwo: If it is compiled inline as in the screenshot example below, then there is no method call. I any case, I removed the Check class. 

image

Cql/Cql.Compiler/ExpressionBuilderContext.cs
baseTwo: Done

Cql/Cql.Compiler/Infrastructure/ReflectionUtiltity.cs
baseTwo: Done

Cql/Cql.Compiler/TypeManager.cs
baseTwo: Reverted

Cql/Cql.Compiler/CqlOperatorsBinding.cs
ewoutkramer: I actually like it, and I don't think these small performance differences mean anything in the compiling stage - reading the files, I/O, generating dynamic types, compiling Linq Expressions and calling the Roslyn compiler make these (inlined?) calls irrelevant.

PR #180, opened 2024-02-15
172 Use Dependency Injection Part 1 : Introduce the host builder

baseTwo:
Introduce the .net core Host feature. Out of the box there are many features like service creation and lifetime management, providing dependency injection between services. 

The first step is to build the host and configure the main service `PackagerService` which will be run synchronously. It accepts a `PackagerOptions` that contains the validated switches from the command line.

Additionally, the `PackagerService.Run` method is wrapped up in a try-catch which logs any exceptions, but displays a less noise respose to the console by mentioning that an error occurred and the logs needs to be consulted.

Dependencies are described in the `DependencyInjection` class. Currently the following classes are managed: `PacagerService`, `ResourcePackager`, `FhirResourceWriter` and `CSharpResourceWriter`. The latter two will be registered as keyed services (so the ResourceWriter abstraction can be retrieved by name Fhir or CSharp, as well as IEnumerable services, so they can be injected into `ResourcePackager`.


Cql/Cql.Packaging/PublicAPI.Unshipped.txt
```diff
@@ -46,11 +46,11 @@ virtual Hl7.Cql.Packaging.PackageArgs.EqualityContract.get -> System.Type!
 override Hl7.Cql.Packaging.PackageArgs.ToString() -> string!
 Hl7.Cql.Packaging.ResourcePackager
 Hl7.Cql.Packaging.ResourcePackager.Package(Hl7.Cql.Packaging.PackageArgs! args) -> void
-Hl7.Cql.Packaging.ResourcePackager.ResourcePackager(Microsoft.Extensions.Logging.ILoggerFactory! logFactory, params Hl7.Cql.Packaging.ResourceWriters.ResourceWriter![]! resourceWriters) -> void
+Hl7.Cql.Packaging.ResourcePackager.ResourcePackager(Microsoft.Extensions.Logging.ILoggerFactory! logFactory, System.Collections.Generic.IEnumerable<Hl7.Cql.Packaging.ResourceWriters.ResourceWriter!>! resourceWriters) -> void
```
ewoutkramer, line 50: Can we keep an overload with the `params`?

Cql/Cql.Packaging/ResourceWriters/CSharpResourceWriter.cs
```diff
@@ -13,7 +14,7 @@ public class CSharpResourceWriter : ResourceWriter
         /// </summary>
         /// <param name="outDirectory">the output directory</param>
         /// <param name="logger">logger</param>
-        public CSharpResourceWriter(DirectoryInfo outDirectory, ILogger logger) : base(outDirectory, logger)
+        public CSharpResourceWriter([FromKeyedServices("CSharp")] DirectoryInfo outDirectory, ILogger<CSharpResourceWriter> logger) : base(outDirectory, logger)
```
ewoutkramer: He! That dit not exist before! Ah, it's a .NET8 thing!

Cql/PackagerCLI/PackagerOptions.Validator.cs
```diff
@@ -0,0 +1,101 @@
...
+                    AddError($"{arg.name}: directory '{dir}' could not be created.");
+                    return false;
+                }
+
+                return true;
+            }
+
+            void AddError(string message)
+            {
+                if (!slimLock.Wait(Timeout.Infinite))
```
ewoutkramer: I can't figure out why we need to lock!

Cql/PackagerCLI/PackagerOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+            [ArgNameFhirDirectory]    = ConfigSection + ':' + nameof(FhirDirectory),
+            [ArgNameCSharpDirectory]  = ConfigSection + ':' + nameof(CSharpDirectory),
+            [ArgNameDebug]            = ConfigSection + ':' + nameof(Debug),
+            [ArgNameForce]            = ConfigSection + ':' + nameof(Force),
+            [ArgNameCanonicalRootUrl] = ConfigSection + ':' + nameof(CanonicalRootUrl),
+            // @formatter:on
+        };
+
+    public const string ArgNameElmDirectory = "--elm";
+    public DirectoryInfo ElmDirectory { get; set; } = null!;
```
ewoutkramer, line 8: Is this a syntactic way to express that you promise this property never to be null in the rest of the application, except just before initialization?

Cql/PackagerCLI/DependencyInjection.cs
```diff
@@ -0,0 +1,66 @@
...
+            if (descriptors.Count > 0)
+            {
+                services.TryAddEnumerable(descriptors);
+            }
+        }
+    }
+
+    private static PackagerOptions? TryGetPackagerOptions(IServiceCollection services)
+    {
+        using var sp = services.BuildServiceProvider(new ServiceProviderOptions { ValidateOnBuild = false, ValidateScopes = false });
```
ewoutkramer: What is this doing? (sorry, I am only superficially familiar with the services and hosting libraries, so this makes my head spin ;-))

Cql/PackagerCLI/DependencyInjection.cs
```diff
@@ -0,0 +1,66 @@
...
+            .ValidateOnStart();
+        services.AddSingleton<IValidateOptions<PackagerOptions>, PackagerOptions.Validator>();
+    }
+
+    public static void AddResourcePackager(this IServiceCollection services)
+    {
+        services.TryAddPackagerOptions();
+        services.TryAddSingleton<ResourcePackager>();
+
+        var unvalidatedPackagerOptions = TryGetPackagerOptions(services);
```
ewoutkramer: This feels like higher math to me!

ewoutkramer: Are we setting up a nested container, to get to the options before they are really validated by the framework itself?

ewoutkramer: (Ewout is now wondering whether we are doing such exotic things that this is not supported out of the box - maybe because of the DirectoryInfo settings?)

Cql/Cql.Packaging/ResourceWriters/CSharpResourceWriter.cs
baseTwo: Indeed

Cql/PackagerCLI/PackagerOptions.cs
baseTwo: Yes, after the validation of this options class, we can safely assume it will always be not-null. The options are validated before injected into services, so they dont need to check again

Cql/PackagerCLI/DependencyInjection.cs
baseTwo: I rewrote this so we dont have to create a serviceprovider during the process of describing the services.

Cql/Cql.Packaging/PublicAPI.Unshipped.txt
baseTwo: We can, I'll do this now

baseTwo: Done

Cql/PackagerCLI/PackagerOptions.Validator.cs
baseTwo: There were previously some async code in here which is now removed. So no need for a lock anymore

PR #188, opened 2024-02-20
Inference refactoring

EvanMachusak:
- Refactored invocation signature matching by separating concerns into:
  - Type conversion via the TypeConverter service
  - Signature matching via the InvocationBuilder service
  - ELM node creation via the ElmFactory

The TypeConverter attempts to exactly match [Developer's Guide §4.9](https://cql.hl7.org/03-developersguide.html#conversion-precedence) by checking whether the proposed type conversion is satisfied.  This calculates the cost of the conversion.  

Signature matching is based on the lowest single argument maximum cost, e.g. a single argument which requires a list promotion is automatically more expensive than a function with any number of interval demotions.  When two signatures have the same maximal single argument cost, the total cost of all conversions is then considered.

It is unclear whether that approach is correct, but can be now easily changed in the InvocationBuilder's `SelectBestOverload` function.

`CallBuilders` extension methods have been eliminated in favor of `InvocationBuilder.Invoke(function, arguments)`.  The usage of `SystemLibrary.Add.Call` was nice, but with the introduction of multiple new services creates an unwieldy argument list consisting of the `InvocationBuilder`, the `TypeConverter`, and the `ElmFactory` and felt contrary to the dependency injection architecture.

- Eliminated the `GenericParameterAssignments` class in favor of an `IDictionary<string, TypeSpecifier>` instead

- Generic type matching no longer allows `T` to match lists or intervals, requiring `List<T>` and `Interval<T>` to be explicitly described instead.  

Without this change, there are insurmountable ambiguities in several system operator signatures.  To make it easier to declare these system functions, `WithListAndIntervalVariants` was added to take a function like `Equals(T,T)` and automatically create overloads for `Equals(List<T>, List<T>)` and `Equals(Interval<T>, Interval<T>)`.

- Implemented configuration options for EnableListPromotion, EnableListDemotion, EnableIntervalPromotion, and EnableIntervalDemotion.  When these options are not set, some overloads will not resolve since they will require these conversions.

And several more smaller changes.


Cql/Cql.Compiler/Cql.Compiler.csproj
```diff
@@ -26,7 +26,7 @@
 		<InternalsVisibleTo Include="CqlToElmTests" Key="$(LibraryPKHash)" />
 		<InternalsVisibleTo Include="Hl7.Cql.Packaging" Key="$(LibraryPKHash)" />
 		<InternalsVisibleTo Include="Hl7.Cql" Key="$(LibraryPKHash)" />
-		<InternalsVisibleTo Include="Hl7.Cql.CqlToElm.Test" Key="$(LibraryPKHash)"/>
+		<InternalsVisibleTo Include="CqlToElmTests" Key="$(LibraryPKHash)"/>
```
ewoutkramer: This key now appears twice?

Cql/Cql.Firely/Cql.Firely.csproj
```diff
@@ -22,7 +22,7 @@
 		<InternalsVisibleTo Include="CqlToElmTests" Key="$(LibraryPKHash)" />
 		<InternalsVisibleTo Include="Hl7.Cql.Packaging" Key="$(LibraryPKHash)" />
 		<InternalsVisibleTo Include="Hl7.Cql.Packager" Key="$(LibraryPKHash)" />
-		<InternalsVisibleTo Include="Hl7.Cql.CqlToElm.Test" Key="$(LibraryPKHash)"/>
+		<InternalsVisibleTo Include="CqlToElmTests" Key="$(LibraryPKHash)"/>
```
ewoutkramer: duplicate key?

Cql/Cql.Model/Cql.Model.csproj
```diff
@@ -28,7 +28,7 @@
 		<InternalsVisibleTo Include="CoreTests" Key="$(LibraryPKHash)" />
 		<InternalsVisibleTo Include="CqlToElmTests" Key="$(LibraryPKHash)" />
 		<InternalsVisibleTo Include="Hl7.Cql.Compiler" Key="$(LibraryPKHash)" />
-		<InternalsVisibleTo Include="Hl7.Cql.CqlToElm.Test" Key="$(LibraryPKHash)" />
+		<InternalsVisibleTo Include="CqlToElmTests" Key="$(LibraryPKHash)" />
```
ewoutkramer: idem

Cql/CqlToElmTests/NotTest.cs
```diff
@@ -153,6 +153,6 @@ public void Not_Null()
                 library NotTest version '1.0.0'
 
                 define private Not_1: not 1
-            ", "Cannot resolve call*first argument*Integer*cannot implicitly be cast*Boolean.");
+            ", "Could not resolve*");
```
ewoutkramer, line 156: I spent a lot of time producing the nice clear messages on the left, mentioning the incorrect argument.

Cql/Cql.CqlToElm/InvocationBuilder.cs
```diff
@@ -27,270 +22,265 @@ internal record ResolveResult<T>(T Result, int Cost, string? Error)
...
             }
+            else if (type is ListTypeSpecifier listType)
+                return ReplaceGenericTypes(listType.elementType, replacements).ToListType();
+            else if (type is IntervalTypeSpecifier intervalType)
+                return ReplaceGenericTypes(intervalType.pointType, replacements).ToIntervalType();
+            else return type;
+        }
 
-            // TODO: interval demotion https://cql.hl7.org/03-developersguide.html#promotion-and-demotion
+        internal ConversionResult<Expression> TryGenericReplacement(TypeSpecifier operandType, Expression argument, IDictionary<string, TypeSpecifier> replacements)
```
ewoutkramer: This function is called TryXXX, but does not stick to the classic Try pattern (return bool, out param) - is this intentional?

Cql/Cql.CqlToElm/ModelProvider.cs
```diff
@@ -276,15 +282,18 @@ internal static Elm.TypeSpecifier GetTypeSpecifierForElement(this ClassInfoEleme
         /// </summary>
         /// <remarks>This function checks a subtyping relationship, not whether in- or explicit casts may exist between the types.
         /// See https://cql.hl7.org/03-developersguide.html#type-testing for more information.</remarks>
-        public static bool IsSubtypeOf(this Elm.TypeSpecifier subType, Elm.TypeSpecifier superType, IModelProvider provider)
+        internal static bool IsSubtypeOf(this Elm.TypeSpecifier subType, Elm.TypeSpecifier superType, IModelProvider provider)
         {
-            if (superType == subType || superType == SystemTypes.AnyType)
```
ewoutkramer, line 281: The `== SystemTypes.AnyType` was meant to speed up this function - as long as we can assume any type is a subclass of Any, which I think it is by definition?

Cql/CqlToElmTests/NotTest.cs
ewoutkramer: But I guess the intent is to stay one on one with the Java output?

Cql/Cql.CqlToElm/TypeConverter.cs
```diff
@@ -0,0 +1,426 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Hl7.Cql.CqlToElm
+{
+    /// <summary>
+    /// Implements type conversion according to the rules defined by the specification. 
+    /// </summary>
+    /// <seealso href="https://cql.hl7.org/03-developersguide.html#conversion-precedence"/>
+    internal class TypeConverter
```
ewoutkramer: Can we give this another name?  There is already a TypeConverter in the project, and I think this class isn't converting per se, as well as building ELM expressions that do the conversion.  ConversionBuilder?

Cql/Cql.CqlToElm/TypeConverter.cs
```diff
@@ -0,0 +1,426 @@
...
+                return ConversionCost.ListPromotion;
+            else
+                return ConversionCost.Incompatible;
+        }
+
+        internal string IncompatibilityMessage(TypeSpecifier from, TypeSpecifier to) =>
+            $"Expression of type '{from}' cannot be cast as a value of type '{to}'.";
+
+        // All types exactly match Any
+        internal virtual bool IsExactMatch(TypeSpecifier from, TypeSpecifier to) =>
```
ewoutkramer: Virtual?  Is that added complexity warranted?

Cql/Cql.CqlToElm/TypeConverter.cs
```diff
@@ -0,0 +1,426 @@
...
+            from == to;
+
+        internal virtual bool IsSubtype(TypeSpecifier from, TypeSpecifier to)
```
ewoutkramer: ModelProvider also has an SubTypeOf, is it necessary to duplicate this functionality?

EvanMachusak:
PR should be good to merge now, @ewoutkramer

ewoutkramer (changes requested):

Cql/CqlToElmTests/NotTest.cs
EvanMachusak: Yes, I tried to match the exact messages that the cql-to-elm reference implementation does.  The messages are mostly the same, but it's phrased "Could not" instead of "cannot".

Cql/Cql.Compiler/Cql.Compiler.csproj
EvanMachusak: For the life of me I could not get `<InternalsVisibleTo>` working when I had `Hl7.Cql.CqlToElm.Test` specified as that project's `<AssemblyName>`.  I'll clean those up.

Cql/Cql.CqlToElm/InvocationBuilder.cs
EvanMachusak: Renamed to `ReplaceGeneric*` instead of `Try*`.  The rationale was that the `ConversionResult` type has a `Success` property on it (along with other things), but I agree with you on the expected usage associated with the word `Try`.

Cql/Cql.CqlToElm/ModelProvider.cs
EvanMachusak: Initially, my thinking that `IsSubType` from the model's perspective should only return `true` when there actually is an unbroken inheritance chain from the type back to the `Any` type declared in the `elm-modelinfo-r1.xml`.  Neither `List` nor `Interval` are actually declared in any model.  I was not sure if argument-to-operand binding during signature resolution was matching based on `List` and `Interval` using `Subtype` conversion precedence, or a different kind.

Bryn later clarified in Zulip that the above logic is correct: *every* type should be considered a subtype of Any, regardless of whether it's in the model or not.

The spec also says this:
`the term subtype applies to all derived types, as well as the type itself.`

I introduced a test called `TypesAreNotSubtypesOfThemselves` that I think violates the spec language [here](https://cql.hl7.org/03-developersguide.html#type-testing), saying: 

>Note that because of the identity relationship above, the term subtype applies to all derived types, as well as the type itself. In the discussions that follow, if a definition must explicitly refer to only derived types, the term proper subtype will be used.

I think I will rather rename this current implementation `IsProperSubtype`, including the check against Any, and then implement `IsSubType` that check if `subType == superType || IsProperSubtype`.

I am going to allow `Any` to be a proper subtype of `Any`.

All tests will pass with these changes.

Cql/Cql.CqlToElm/TypeConverter.cs
EvanMachusak: I asked ChatGPT:
> User
what is the process of changing one argument's type to another when binding an argument to an operand?

> The process of changing one argument's type to another when binding an argument to an operand in programming languages, particularly in the context of function calls or operations, is known as **type coercion** or **type conversion**. This process can be either implicit or explicit.

Since `TypeConverter` is taken on the C# side, I'll use `Coerce` as the verb instead of `Convert`.  Generally, when I name classes, particularly injected services I think about the main verb.  In this case the key function is called `Convert` (to be renamed `Coerce`).  I really don't want to use the word `Coercer` so I'll call it `CoercionProvider`.

Cql/Cql.CqlToElm/TypeConverter.cs
EvanMachusak: I felt like we may encounter scenarios where the only way to support certain CQL files was to intentionally subvert a small part of the coercion algorithm so we could match a buggy implementation's behavior, while preserving the rest of the behavior.  But since these methods are internal we can just make them virtual when and if that happens, or we can just create an interface for this service and inject that interface instead of the implementation.  I'll remove it.

Cql/Cql.CqlToElm/TypeConverter.cs
EvanMachusak: The functionality in this class works and passes all tests.  The implementation in `ModelProvider` does not.

We need to review the usage in general around extension methods.  I have a distaste for this pattern:

`from.IsSubtypeOf(to, ModelProvider);`

If we could have made it:

`from.IsSubtypeOf(to)`

Then that is very good usage.  

But if the core functionality of this method has to be implemented through `ModelProvider`, then I find it much more preferable to make the call pattern like this:

`ModelProvider.IsSubType(from, to)`

Generally, once you have to provide a service that implements the functionality as a parameter to the extension method, it shouldn't be an extension method anymore.

EvanMachusak: Here's an example test that `ModelProvider`'s `IsSubType` chokes on:
```
        [TestMethod]
        public void IntervalIncludesNull()
        {
            var arguments = new Expression[] { Interval(Integer(1), Integer(10)), Null };
            var result = InvocationBuilder.MatchSignature(SystemLibrary.Contains, arguments);
            result.Compatible.Should().BeTrue();
            result.Arguments[0].Result.Should().HaveType(SystemTypes.IntegerType.ToIntervalType());
            result.Arguments[1].Result.Should().HaveType(SystemTypes.IntegerType);
         }
```
This test matches how the cql-to-elm reference engine produces ELM for a CQL expression of:
`Integer[1,10] contains null`

When using the `ModelProvider`'s implementation, this code gets hit:
```
            else if (subType is Elm.IntervalTypeSpecifier intervalSub && superType is Elm.IntervalTypeSpecifier intervalSuper)
            {
                // covariance on intervals
                return intervalSub.pointType.IsSubtypeOf(intervalSuper.pointType, provider);
            }
```

The current algorithm for inferring generics is to see which inference is cheaper, so we try inferring both `Integer` and `Any` for `Contains(Interval<T>, T)`.

When we infer `T` to be `Any`, we now have a subtype relationsip where `Interval<Integer>` is a subtype of `Interval<Any>` (precedence = 2).  This is cheaper than treating the `null` to be an `Integer` (compatible, precedence = 3), so the conversion precedence requires we pick the cheaper match and we change the first argument, `Interval[1,10]`, to be an `Interval<Any>`.

At the same time, this is true in the reference engine:
`Interval[1,10] is Interval<Any>`.
But false in ours, and that makes sense, because it is tantamount to this:
```
var intInterval = new CqlInterval<int?>(1,10, true, true);
(intInterval is Interval<object>).Should().BeFalse();
```
The C# compiler raises an error because Interval is a concrete type, not an interface or a delegate and can't be covariant.

The `ModelProvider`'s implementation of `IsSubType` is, I believe, correct.  You can find specification language that justifies every line of code in that method.

I have been tearing my hair out the last week trying to create a set of rules that both universally follow the specification but also pass all the tests.  That is impossible.

I have considered that maybe what we should be doing is only inferring the generic on the first argument, and if any acceptable match is found, use that.  But the following expressions are all legal in CQL:

```
define Contains1: Interval[1, 10] contains null
define Contains2: null contains 5
define Contains3: Interval[1 as Any, 10 as Any] contains (5 as Any)
```
`Contains2` infers the `null` operand to be an `Interval<Integer>`.  If the reference implementation used the approach of accepting the first matching generic inference instead of the overall cheapest, it would infer T to be `Any` and then cast the `5` to `Any`, resulting in the legal expression `Contains3`.  Note that `Contains2` is false and `Contains3` is `true` in the reference engine.  So adopting the `Contains3` expression using a different generic inference algorithm would actually produce a different runtime result.

ewoutkramer: We need it to look at all arguments, otherwise the Add would not work either -   `Add(4.0,4)` would select `Add(double,double)` as the overload, which is fine, but `Add(4,4.0)` would select the `Add(integer,integer)` overload, which would fail. That was the original reason I had to consider all arguments for assignment of generic parameters.

Issue #131, opened 2023-12-18
Verify we have implemented missing converters for tuples in modelinfo


EvanMachusak:
See #129 

We should create some test code that:

- Dynamically generates `Copy T` functions for all `T` of type `ClassInfo` present in a `ModelInfo` class which creates a new named tuple and assigns all of its properties to `obj`'s value, e.g.:
```
define function "Copy Patient"(obj FHIR.Patient):
  FHIR.Patient {
     id: obj.id,
     birthDate: obj.birthDate,
     ... all other elements of Patient follow ...
}
```
- Fix the cryptic `ArgumentException` that gets thrown when a conversion required for the member binding is missing, with information about which conversion is being attempted and not available
- Add all missing converters for the FHIR model

This will allow us to be sure that we can dynamically create FHIR model tuples in CQL.

baseTwo:
Closing as done - all three items in this issue are implemented on `develop`, and I verified them by running the tests.

**1. "Copy T" functions for every `ClassInfo` in the model**

[`Cql/CoreTests/Input/ELM/HL7/FHIRConversionTest.cql`](Cql/CoreTests/Input/ELM/HL7/FHIRConversionTest.cql) is exactly the test asked for: 3,155 lines containing 143 `define function "<Resource>Resource"(arg <Resource>)` definitions - one for every FHIR R4 resource from `Account` through `VisionPrescription` - each of which reconstructs the resource by assigning all of its top-level elements. It is compiled end-to-end by `CoreTests.CodeBuilderTests.FHIRConversionTest_1_0_0` (Cql/CoreTests/CodeBuilderTests.cs:45).

There is also `CodeBuilderTests.Medication_Request_Example_Test` (Cql/CoreTests/CodeBuilderTests.cs:70), annotated in the source with `// https://github.com/FirelyTeam/firely-cql-sdk/issues/129`, which compiles the exact `MedicationRequest { ... }` CQL from #129 (Cql/CoreTests/Input/ELM/Test/Medication_Request_Example.cql).

**2. The cryptic `ArgumentException` is fixed**

[...]

PR #192, opened 2024-02-21
Add missing type conversion for FHIR.Code needed for MeasureReport resource

igajurelNCQA:
Added conversion for FHIR.Code type


igajurelNCQA:
Added conversion for type FHIR.Code

This enables us to convert CQL we have authored which returns MeasureReport resources, internally containing FHIR.Code, currently failing to be converted in the firely cql engine.

_partial_ fix for https://github.com/FirelyTeam/firely-cql-sdk/issues/131

EvanMachusak (approved):

PR #203, opened 2024-02-23
1.1/202 cleanup

baseTwo:
## Work Done
* DirectedGraph did not have centralized state for edges, they were duplicated between the graph and the nodes. Changed to only keep track of edges inside the graph as a Dictionary<string, HashSet<DirectedGraphNode>> 
* DirectedGraphNode was changed from class to struct.
* Code duplication was reduced: DetermineBuildOrder, and graph merging.
* Dead code that was removed.

## Visualizing DirectedGraph
It is possible to show the directed graph by inspecting the `GraphvizDiagram` property on the graph, or just copying the `GraphvizDiagramUri` to the browser to view it visually. Nodes are labeled by their node id and topologically sorted ordinal.

An example is shown below
image


Cql/Graph/DirectedGraphEdge.cs
```diff
@@ -24,21 +26,20 @@ internal class DirectedGraphEdge
 
         public DirectedGraphEdge(string fromId, string toId)
         {
+            FromId = DirectedGraphNode.StartId;
```
EvanMachusak: Copy paste error?

baseTwo: This file was removed. Edges are now maintained internally inside the DirectedGraph.

PR #205, opened 2024-02-26
1.1/166 Writing C# files up until the last successful package

baseTwo:
Related to Issue #166 

Currently C# and FHIR resources are written only after successfully processing all cql and elm files. This all-or-nothing operation makes it difficult to know how far we are in attempting to fix bugs on a new set of data.

As part of this PR, more classes were updated so that dependencies are injected through contructors, rather than passed along methods or indirectly through other class members. See **Service Dependencies** below for the graph how it all fits together.

Observe that the `PackagerCliProgram` includes a dependency to `LibraryPackagerService` and also the `ResourceWriter`'s that write out the compiled FHIR resources to C-Sharp and FHIR directories. The problem was that all resources were generated first, and then written out in the end after all inputs were processed successfully.

The change involves putting in a catch when enumerating the input elm libraries in `LibraryPackagerService.PackageResources`. When an exception occurs, they are wrapped up in a `ExceptionDispatchInfo` to be rethrown at a later stage. So all files that were processed successfully up to this point into resources, are written out to the resource writers, via the callback provided for it in `LibraryPackageCallbacks._onLibraryResourceCreated`. 

This callback is handled in `PackagerCliProgam.RunCore' whenever a resource is created. There are also resources remaining afterwards for the Tuple types, and they are written out just before the end of the method.

## Service Dependencies (excluding ILogger and IOptions)
```mermaid

classDiagram
    direction LR

    %% Inheritance  

    %% Injected Dependencies
    ExpressionBuilderService ..> AssemblyCompiler : injected
    TypeResolver ..> AssemblyCompiler : injected

    OperatorBinding ..> ExpressionBuilderService : injected
    TypeManager ..> ExpressionBuilderService : injected

    IEnumerable_ResourceWriter_ ..> PackagerCliProgram : injected 
    LibraryPackager ..> PackagerCliProgram : injected 

    TypeResolver ..> TypeManager : injected (Fhir)

    TypeResolver ..> LibraryPackager : injected (Fhir) 
    ExpressionBuilderService ..> LibraryPackager : injected 
    
    TypeResolver ..> CSharpSourceCodeWriter : injected (Fhir)

    %% Rest
    
```


Cql/Cql.Packaging/LibraryPackagerService.cs
```diff
@@ -0,0 +1,394 @@
+using System.Runtime.ExceptionServices;
```
ewoutkramer: Isn't this the LibraryPackager?  I mean specifically the name: why would we add the Service suffix here?  For consistency: when would we add Service and when don't we do that?

Cql/Cql.Packaging/ResourcePackageArgs.cs
```diff
@@ -2,15 +2,15 @@
 
 namespace Hl7.Cql.Packaging;
 
-/// <summary>Arguments for the <see cref="ResourcePackager.Package(PackageArgs)"/> method.</summary>
-public record PackageArgs
+/// <summary>Arguments for the <see cref="ResourcePackager.Package(ResourcePackageArgs)"/> method.</summary>
+public record ResourcePackageArgs
```
ewoutkramer, line 6: Are there other kinds of package args that we forsee: if so, this is wise. If not, is the rename worth changing the public surface for?

Cql/Cql.Packaging/ResourcePackager.cs
```diff
@@ -64,35 +70,29 @@ public ResourcePackager(ILoggerFactory logFactory, params ResourceWriter[] resou
         {
             if (resourceWriters.Length == 0) return; //Skip since no writers provided
 
-            var packages = LibraryPackager.LoadLibraries(elmDir);
-            var graph = packages.Values.GetIncludedLibraries();
-            var typeResolver = new FhirTypeResolver(ModelInfo.ModelInspector);
+            ModelInspector modelInspector = ModelInfo.ModelInspector;
```
ewoutkramer, line 73: Can we use var? Was this an accidental rewrite?

Cql/Cql.Packaging/ResourcePackager.cs
```diff
@@ -64,35 +70,29 @@ public ResourcePackager(ILoggerFactory logFactory, params ResourceWriter[] resou
...
-                resource => CanonicalUri(resource, resourceCanonicalRootUrl),
-                logFactory);
+            IDictionary<string, Library> librariesByNameAndVersion = LibraryLoader.LoadLibraries(elmDir);
+            DirectedGraph directedGraph = librariesByNameAndVersion.Values.GetIncludedLibraries();
+            IEnumerable<Resource> resources = resourcePackagerService.PackageResources(elmDir, cqlDir, directedGraph, callbacks);
 
             afterPackageMutator?.Invoke(resources);
 
-            foreach (var writer in resourceWriters)
+            foreach (ResourceWriter resourceWriter in resourceWriters)
```
ewoutkramer: I guess so ;-)

Cql/CodeGeneration.NET/AssemblyCompiler.cs
```diff
@@ -24,85 +23,104 @@
...
+        private readonly ExpressionBuilderService _expressionBuilderService;
+        private readonly Factory<CSharpSourceCodeWriter> _cSharpSourceCodeWriterFactory;
+        private readonly Lazy<Assembly[]> _referencesLazy;
 
-        internal IDictionary<string, AssemblyData> Compile(IEnumerable<Library> elmPackages,
-                    ILoggerFactory logFactory)
+        public AssemblyCompiler(
+            ExpressionBuilderService expressionBuilderService,
+            [FromKeyedServices("Fhir")] TypeResolver typeResolver,
+            Factory<CSharpSourceCodeWriter> cSharpSourceCodeWriterFactory,
```
ewoutkramer: Why don't we inject  a CSharpSourceCodeWriter - do we need this Factory<>?

Cql/CodeGeneration.NET/Factory.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Hl7.Cql.CodeGeneration.NET;
+
+internal abstract class Factory<T>
```
ewoutkramer: I've looked at this code and I don't get what it does.  It's an implicit cast of some value of T to a factory<T> that would then produce this value.  Why don't we just provide the value?

Cql/PackagerCLI/DependencyInjection.cs
```diff
@@ -78,4 +84,54 @@ private static void TryAddConfiguredResourceWriters(IServiceCollection services,
             services.TryAddEnumerable(resourceWritersServiceDescriptors);
         }
     }
+
+
+    public static void TryAddCompilationServices(this IServiceCollection services)
+    {
+        services.TryAddSingleton<OperatorBinding, PackagerCqlOperatorBinding>();
+        services.TryAddSingleton<CSharpSourceCodeWriter>();
```
ewoutkramer, line 92: So, her we do inject it as a singleton - that seems to make sense to me.  But then, why is the Factory<CsharpSourceCodeWriter> there?

Cql/Cql.Packaging/ResourcePackager.cs
baseTwo: Probably, will fix

ewoutkramer:
In general: I am a fan of having clear responsibilities and classes that have clear purpose, so I agree with lots of the restructuring that you are doing, but I am wondering about when to make stuff into a service and when not to. Clearly not everything is a service. What would be the "lower boundary" for which to turn a class into a service?  E,g. why does the CSharpSourceCodeWriter need to be injected?  Do we want to be able to insert different kind of CSharpSourceCodeWriters?  And e.g. what about the ExpressionBuilder?

Cql/Cql.Packaging/ResourcePackager.cs
baseTwo: I had a look, this was intentional. I want to the types involved to construct the dependencies needed to build the packages to be explicit.

Cql/CodeGeneration.NET/AssemblyCompiler.cs
baseTwo: I will move the AssemblyCompiler.BuildCSharpSourceCodeWriter method into the CSharpSourceCodeWriter, then it can be injected

Cql/PackagerCLI/DependencyInjection.cs
baseTwo: Well spotted, with the factory pattern this singleton was never used. I removed the factory pattern and inject the type resolved into the CSharpSourceCodeWriter, so that it can correctly build the usings and aliases. So no need for the `AssemblyCompiler.BuildCSharpSourceCodeWriter` method

Cql/CodeGeneration.NET/Factory.cs
baseTwo: Removed this pattern, no longer needed.

Cql/Cql.Packaging/ResourcePackager.cs
baseTwo: Yes, my refactoring for the code lines above that explicitly change var to types, should not be necessary for this code. Reverted back to var

Cql/Cql.Packaging/ResourcePackageArgs.cs
baseTwo: This renamed was done because there were already both a `LibraryPackager` and a `ResourcePackager`. So `PackageArgs` is a bit ambiguous here. Now at least it's clear it belongs to the latter when passed in to `ResourcePackager.Package`. Happy to revert if it is a big issue on the public api??

baseTwo:
> In general: I am a fan of having clear responsibilities and classes that have clear purpose, so I agree with lots of the restructuring that you are doing, but I am not used to splitting up relatively small programs into a host of injectable services. What would be the "lower boundary" for which to turn a class into a service? So, just to understand the decisions made here: why does the CSharpSourceCodeWriter need to be injected? Do we want to be able to insert different kind of CSharpSourceCodeWriters? And e.g. what about the ExpressionBuilder?

ExpressionBuilder is a bit of a special case, it's quite massive but it still has a single responsibility, which is to build expressions. The idea behind splitting classes into smaller types is so that they keep to a single responsibility. 

Taking this CSharpSourceCodeWriter as an example, When do we want to inject something else? Well, we don't have sufficient tests in place yet, but when they are added, this code be swapped out for a mocked service.

Cql/Cql.Packaging/ResourcePackager.cs
ewoutkramer: What was the reason you are newing up the dependencies here yourself, and not let them be injected?

ewoutkramer (approved):

PR #211, opened 2024-02-26
Fix issue #210: ExpressionBuilder.GetProperty not using TypeResolver

### LINKED ISSUE #210: ExpressionBuilder skips the TypeResolver in some cases
When handling Tuple and Instance ELM nodes, a helper [function called `GetProperty` is used.](https://github.com/FirelyTeam/firely-cql-sdk/blob/9d3cbd97369d6973b4e5ce2ddea6c62c3d074a32/Cql/Cql.Compiler/ExpressionBuilder.cs#L2543) This helper unwraps the `Value` property from `Nullable`, but then uses reflection directly to locate the property:

` var member = type.GetProperty(name, BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance) ?? throw new ArgumentException($"Unknown property {name} on type {type}.");`

This is incorrect, because it circumvents the `TypeResolver` and fails to resolve properties like `id` on `Element` subtypes, which is called `ElementId` in the Firely SDK.

The fix is to replace the use of reflection with a call into `TypeResolver`.

EvanMachusak:


EvanMachusak:
Paul - in 1.1 we should consider eliminating this method by moving the Nullable value unwrapping into the FhirTypeResolver class in the Cql.Firely project and replacing calls to this helper method to `TypeResolver.GetProperty` directly.

Cql/Cql.Compiler/ExpressionBuilder.cs
```diff
@@ -2540,7 +2540,7 @@ protected Expression ParameterRef(ParameterRef op, ExpressionBuilderContext ctx)
 
         }
 
-        protected MemberInfo GetProperty(Type type, string name)
+        protected internal MemberInfo GetProperty(Type type, string name)
```
baseTwo, line 2543: Any reason this is opened up to internal? I noticed the test, but perhaps it could be put on the TypeResolver itself.

baseTwo (approved):

baseTwo:
Created issue #212 as a result from a review comment

PR #213, opened 2024-02-26
Casting issue

gshakya-ncqa:
Resverse assignabilty check. 
Checked **type** is assignable to **Operand.type**


gshakya-ncqa:
When converting code from ELM to C#, there were a lot of warnings saying "Potentially unsafe cast from". This happened, for example, when a date object was set to null. However, setting a date to null isn't actually unsafe. The warnings were showing up because of a mix-up in comparing how things could be assigned. After fixing this, the warning now shows up correctly when there's a real problem.

baseTwo (approved):

PR #214, opened 2024-02-26
Adding conversions

gshakya-ncqa:
Partial fix for #131 
I have created CQL functions to generate a Resource for each FHIR Resource type. These functions ensure that all necessary type conversions are tested.
Specifically, I've added type conversions for the following
- Markdown
- Instant
- FhirUrl

However, functions for certain resources have been temporarily commented out. These include CapabilityStatement, CompartmentDefinition, ExampleScenario, GraphDefinition, ImplementationGuide, MessageDefinition, OperationDefinition, and Questionnaire.

The reason for this is that these resources contain a ResourceType element. The CapabilityStatement.Rest.Resource.Type element does not resolve to Code<ResourceType>, leading to inconsistent test results. Occasionally, the FHIR.ResourceType element is resolved to Code, causing the test to fail intermittently.


Cql/Cql.Firely/FhirTypeConverter.cs
```diff
@@ -60,6 +61,19 @@ internal static TypeConverter ConvertFhirToCqlPrimitives(this TypeConverter conv
             add((M.FhirString p) => p.Value);
             add((M.FhirBoolean p) => p.Value);
             add((M.FhirDecimal p) => p.Value);
+            add((M.Markdown p) => p.Value);
+            add((M.Instant p) => p.Value);
+            add((M.Instant p) =>
```
baseTwo, line 66: I would recommend using a switch expression

baseTwo (approved):
Looks good, left one small comment, but approving nonetheless.

ewoutkramer:
We can't seem to update it to the latest version of the dev branch, @gshakya-ncqa can you remerge and push?

gshakya-ncqa:
Remerged and Pushed

PR #218, opened 2024-02-27
1.1/215 Create one solution containing all projects (Cql + Demo) and one only filtered on Cql projects

baseTwo:
Related to #216 

* Fixed the Demo solution by adding missing projects
* Renamed `Demo.sln` -> `CqlAndDemo.sln` (basically all projects in this repo)
* Replaced `Cql.sln` -> `Cql.slnf` (which is a filtered solution based on `CqlAndDemo.sln` but excluding the Demo projects)
* Update build pipeline to use the `Cql.slnf` as solution file, otherwise it defaults to `CqlAndDemo.sln`
* Discover some tests in the Demo solution which did not compile, because public code was removed. 
* Added the `JetBrains.Annotations` nuget package in the solution. (This is only to enhance developer experience in the IDE and isn't a dependency after compilation). The `[UsedImplicitly]` attribute tells the ReSharper / Rider IDE that a particular piece of code should not be considered unused. Ideally we should create tests to cover the full public API, until then, this will have to do.
* Reintroduce the public code that was deleted from the develop branch and attribute with `[UsedImplicitly]` :
  * `GetProperty` and `Lambda` on `ExpressionBuilderService`
  * `GetDependencies` and `AllLibraries`  in `LibraryExtensions` in Packing project
  * `GetIncludedElmLibraries` in `LibraryExtensions` in Elm project
  * `LoadLibraries`, `LoadResources` and `LoadElm` in `LibraryPackager`
* There was quite a lot of boilerplate code to create the all the required dependencies for the `ExpressionBuilder` or `LibraryPackager` classes. Introduced `ExpressionBuilderFactory` and `LibraryPackager` classes respectively to simplify this. This is internal for now, as an alternative to the .net host IServiceProvider. Do not use these two factories as injectables themselves!


Demo/Test/Test.csproj
```diff
@@ -27,7 +27,18 @@
     <ProjectReference Include="..\Measures\Measures.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
```
ewoutkramer, line 30: Is this necessary? We'll never release the demo?

ewoutkramer (approved):

Demo/Test/Test.csproj
baseTwo: Probably not, but this is just the standard stuff when creating a new project. We could take it out, but doesn't harm leaving it here

Issue #229, opened 2024-03-06
Improve exception information in the logs while building library defitions


baseTwo:
Improve exception information when library definitions building fails, such that information is clear where the exception was caused:

* The library name, version and ordinal 
* The expression definition name and ordinal
* The expression paths

Where relevant the types and locators need to be included.

Create an exception type for this which must be thrown during expression building.

PR #230, opened 2024-03-06
1.1/229 improve exception detail

baseTwo:
PR for issue #229 

Throw `ExpressionBuildingException` exceptions when building expressions. The `ExpressionBuilderContext` is passed in to this exception, so that it can build a descriptive message from it with the expression path up to the point where the exception occurred. This is possible due to `ExpressionBuilderContext` keeping a reference to the current elm `Element` and to outer `ExpressionBuilderConext`s.

The `OperatorBinding` dependency was moved from `ExpressionBuilder` to `ExpressionBuilderContext`, where it is decorated to catch exceptions and rethrow them as `ExpressionBuildingException`.

Here is what exception output looks like now in the log:
```
System.InvalidOperationException: Failed building the 2 of 66 libraries 'Status-1.6.000'. See InnerException for more details.
 ---> Hl7.Cql.Compiler.ExpressionBuildingException: Exception while building the 31th of 40 expression definition 'BloodPressure ' in library 'Status-1.6.000' at location '193:1-195:57'. See InnerException for more details.
 ---> Hl7.Cql.Compiler.ExpressionBuildingException: Cannot convert System.Nullable`1[Hl7.Fhir.Model.ObservationStatus] to Hl7.Fhir.Model.Code`1[Hl7.Fhir.Model.VerificationResult+StatusCode] (Parameter 'source') Expression Path:
-  FunctionDef named 'Patient.BloodPressure' of type 'List<{http://hl7.org/fhir}Observation>' at location '193:1-195:57'
-  Query of type 'List<{http://hl7.org/fhir}Observation>' at location '194:3-195:57'
-  In of type '{urn:hl7-org:elm-types:r1}Boolean' at location '195:5-195:57'
-  Property named 'value' of type '{http://hl7.org/fhir}Status' at location '195:11-195:18'
 ---> System.ArgumentException: Cannot convert System.Nullable`1[Hl7.Fhir.Model.ObservationStatus] to Hl7.Fhir.Model.Code`1[Hl7.Fhir.Model.VerificationResult+StatusCode] (Parameter 'source')
   at Hl7.Cql.Compiler.CqlOperatorsBinding.BindConvert(MemberExpression operators, Expression source, Expression 

   ... rest of stack trace ...
```

Other:
* Small other fix, the PackageCLI log output file name was changed from build.txt to build.log. 
* Demo.sln is deleted, because the CqlAndDemo.sln replaced it in a previous PR.


Cql/Cql.Compiler/ExpressionBuilderService.cs
```diff
@@ -28,22 +28,22 @@ internal class ExpressionBuilderService
     public DefinitionDictionary<LambdaExpression> BuildLibraryDefinitions(Library library) => 
         ExpressionBuilder.BuildLibraryDefinitions(_operatorBinding, _typeManager, _logger, library);
 
-    public MemberInfo GetProperty(
+    public MemberInfo? TryGetProperty(
```
ewoutkramer: I like to keep following the usual TryXXXX pattern - certainly in a public interface, so this should then return a bool & have an out parameter!

Cql/Cql.Compiler/Infrastructure/EnumerableExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// Pair items in an enumeration with their ordinals starting at 1.
+    /// </summary>
+    public static IEnumerable<(T Item, Ordinal Ordinal)> WithOrdinals<T>(this IEnumerable<T> source)
+    {
+        int i = 0;
+        foreach (var item in source)
+        {
+            ++i;
+            yield return (item, new Ordinal(i));
```
ewoutkramer: `source.Select((x,ordinal) => (x,new Ordinal(ordinal)))`?

Cql/Cql.Compiler/TypeManager.cs
```diff
@@ -120,10 +121,10 @@ public TypeManager(TypeResolver resolver, string assemblyName = "Tuples", string
                 {
                     Type? sourceType = null;
                     if (propertyExpression.source != null)
-                        sourceType = TypeFor(propertyExpression.source!, context);
+                        sourceType = TypeFor(propertyExpression.source!, ctx.Deeper(propertyExpression.source));
```
ewoutkramer, line 124: That looks like a real change in the moment when ctx.Deeper is called - is this on purpose?

ewoutkramer (changes requested):

Cql/Cql.Compiler/TypeManager.cs
baseTwo: Yes, this is on purpose. Exceptions thrown in TypeFor must be thrown such that the `source` element is dumped as part of the exception message.

Cql/Cql.Compiler/ExpressionBuilderService.cs
baseTwo: I'm going to change it back to the original name. It should be clear from the nullable return type that the result can be null

ewoutkramer (approved):

PR #231, opened 2024-03-07
1.1/Extract LibraryExpressionBuilder, Move state from Builders into Contexts.

baseTwo:
Related to issue #232  *This PR is setting the groundwork for cross library contexts, it will be done in two parts.*

* State moved from the Builders into Contexts, such that the builders only keep dependencies to utility types.
* Extract expression building on library-level into separate classes `LibraryExpressionBuilder`, `LibraryExpressionBuilderContext` and `LibraryExpressionBuilderFactory`
* PackageCLI was building ELM libraries twice, removed the first pass as it was wasteful. Unfortunately this also means writing out c# and fhir resource files on the fly is no longer possible. This whole process needs some rethink. 
* `LibraryExpressionBuilderContext` and `ExpressionBuilderContext`can be stacked such that the path from library element to current element being processed can be traced. This is especially useful when throwing exceptions. 
* Changed keyed DI services to plain ones (removed `[FromKeyedServices("Fhir")]`)

## Service Dependencies (excl Logger and Options)

The README has been updated in the PackageCLI  and a diagram was added showing the service dependencies.

```mermaid
classDiagram
    direction LR

    class TypeManager {
    }

    %% Inheritance  

    %% Injected Dependencies
    LibraryExpressionBuilder ..> AssemblyCompiler : injected

    TypeManager ..> ExpressionBuilder : injected

    IEnumerable_ResourceWriter_ ..> PackagerCliProgram : injected 
    LibraryPackager ..> PackagerCliProgram : injected 

    TypeResolver ..> LibraryPackager : injected
    LibraryExpressionBuilder ..> LibraryPackager : injected 
    
    TypeResolver ..> CSharpSourceCodeWriter : injected

    OperatorBinding ..> LibraryExpressionBuilder : injected

    %% Rest
```


Cql/CodeGeneration.NET/AssemblyCompiler.cs
```diff
@@ -63,26 +61,28 @@ internal class AssemblyCompiler
...
         {
-            var all = new DefinitionDictionary<LambdaExpression>();
-            foreach (var package in elmPackages)
+            foreach (var (elmLibrary, ordinal) in elmLibraries.WithOrdinals())
             {
-                var expressions = _expressionBuilderService.BuildLibraryDefinitions(package);
-                all.Merge(expressions);
+                var packageDefinitions = new DefinitionDictionary<LambdaExpression>();
+                var libctx = _libraryExpressionBuilder.CreateContext(packageDefinitions, elmLibrary, ordinal);
+                _libraryExpressionBuilder.ProcessLibrary(libctx);
```
ewoutkramer: It's now implicit how built expressions are ending up in `packageDefinitions`, I assume that by passing `packageDefinitions` into the context, and then running `ProcessLibrary` on that context, the expressions are added to the `packageDefinitions`, but is there any way to make that more explicit?

Cql/CoreTests/ExpressionBuilderTests.cs
```diff
@@ -1,45 +1,59 @@
...
+        private static LibraryExpressionBuilderFactory Factory = new(LoggerFactory);
 
         [TestMethod]
         public void AggregateQueries_1_0_0()
         {
             var elm = new FileInfo(@"Input\ELM\Test\Aggregates-1.0.0.json");
             var elmPackage = Hl7.Cql.Elm.Library.LoadFromJson(elm);
-            _ = NewExpressionBuilder().BuildLibraryDefinitions(elmPackage);
+            var definitions = new DefinitionDictionary<LambdaExpression>();
+            var libctx = Factory.LibraryExpressionBuilder.CreateContext(definitions, elmPackage, Ordinal.Unspecified);
```
ewoutkramer: The ordinals are passed in for a sequential number in the exception message, but it does pollute the CreateContext function a bit now I think - if you have the name of the offending library in the exception message, do we really need the ordinal number?

Cql/CoreTests/ExpressionBuilderTests.cs
```diff
@@ -1,45 +1,59 @@
...
+            Assert.IsTrue(definitions.Libraries.Any());
         }
 
         [TestMethod]
         public void QueriesTest_1_0_0()
         {
             var elm = new FileInfo(@"Input\ELM\Test\QueriesTest-1.0.0.json");
             var elmPackage = Hl7.Cql.Elm.Library.LoadFromJson(elm);
-            _ = NewExpressionBuilder().BuildLibraryDefinitions(elmPackage);
+            var definitions = new DefinitionDictionary<LambdaExpression>();
```
ewoutkramer: It looks a bit repetitive over the last 4 tests. worth avoiding it here?

ewoutkramer (changes requested):

ewoutkramer:
An impressive set of work. I think the separation of concerns between building a package (LibraryExpressionBuilder) and building expressions (ExpressionBuilder) in that package make sense.

Cql/CodeGeneration.NET/AssemblyCompiler.cs
baseTwo: I will improve on this in the second continuation of cross library contexts. There will still be some shuffling around of state at the top and library level

Cql/CoreTests/ExpressionBuilderTests.cs
baseTwo: It applies on not just libraries, but any element in a list. I could remove this entirely and think about dealing with List elements in another way.

Cql/CoreTests/ExpressionBuilderTests.cs
baseTwo: I'll rework this

EvanMachusak:
With regard to:
> PackageCLI was building ELM libraries twice, removed the first pass as it was wasteful. Unfortunately this also means writing out c# and fhir resource files on the fly is no longer possible. This whole process needs some rethink.

The easiest way to think about how this process works is to liken it to the way C and C++ compilation works.  In those languages, you can *compile* a single file at a time.  You must then *link* all of those objects into a final library.

We can build ELM files intoLambdas (or Delegates), and even write their C# source code, by themselves, one at a time.  You can create a `Library` resource that has attachments of the CQL, the ELM, and the C# source code file.  The C# source won't compile on its own without its dependencies, but that's true of the ELM too - e..g, the Java Engine operates on the ELM files, and therefore needs all dependencies to be able to execute.  The `Library` resource has a `relatedArtifacts` property that we use to identify the dependencies, in much the same way in C you would `#include` the other symbols you need.

What we *can't* do is also include DLLs within the `Library` resources unless we have the entire dependency graph's C# files, in the same way that we can't link compiled C objects into a final library unless we have all of their dependencies present.

So, what you may want to refactor the PackagerCLI to do is:

First pass: create `Library` resources with CQL, ELM, and C# attachments.
2nd pass: given a dependency graph of `Library` resources, compile all C# files using Rosyln to create assemblies (1 per Library), and add the DLL as an attachment back to each `Library`.

This also has some utility as in some use cases we may not want to ship pre-built assemblies.  Users may want to compile their own assemblies using different options.  We may want to create a `LibraryCompiler` class that does this, which could then give users options on how to compile things.

ewoutkramer (approved):

PR #233, opened 2024-03-08
Overload resolution cohesion

EvanMachusak:
Adjusted overload resolution algorithms:

* Compatible check now only considers `Null` literals.  The intent of allowing `Any` typed values to be passed to functions with non-Any-typed arguments was to prevent users from needing to have usage like `Add(null as Integer, 5)`.
* If the target of coercion is an invalid interval, and the argument does not exactly match this interval type already, the cost is considered `Incompatible`.  The rationale used by the reference implementation is to avoid promoting to `Interval<Any>` because it is nonsensical.  This check is done after `ExactMatch` test, but before `IsSubtype`, because `IsSubtype` would return true for any `Interval<T>` and `Interval<Any>`, as explicitly described in the spec.   Valid intervals have point types which are both orderable via comparison operators and have `Predecessor` and `Successor` operators defined for them.
* `IsSubtype` now matches the spec's language exactly and uses the `ModelProvider` implementation
* The cheapest overload is now the lowest total cost instead of the one with the cheapest single argument coercion.  This is required for certain tests with the above changes also in place.

This PR lays the groundwork for creating a `MessageProvider` service to centralize error reporting.  The message language is stored in a resource file so that the door is open for localization.


EvanMachusak:
As it turns out, all tests will still pass when implementing `Compatible` as checking that `from` type is the `Any` type, so I made that change to align exactly with the spec language.

EvanMachusak:
I've documented how this all works [here](https://github.com/FirelyTeam/firely-cql-sdk/wiki/Function-overload-resolution)

Cql/Cql.CqlToElm/Messages.Designer.cs
```diff
@@ -0,0 +1,81 @@
+﻿//------------------------------------------------------------------------------
```
ewoutkramer, line 1: Are we starting multi-language support here?

ewoutkramer (approved):

Issue #195, opened 2024-02-23
References to codesystems defined elsewhere are not yet supported


ewoutkramer:
The CMS measures contain a file `QICoreCommon` that list several re-useable aliases for common codesystems:

```
codesystem "LOINC": 'http://loinc.org'
codesystem "SNOMEDCT": 'http://snomed.info/sct'
codesystem "ActCode": 'http://terminology.hl7.org/CodeSystem/v3-ActCode'
(etc)
```

These can then be referenced in other files like so:

```
code "Every eight hours (qualifier value)": '307469008' from QICoreCommon."SNOMEDCT" display 'Every eight hours (qualifier value)'
```

Our `ProcessCodeDef` function in the ExpressionBuilder does not yet support this, instead it expects all codesystems to be defined in the same file as they are referenced.

ewoutkramer:
See #196 - we'll have access to the external libraries so we can look this up.

PR #238, opened 2024-03-14
1.1/cross library codesystems

baseTwo:
Work for issue #195 (Add support for codesystems imported via includes)

## Work
* Introduce a `LibrarySet` which contains logic to load libraries in parallel and to store them in topological order. This logic was previously scattered between too many types and extensions. The following types were removed: `DirectedGraphExtensions`, `EnumerableExtensions`, `JsonConverterExtensions` (not used), `LibraryExtensions` (Elm)
* Introduce the top-level builder/context which operates on a `LibrarySet`, called `LibrarySetExpressionBuilder`/`LibrarySetExpressionBuilderContext`.
* `LibraryExpressionBuilderContext` builds the codesystems for the current library and from the includes.
* It is no longer the `AssemblyCompiler`'s purpose to build the definitions for the library set, instead that work must be done in the `LibrarySetExpressionBuilder` beforehand in the `LibraryPackager`

# Tests added
* `LibrarySetTests`
* `LibrarySetExpressionBuilderTests` (tests cross library codesystems)

## Other
* Move Validation into the ELM `Library`, such that it will be called after loaded from JSON, so that we don't have to clutter code everywhere else with it. e.g. there must be a valid identifier and includeDefs.
* Moved loading ELM from JSON from static `LibraryLoader` into ELM `Library` class
* Many properties are common between ELM elements, introduce interfaces to the partial classes to retrieve name, name and version, libraryName.
* Introduce a shared exception `CqlException<TError>` type, where `TError` is a struct "error" argument derived from `ICqlError` and allows error messages to be built. It's a lot easier to write many of these error types, than to create exception types.
* `ExpressionBuildingException` changed to `CqlException<ExpressionBuildingError>`
* Factories use inheritance on each other to avoid duplication.
* Update `ResourcePackager` to allow `LibraryPackager` as a dependency, and then have `PackagerCliProgram` depend on `ResourcePackager`
* Updated the dependencies graph (see Service Dependencies)

## Service Dependencies

```mermaid
classDiagram
    direction LR

    class TypeManager {
    }

    %% Inheritance  

    %% Injected Dependencies

    TypeManager ..> ExpressionBuilder : injected

    ResourcePackager ..> PackagerCliProgram : injected 


    IEnumerable_ResourceWriter_ ..> ResourcePackager : injected 
    LibraryPackager ..> ResourcePackager : injected


    TypeResolver ..> LibraryPackager : injected
    LibrarySetExpressionBuilder ..> LibraryPackager : injected
    
    TypeResolver ..> CSharpSourceCodeWriter : injected

    OperatorBinding ..> LibraryExpressionBuilder : injected

    LibraryExpressionBuilder ..> LibrarySetExpressionBuilder : injected

    %% Rest
```


Cql/Cql.Compiler/TypeManager.cs
```diff
@@ -56,22 +55,26 @@ internal class TypeManager
...
                 assemblyName = "Tuples";
+
             if (string.IsNullOrWhiteSpace(tupleTypeNamespace))
                 tupleTypeNamespace = "Tuples";
+
             AssemblyName = assemblyName;
             var assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(new AssemblyName(AssemblyName), AssemblyBuilderAccess.Run);
             TupleTypeList = new List<Type>();
-            Hasher = new Hasher();
+            Hasher = hasher;
```
ewoutkramer: Do we need to turn this into a dependency? Is there a reason you might want to replace this with another dependency?

Cql/Cql.Packaging/ResourcePackager.cs
```diff
@@ -1,16 +1,15 @@
...
 using Microsoft.Extensions.Logging;
-using Library = Hl7.Cql.Elm.Library;
 
 namespace Hl7.Cql.Packaging
 {
     /// <summary>
     /// Encapsulates packaging of resources and outputting them using the specified resource writers
     /// </summary>
-    public class ResourcePackager
+    [Obsolete]
```
ewoutkramer: Can you add to the Obsolete message what it is replaced with? So, what should the user of the class do when it gets this Obsolete message?

Cql/Elm/ElmJson.cs
```diff
@@ -1,139 +1,2 @@
...
-using System.IO;
-using System.Linq;
-using System.Reflection;
-using System.Text;
-using System.Text.Json;
-using System.Xml.Serialization;
-
-namespace Hl7.Cql.Elm
-{
-    internal class ElmJson
```
ewoutkramer, line 14: There were not in use? I once heard Evan complain he had to do a lot of work to make System.Text.Json do what he wanted, was assuming he meant this.....funny if it is not used in the end....

ewoutkramer (changes requested):
Just a few minor remarks/questions.

Cql/Cql.Compiler/TypeManager.cs
baseTwo: I change it back to a privately managed dependency as it was

Cql/Cql.Packaging/ResourcePackager.cs
baseTwo: I thought this would eventually be replaced by the ProgramCliPackager, but now I just injected ResourcePackager in the ProgramCliPackager, and LibraryPackager into ResourcePackager.

ewoutkramer (approved):

PR #247, opened 2024-03-22
Fix query lets

### LINKED ISSUE #199: A `let` cannot refer to an alias of a let that comes before it in the same scope
In CMS AntidepressantMedicationManagementFHIR.cql we find the following (correct) snippet of CQL:

```
define "Denominator Exclusions":
  Hospice."Has Hospice Services"
    or exists ( ( Status."Active Medication" ( ["MedicationRequest": "Antidepressant Medication"] ) ) ActiveAntidepressant
        let dosage: singleton from ActiveAntidepressant.dosageInstruction,
        timing: dosage.timing,
        boundsPeriod: timing.repeat.bounds,
        ActiveMedPeriod: boundsPeriod Meds
```

note that there is a `let` statement that first introduces `dosage` and then `timing` that uses `dosage`. This is useful and legal, but our current code batches up all lets in the same scope, and then adds them to the scope at the same time, which means that referring to a previous let in the same scope does not work.

I have fixed this in `SingleSourceQuery` by having each `let` introduce a new scope, with the subsequent lets defined in nested scopes. This is actually the formal way this is done in ML-style languages, so we might leave it at that.

ewoutkramer:
Fixes #199


Cql/Cql.Compiler/ExpressionBuilderContext.cs
```diff
@@ -263,6 +263,8 @@ internal ExpressionBuilderContext WithImpliedAlias(string aliasName, Expression
             return new ExpressionBuilderContext(this, element);
         }
 
+        internal ExpressionBuilderContext Pop() => _outerContext ?? throw new InvalidOperationException("Cannot pop the root context.");
```
baseTwo, line 266: Suggest that the `Deeper` method is renamed to `Push`

baseTwo (approved):

Issue #200, opened 2024-02-23
Extension.value is a DataType, and using it will raise errors.


ewoutkramer:
Caveat: I haven't fully researched this one, but it seems to me that choice types are again the culprit here:

```
  Patient.ethnicity E
    return Tuple {
      codes: { E.ombCategory } union E.detailed,
      display: E.text
    }
```

In this fragment from CMS SupplementalDataElements.cql, we see `E.ethnicity` and `E.ombCategory`, which are aliases for an extension defined in QICore.  The value of an Extension (in our POCO model) is DataType, and somewhere along the line, when these extension values are used, we are getting an invalid cast from `DataType to string` (for display).  

I have for now added a new cast that will cast a DataType to type string when a string is expected. This might actually work, since the CQL->ELM compiler has already checked that the assignment is valid, so we might as well assume it is.

ewoutkramer:
Added another cast from DataType to CqlInterval<DateTime> that will work when the datatype is a period.

ewoutkramer:
In general, the FHIR TypeConverter is *the* place to know that `DataType` actually means: a choice type from the underlying model. So it is its responsibility to try to convert (with `as`) a `DataType` to one of its subclasses (or return null otherwise).

PR #249, opened 2024-03-22
Improvements for dealing with CMS measures - part 2

### LINKED ISSUE #197: Query results are demoted to singles too aggressively
The current implementation of Query in the ExpressionBuilder demotes the result of a query to a single element if the source of the query is a non-list type. This implicitly implements automatic list-demotion which works well most of the time. However, this simplification is done even if the expected result of the query *is* a list, e.g. when compiling this line from CMS `AHAOverall`:

```
collapse (Event.dosageInstruction.timing.repeat.bounds DoseTime
                                        return DoseTime.toInterval())
```

the result of the nested query will be a single Interval (of DateTime), which is incorrect, since the `collapse` function requires a list as input. 

For now, if have stopped the Query translator from calling the `Single` operator if the required return type of the query indicates a list type.

ewoutkramer:
* Improves dealing with queries on non-list sources that result in list items
* Avoids use of choice-typed objects on crucial places by improving the underlying CQL
* Improved handling of choice-typed objects represented as DataType objects in the POCOs.

For the last bullet, I had to enrich the TypeResolver to be able to communicate which type is used by the model to represent choice types.

Fixes #197 and #200.


ewoutkramer:
Making this a draft, as I still have to fix errors in the generated C# that belong to these issues.

baseTwo (approved):

Issue #196, opened 2024-02-23
FHIRHelpers hack - how far do we go?


ewoutkramer:
We have special cases built-in for handling calls to the FHIRHelpers library.  Each of the functions we've encountered so far has manual handling in `GetFunctionRefReturnType`. Could we, with our current understanding, actually compile this file and refer to it, so we don't need to manually figure out the return types?

For now I have added:
* ToInterval - existed, but changed it so it will return Interval<DateTime> when we don't know the type statically (will need to revisit why the type is unknown there, probably because it is a property that is a choice element)
* ToValue - this returns a complex `Choice<all FHIR datatypes supported by function, Choice<all other types>>`.  For now I made it return `typeof(object)`, but what to do here?

EvanMachusak:
This is supposed to be gotten from the modelinfo.  These are implicit conversions implemented through FHIRHelpers.

ewoutkramer:
Ok, during our call we agreed we would add support for reaching out to dependent libraries from the ExpressionBuilder, so we can actually look this up in the compiled FHIRHelpers. This is also necessary to fix #195 .  And we can avoid having modelinfo this way.

PR #256, opened 2024-03-26
2.0/Fix/Cross library definitions (remove temporary fhirhelpers code)

baseTwo:
ℹ️Work done for issue #196 

## Cross Library Function Calling
* Possible to call functions from included libraries. 
* `LibraryExpressionBuilder` includes definitions from includes before processing its own definitions. 
* In `ExpressionBuilder.FunctionRef`, removed hacky code where calls to FHIRHelper functions were generated via CQL Operator binding, and instead, call FHIRHelper functions direcly. (see changes in outputed C# files)
* In `ExpressionBuilder.GetFunctionRefReturnType`, remove hacky code to determine function result types

## Tests
* Fixed corrupt FHIRHelpers-4.0.1.json in tests folder
* `Medication_Request_Example_Test` must now also include FHIRHelpers as a dependency to run

## Other
* Renamed `LibrarySetExpressionBuilder._allDefinitions` -> `LibrarySetExpressionBuilder.LibrarySetDefinitions`  and make public
* Renamed `LibraryExpressionBuilder.Definitions` -> `LibraryExpressionBuilder.LibraryDefinitions` 
* Renamed `AlreadyCreatedStreamPreviouslyForLibraryNameError` -> `AlreadyCreatedStreamPreviouslyForLibraryNameError`
* All ELM elements that has a name and version, will display this in ToString(). DebuggerVisualizer as well.
* Add nullability hints on the `DefinitionDictionary<T>.Try*` methods.
* Add another `DefinitionDictionary<T>.Merge` method to merge expressions for a single library.
* Logs to the console are less cluttered (no need to see the class names where the messages are coming from)


Cql/Cql.Compiler/ExpressionBuilder.cs
```diff
@@ -1034,33 +1036,11 @@ protected Expression FunctionRef(FunctionRef op)
                 .Concat(new[] { functionType })
                 .ToArray();
 
-            var funcType = GetFuncType(funcTypeParameters);
-
-            // FHIRHelpers has special handling in CQL-to-ELM and does not translate correctly - specifically,
-            // it interprets ToString(value string) oddly.  Normally when string is used in CQL it is resolved to the elm type.
-            // In FHIRHelpers, this string gets treated as a FHIR string, which is normally mapped to a StringElement abstraction.
-            if (op.libraryName != null && op.libraryName.StartsWith("fhirhelpers", StringComparison.OrdinalIgnoreCase))
```
ewoutkramer, line 1054: It could well be that we still need this to be here, as it seems to suggest the ToString() function in FHIRHelpers is incorrect (for our purposes)

Cql/CoreTests/Input/ELM/Libs/FHIRHelpers-4.0.1.json
ewoutkramer, line 1: What happened to this file? This is output from the compilation from ELM, right? did we run a different cql2elm converter?

Cql/Elm/PublicAPI.Unshipped.txt
```diff
@@ -813,8 +813,6 @@ static readonly Hl7.Cql.Elm.Library.JsonSerializerOptionsStrict -> System.Text.J
 ~Hl7.Cql.Elm.CodeFilterElement.value.set -> void
 ~Hl7.Cql.Elm.CodeFilterElement.valueSetProperty.get -> string
 ~Hl7.Cql.Elm.CodeFilterElement.valueSetProperty.set -> void
-~Hl7.Cql.Elm.CodeRef.libraryName.get -> string
```
ewoutkramer: Was that an element we added in a partial class and that we have removed again?

Demo/Measures-cms/AdvancedIllnessandFrailty-1.8.000.cs
```diff
@@ -94,6 +94,51 @@ public AdvancedIllnessandFrailty_1_8_000(CqlContext context)
 
     #endregion
 
+	private CqlCode[] LOINC_Value()
+	{
+		var a_ = new CqlCode[0]
```
ewoutkramer: This looks wrong, in the old version there were codes returned here, but now it is empty.

Demo/Measures-cms/CumulativeMedicationDuration-4.0.000.cs
```diff
@@ -140,6 +140,30 @@ public CumulativeMedicationDuration_4_0_000(CqlContext context)
 
     #endregion
 
+	private CqlCode[] V3TimingEvent_Value()
+	{
+		var a_ = new CqlCode[0]
```
ewoutkramer: This looks wrong, in the old version there were codes returned here, but now it is empty. (etcetera)

Demo/Measures/AdultOutpatientEncountersFHIR4-2.2.000.cs
```diff
@@ -133,8 +133,8 @@ private IEnumerable<Encounter> Qualifying_Encounters_Value()
 		var n_ = context.Operators.ListUnion<Encounter>(k_, m_);
 		bool? o_(Encounter ValidEncounter)
 		{
-			var q_ = context.Operators.Convert<string>(ValidEncounter?.StatusElement);
-			var r_ = context.Operators.Equal(q_, "finished");
+			var q_ = FHIRHelpers_4_0_001.ToString(ValidEncounter?.StatusElement);
```
ewoutkramer: I think this should be the original Convert<string>, but that's because of a piece of code that we removed and that I commented on above.

Demo/Measures/AdultOutpatientEncountersFHIR4-2.2.000.cs
```diff
@@ -133,8 +133,8 @@ private IEnumerable<Encounter> Qualifying_Encounters_Value()
...
+			var r_ = context.Operators.EnumEqualsString(q_, "finished");
```
ewoutkramer: Do we understand why this happens now?

ewoutkramer (changes requested):
There are some suspicous changes to the generated code - maybe you know what's going on?

Cql/CoreTests/Input/ELM/Libs/FHIRHelpers-4.0.1.json
baseTwo: I don't know how it was generated originally, but it could not be parsed to ELM. I just copied FHIRHelper.json from the Demo library set folder to the test folder

Cql/Cql.Compiler/ExpressionBuilder.cs
ewoutkramer: Although I cannot follow Evan's wording here, he's right that there's something odd about the ToString() function - it uses the `.value` property, which does not really exist in FhirPath, to refer to the atomic value of a primitive (not the primitive itself)  (BTW it even does that for non-primitives). This `value` property requires special handling in the model, so the actual `ToString` function cannot currently be used or called. So we need to reroute the call to the convert to string. Since this ultimately calls the model-specific TypeConverter, it gives the model the opportunity to do model-specific mapping of the `value` property, which is exactly what we want.

Cql/Elm/PublicAPI.Unshipped.txt
baseTwo: This was an accident. There is actually quite a lot of public members missing from this file from the generated ELM.g.cs. I'll put them all in

baseTwo: This was an accident removal. I see a lot of methods are missing from the generated elm file, they have all been added.

Cql/Elm/PublicAPI.Unshipped.txt
```diff
@@ -531,12 +572,18 @@ Hl7.Cql.Elm.Quantity.value.get -> decimal
 Hl7.Cql.Elm.Quantity.value.set -> void
 Hl7.Cql.Elm.Quantity.valueSpecified.get -> bool
 Hl7.Cql.Elm.Quantity.valueSpecified.set -> void
+Hl7.Cql.Elm.Quantity1
```
ewoutkramer: There are differences here that I cannot explain, like Quantity1 (and even Long and Decimal being added? Why weren't they there before?)

ewoutkramer (dismissed):

ewoutkramer (approved):

Cql/Elm/PublicAPI.Unshipped.txt
baseTwo: Generated again without types.xsd

Issue #255, opened 2024-03-26
2.0/Cleanup/Improve Factory Types


baseTwo:
Create a factory for AssemblyCompiler. Also move the Deferred calls in the ctors to virtual methods, so that any consumer developer has the ability to swap out implementation types.

1. Move classes into files matching the same name. 
2. Scopes should also be maintained in a stack in ExpressionBuilder, similar to Pushing Elements
3. 
4. Do not pass tuple types from the LibraryDefinitionBuilder to the AssemblyCompiler via the TypeManager. Should be an output from the former then be passed via the pipeline
5. Generated c# files need to end file extensions with .g.cs [See why](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references#:~:text=to%20address%20yet.-,Important,Generators%20can%20opt%2Din%20using%20the%20%23nullable%20preprocessor%20directive.,-By%20default%2C%20nullable). Preferably they should also start with the <autogenerated> comment block at the top of the file

PR #257, opened 2024-03-27
2.0/Cleanup/Improve Factory Types

baseTwo:
ℹ️Work for issue #255 : "Create a factory for AssemblyCompiler. (Also move the Deferred calls in the ctors to virtual methods)"

* Cleaned up Factories to be more flexible for consumer developers
* Create one factory class per project that is capable of creating all the types necessary in that project
* Favour using factories over service host dependency injection in PackagerCliProgram

## Other
* Make block-scoped namespaces the default in editorconfig
* Fix: The `--override-utc-date-time` argument wasn't used until now to generate fhir resource dates
* Add the FHIR Resources output under LibrarySets


baseTwo:
I notice the fhir resource output isn't in the repo anymore. Need to put this back

Cql/Elm/PublicAPI.Unshipped.txt
```diff
@@ -2,1286 +2,61 @@
 const Hl7.Cql.Elm.Library.JsonMimeType = "application/elm+json" -> string!
 const Hl7.Cql.Elm.Library.LibraryNodeProperty = "Library" -> string!
 const Hl7.Cql.Elm.Library.XmlMimeType = "application/elm+xml" -> string!
-Hl7.Cql.Elm.Abs
```
ewoutkramer: Why do these disappear?

ewoutkramer (changes requested):

ewoutkramer:
Wait, it's no longer up to date. Let's check that first.

ewoutkramer (approved):

Issue #258, opened 2024-03-28
2.0/Cleanup/Move classes into files matching the same name


baseTwo:
As a follow up to issue #253 in which a lot of refactoring was done, classes renamed, etc. The file names were kept so that it was easier to review pull requests, but now types are mixed in a file and not matching class name to file name.

Wait until there are no more open pull request before moving the types to the correct files.

PR #265, opened 2024-03-28
2.0/move types into matching files

baseTwo:
ℹ️Work for issue #258 

# Work
* Move types into files matching on the name
* Rename `NoCategoryConsoleFormatterDecorator` -> `CleanConsoleFormatter`

# Other
* Removed some `*.cmd` files that are no longer needed
* Include the generated Resources in the repo
* Make dll output deterministic (otherwise a fresh guid is put into the dll every time, making resources differ a slight bit every time)  Note: There may still be differences in the exact output when compiling between a Windows or Mac machine. Might decide to not include Resources in the repo at  a later stage, but since only @ewoutkramer and @baseTwo are working on the CMS measures right now, it is useful to keep it there.


ewoutkramer (dismissed):

Demo/Elm/Build/Elm.Targets.xml
```diff
@@ -7,7 +7,7 @@
 	<Target Name="ELM to C#"
 			BeforeTargets="Compile">
 		<Message Text="Converting ELM to C#"/>
-		<Exec Command='"$(SolutionDir)/Cql/PackagerCLI/bin/debug/net6.0/Hl7.Cql.Packager.exe" --canonical-root-url https://fire.ly/fhir/ --elm "$(SolutionDir)\LibrarySets\Demo\Elm" --cql "$(SolutionDir)\LibrarySets\Demo\Cql" --cs "$(MSBuildProjectDirectory)/../Measures" --fhir "$(MSBuildProjectDirectory)/../Test/Resources"' />
+		<Exec Command='"$(SolutionDir)/Cql/PackagerCLI/bin/debug/net6.0/Hl7.Cql.Packager.exe" --override-utc-date-time "1970-01-01T00:00:00Z" --canonical-root-url https://fire.ly/fhir/ --elm "$(SolutionDir)\LibrarySets\Demo\Elm" --cql "$(SolutionDir)\LibrarySets\Demo\Cql" --cs "$(MSBuildProjectDirectory)/../Measures" --fhir "$(SolutionDir)\LibrarySets\Demo\Resources"' />
```
ewoutkramer: Pretty sure these should all be forward slashes to make this work under Mac?

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #267, opened 2024-04-02
2.0/overload conversions compiler errors

### LINKED ISSUE #194: Incorrect translation of binding name in QICore model?
The CMS file `Status.cql` contains the following function:

```
 //Observation Vital Signs
define function "BloodPressure"(Obs List<"QICore.observation-bp">):
  Obs O
    where O.status in { 'final', 'amended', 'corrected' }
```

Note that here, `QICore.observation-bp` is used, not observation, and looking at its mapping in https://raw.githubusercontent.com/cqframework/clinical_quality_language/master/Src/java/quick/src/main/resources/org/hl7/fhir/qicore-modelinfo-4.1.1.xml, the status element maps to `QICore.Status`

```
      <element name="status" elementType="QICore.Status" target="%value.value" description="registered | preliminary | final | amended +" definition="The status of the result value." comment="This element is labeled as a modifier because the status contains codes that mark the resource as not currently valid." mustSupport="true">
         <binding name="Status" strength="Required"/>
      </element>
```

In contrast, in the base Observation in the QICore model, the type of this `status` element is actually `QICore.ObservationStatus`.

When the above CQL is translated to ELM, we find:

```
 "operand" : [ {
                     "locator" : "195:11-195:18",
                     "resultTypeName" : "{http://hl7.org/fhir}Status",
                     "path" : "value",
                     "type" : "Property",
                     "source" : {
                        "path" : "status",
                        "scope" : "O",
                        "type" : "Property"
                     }
```

The type has been translated to the FHIR binding `Status`, which is *not* the binding for `Observation.status` but the binding for `VerificationResult.status` -- causing a casting error in our ExpressionBuilder.

### LINKED ISSUE #200: Extension.value is a DataType, and using it will raise errors.

### LINKED ISSUE #201: Java CQL->ELM compiler seems to use FHIR types or bindings that do not exist.
In the file AntithromboticTherapyByEndofHospitalDay2FHIR.cql we find:

```
define "Reason for Not Administering Antithrombotic":
  ["MedicationAdministrationNotDone": medication in "Antithrombotic Therapy for Ischemic Stroke"] MedicationAdm
    where ( MedicationAdm.statusReason in "Medical Reason for Not Providing Treatment"
        or MedicationAdm.statusReason in "Patient Refusal"
    )
    return {
      id: MedicationAdm.id,
      authoredOn: MedicationAdm.recorded
    }
```

According to the generated ELM, the type of  `authoredOn` in the returned tuple is `{http://hl7.org/fhir}NotDoneRecorded`, which is not a known type.

I will remove this statement to continue.

baseTwo:
* Changed the way we resolve function calls: instead of selecting directly by type of the parameters, we now allow overloads to be selected that require casts too - and Convert() will now insert those casts.
* Changed the TypeConverter so we can have conversions that go beyond a fixed from/to type - necessary to support conversions that are more dynamic (in this case, from DataType to any subclass of DataType).
* Added a correction function that translates incorrect QICore profiled type names to the core type name.
* Paul moved stuff around ;-)
* Upgraded cql2elm to version 3.1.0

Fixes #194.
Fixes #200.
Fixes #201.


Cql/CoreTests/ModelTest.cs
```diff
@@ -90,7 +90,7 @@ private class UnitTestTypeResolver : BaseTypeResolver
             internal override PropertyInfo GetPrimaryCodePath(string typeSpecifier) => throw new NotImplementedException();
 
             internal override bool ShouldUseSourceObject(Type type, string propertyName) => true;
-            internal override bool RepresentsChoiceType(Type type) => false;
+          //  internal override bool RepresentsChoiceType(Type type) => false;
```
baseTwo: Remove

Cql/Cql.Abstractions/TypeResolver.cs
```diff
@@ -190,14 +190,14 @@ public abstract class TypeResolver
...
 
-        /// <summary>
-        /// Returns true when the given type is used by the model to represent a FHIR choice type.
-        /// Since C# does not have a direct representation of choice types, this method is used by
-        /// the infrastructure to insert dynamic handling of choice types at runtime.
-        /// </summary>
-        /// <param name="type"></param>
-        /// <returns></returns>
-        internal abstract bool RepresentsChoiceType(Type type);
+        // /// <summary>
```
baseTwo: Remove

Cql/CoreTests/ModelTest.cs
ewoutkramer: ok!

Cql/Cql.Firely/FhirTypeResolver.cs
```diff
@@ -106,10 +106,10 @@ internal override bool ShouldUseSourceObject(Type type, string propertyName)
             return type == typeof(FhirDateTime) && propertyName == "value";
         }
 
-        internal override bool RepresentsChoiceType(Type type)
-        {
-            return type == typeof(DataType) || type == typeof(Element);  
-        }
+        // internal override bool RepresentsChoiceType(Type type)
```
baseTwo: Remove

Cql/Cql.Runtime/BaseTypeResolver.cs
```diff
@@ -123,11 +123,20 @@ public abstract class BaseTypeResolver : TypeResolver
         /// <inheritdoc/>
         public override Type? ResolveType(string typeSpecifier)
         {
-            if (Types.TryGetValue(typeSpecifier, out var type))
+            var correctedTypeSpecifier = CorrectQiCoreExtensionTypes(typeSpecifier);
+            
+            if (Types.TryGetValue(correctedTypeSpecifier, out var type))
                 return type;
-            else throw new ArgumentException($"Type {typeSpecifier} is not bound", nameof(typeSpecifier));
+            else throw new ArgumentException($"Type {correctedTypeSpecifier} is not bound", nameof(typeSpecifier));
```
baseTwo, line 130: Blank lines after a return and the else is not necessary

Demo/Cql/Cql.csproj
```diff
@@ -5,4 +5,9 @@
 
 	<PropertyGroup>
 	</PropertyGroup>
+
+	<ItemGroup>
```
baseTwo, line 6: ItemGroup (and PropertyGroup above) can be removed

baseTwo (commented):
I cant click Request changes?

baseTwo:
PR description needs to be supplied and linked with associated issue number

Cql/Cql.Compiler/Expressions/FunctionCallExpression.cs
```diff
@@ -54,7 +54,7 @@ internal class FunctionCallExpression : Expression
         public override Expression Reduce()
         {
             var argumentTypesExpressions = Arguments
-                .Skip(1)
+                //.Skip(1)
```
ewoutkramer: Pretty sure this is wrong.

Cql/Cql.Firely/FhirTypeResolver.cs
ewoutkramer: Remove

ewoutkramer (changes requested):

Cql/Cql.Firely/FhirTypeResolver.cs
ewoutkramer: done

Demo/Cql/Cql.csproj
ewoutkramer: done.

ewoutkramer:
> PR description needs to be supplied and linked with associated issue number

That would be asking changes of yourself - although I agree that it procedurally can make sense to do so.

ewoutkramer (approved):

PR #268, opened 2024-04-02
Upgraded to latest + moved SDK version to a csproj property.

ewoutkramer:
Upgraded the CQL library to use the latest 5.7.0 Firely .NET SDK.


Demo/cql-demo.props
```diff
@@ -4,6 +4,10 @@
 		<TargetFrameworks>net8.0</TargetFrameworks>
 	</PropertyGroup>
 
+	<PropertyGroup>
```
baseTwo, line 7: Maybe this property group can be joined with the one above?

baseTwo (approved):
Approved with one small comment

Demo/cql-demo.props
ewoutkramer: I kind of liked the grouping: compiler options (targetframework) vs. constants used as macros in our references....

Issue #220, opened 2024-02-28
Multi-source queries with singletons as source are not/not completely supported


ewoutkramer:
The `MultiSourceQuery()` function - much like the normal Query function in principle allows singletons to work as source, by wrapping them in an array. However, this wrapper code seems to come into play too late:

```csharp
        var multiSourceTupleType = TypeManager.TupleTypeFor(tupleSpecifier, ctx, (type) =>
                IsOrImplementsIEnumerableOfT(type)
                    ? TypeManager.Resolver.GetListElementType(type, true)!
                    : throw new NotSupportedException("Query sources must be lists."));
   
         var crossJoinedSource = CrossJoin(query.source!, multiSourceTupleType, ctx);
            var source = crossJoinedSource;

            var isSingle = false;
            if (!IsOrImplementsIEnumerableOfT(source.Type))
            {
                var arrayInit = Expression.NewArrayInit(source.Type, source);
                source = arrayInit;
                isSingle = true;
            }
```

We see code that calls `NewArrayInit` when the type does not implement IEnumerable, but already at the first lines (when `TupleTypeFor()` is called) we throw an exception because "Query sources must be lists". This means that code like

```

define "Urinary Symptom Score 6 to 12 Months After Initial BPH Diagnosis":
  Last("Urinary Symptom Score Assessment" USSAssessment
      with "Initial BPH Diagnosis Starts Within 6 Months Before or After Start of Measurement Period" InitialBPHDiagnosis
        such that months between start of InitialBPHDiagnosis.prevalenceInterval() and USSAssessment.effectiveDatetime in Interval[6, 12]
      sort by effectiveDatetime
  )

define "Urinary Symptom Score Change":
  from
    "Urinary Symptom Score Within 1 Month after Initial BPH Diagnosis" FirstUSSAssessment,
    "Urinary Symptom Score 6 to 12 Months After Initial BPH Diagnosis" FollowUpUSSAssessment
    let USSChange: FirstUSSAssessment.valueInteger - FollowUpUSSAssessment.valueInteger
    return USSChange
```

from UrinarySymptomScoreChangeAfterBenignProstaticHyperplasiaFHIR.cql does not compile:  `Urinary Symptom Score 6 to 12 Months....` does a `Last`, so returns a single value - which then breaks the multi source query in `Urinary Symptom Score Change`.

EvanMachusak:
[...]

PR #279, opened 2024-04-05
2.0/multi-source queries

baseTwo:
ℹ️Work for issue #197 and  #220 

# Fixing multi-source queries with cross joins
* Merging `Single`- and `MultiSourceQuery` methods into one `QueryImpl`
* Use a new CQL operator `CrossJoin` instead of repeated nesting of `SelectMany` -- more efficient since it doesn't have to create a new two-item tuple type between each consecutive foreach
* A new `TypeFor` method that can construct cql tuples from a dictionary of (string name, Type type) 
* Allow for up to 8 source on a cross-join on the cql operator context. CMS measures have only up to 7 sources. Will need to expand on this in the future with nested CrossJoins.

# Other Changes
## Dumping debug information on queries to the console. 
Dump the source and cql info to the log for queries. For this it was necessary to know how to get from the library to the cql file, so the library type was modified so that the OriginalFileName is set to the elm file from which it was loaded, and then using a convention to get from the elm file to the cql file
image

## `CqlOperator` Enum:
* List-based operations to start with `List-` for consistency: `ListElementAt`, `ListIndecOf`, `ListSort`
* Added `CrossJoin`

## Extensions
* `SelectToArray` pre-allocates an array to a known size of the source enumeration, then selects the source items into the target array. Less wasteful than calling `enumerable.ToArray`

## `TypeResolver`
* Updated code comments
* Renamed parameter `@throw` to `throwError` (consistency)

## CQL Errors
* Renamed parameters for type names to be consistent: `TypeName`
* Made some parameters nullable

## Dumpify Nuget
* A debug package making it easier to dump objects while debugging to the console. (Similar output than linqpad)

## EditorConfig
Added two config settings for ReSharper, to set max args/params per line to 2, after which they will be put on multilines


ewoutkramer:
Also fixed #197.

Cql/Cql.Compiler/ExpressionBuilder.Query.cs
```diff
@@ -1,57 +1,66 @@
...
         }
 
-        var @return = source;
+        ParameterExpression scopeParameter;
 
-        Type elementType = _typeManager.Resolver.GetListElementType(@return.Type, @throw: true)!;
+        var (@return, isSourcesPromoted) = ctx.ProcessQuerySources(query);
+        var isAnySourcePromoted = isSourcesPromoted.Any(isSourcePromoted => isSourcePromoted);
+        var returnType = ctx._typeManager.Resolver.GetListElementType(@return.Type, true)!;
+        var returnElementType = ctx._typeManager.Resolver.GetListElementType(@return.Type, true)!;
```
baseTwo, line 33: Exactly the same type, remote it and the check below

Cql/Cql.Compiler/ExpressionBuilder.Query.cs
```diff
@@ -86,339 +95,370 @@ protected Expression SingleSourceQuery(Query query)
...
+        //    IEnumerable<A> a = ...;
+        //    IEnumerable<B> b = ...;
+        //    IEnumerable<c> c = ...;
+        //    IEnumerable<(A, B, C)> crossJoinedValueTupleResults = CrossJoin<A, B, C>(a, b, c);
+
+        var tryPromoteSourceExpressions = sourceExpressions.SelectToArray(expr => TryPromoteSource(expr));
+        var promotedSourceExpressions = tryPromoteSourceExpressions.SelectToArray(s => s.source);
+        var isSourcesPromoted = tryPromoteSourceExpressions.SelectToArray(s => s.isSourcePromoted);
+
+        if (promotedSourceExpressions.Any(promotedSource => !IsOrImplementsIEnumerableOfT(promotedSource.Type)))
```
ewoutkramer: Based on the lines above, how can any of the sources still not be promoted?

ewoutkramer (changes requested):

Cql/Cql.Compiler/ExpressionBuilder.Query.cs
baseTwo: I changed the wording here to indicate if the original sources were singletons.

ewoutkramer (approved):

baseTwo:
Replaced with another PR #285

baseTwo:
Replaced with another PR #285

PR #280, opened 2024-04-08
Simplify conditions in CaseWhenThen blocks

ewoutkramer:
1. Our normal Expression rewriting logic would make sure that "if then" *statements* are turned into lambda's so they can be expressed as *expressions*. This logic, however, was not called on the conditions of CaseWhenThen blocks, so CaseWhenThen blocks with CaseWhenThen blocks in their conditions would fail to render correct C#. This PR makes sure case conditions are simplified. This is done by turning each condition into a lambda that can be called by the conditional blocks. The advantage of turning them into lambdas (instead of normal assignments) is that we delay execution, so the (expensive) conditions of else statements are not evaluated before the cases above them have been.

2. Not all invalid characters in definition names were escaped correctly, this has now been fixed
3. Added our normal (SDK) editorconfig settings
4. Several rewrites as suggested by Rider.


Cql/CodeGeneration.NET/ExpressionConverter.cs
```diff
@@ -77,7 +72,7 @@ public string ConvertExpression(int indent, Expression expression, bool leadingI
             }
         }
 
-        public string LibraryName { get; }
+        public string LibraryName { get; } = libraryName;
```
baseTwo, line 75: All assignments to be placed inside the constructor

ewoutkramer: This class has no constructor, as it does not need one...

Cql/CodeGeneration.NET/ExpressionConverter.cs
```diff
@@ -20,13 +20,8 @@
 
 namespace Hl7.Cql.CodeGeneration.NET
 {
-    internal class ExpressionConverter
+    internal class ExpressionConverter(string libraryName)
```
baseTwo, line 23: I'm not convinced we should use primary constructors for non-poco types, because we have different naming convention between parameters and fields, and this forces them to be the same

Cql/CodeGeneration.NET/ExpressionConverter.cs
```diff
@@ -494,7 +489,7 @@ public string ConvertTopLevelFunctionDefinition(int indent, LambdaExpression fun
         // Linq.Expressions needs an explicit conversion from a value type
         // type to object, but the C# compiler will insert that boxing,
         // so we can remove those casts.
-        private static Expression StripBoxing(Expression node)
+        private static Expression stripBoxing(Expression node)
```
baseTwo: Method names are PascalCase. (multiple cases in this class)

Cql/CodeGeneration.NET/ExpressionConverter.cs
ewoutkramer: Here, `libraryName` is just an argument to a constructor, which has always started with a lowercase character. If you want to have properties with the same name, but capitalized, you can (just as you had with conventional constructors), and then assign to the property. Or maybe I don't understand it.  This was more true for *record constructors*, in which the parameters were turned directly into properties, so you had to name the parameters with a capital...

Cql/CodeGeneration.NET/ExpressionConverter.cs
ewoutkramer: weird, but there will be tons of those, as much of this code predates the marriage with our SDK. But I'll certainly rename this one!

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
```diff
@@ -18,16 +18,17 @@ namespace Hl7.Cql.CodeGeneration.NET.Visitors
     /// <summary>
     /// This Visitor will (in most cases) create a new variable for a nested expression, 
     /// and assign the visited node's expression to that variable, thus unwinding the deeply nested
-    /// structure of Linq.Expression.
-    /// e.g. exprA(exprB(4)) will be turned into
+    /// structure of Linq.Expression to make it easier to debug and inspect the intermediate values.
+    ///
+    /// E.g. exprA(exprB(4)) will be turned into
```
baseTwo: Empty /// does not introduce a newline in the documentation. Instead use `<para>` for each paragraph

Cql/CodeGeneration.NET/ExpressionConverter.cs
ewoutkramer: Weird, it's lower case in my code. What's going on here...

ewoutkramer: oh, I see.  No, in the SDK, private methods are lower-cased.

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
```diff
@@ -18,16 +18,17 @@ namespace Hl7.Cql.CodeGeneration.NET.Visitors
...
     ///     var b = exprB(4)
     ///     var a = exprA(b)
     ///     return a;
     /// </summary>
     internal class SimplifyExpressionsVisitor : ExpressionVisitor
     {
-        private bool _atRoot = true;
-        private readonly List<BinaryExpression> _assignments = new();
+        private readonly List<BinaryExpression> _assignments = [];
```
baseTwo, line 31: Keep assignments inside the constructor

ewoutkramer: I totally disagree.

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
ewoutkramer: Github swallowed the tag, but is it the `para` element we need?

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
```diff
@@ -152,26 +162,30 @@ protected override Expression VisitBinary(BinaryExpression node)
 
                 // The interim value of an assignment is clear, we don't need to simplify
                 { NodeType: ExpressionType.Assign } => base.VisitBinary(node),
-
-                _ => simplify(base.VisitBinary(node))
+
```
baseTwo: Remove whitespace-only line

ewoutkramer: All the cases have a whitespace-only line here, so make it clear to which line the comment belongs....

Cql/CodeGeneration.NET/ExpressionConverter.cs
baseTwo: Here 'libraryName' is both the parameter name in the constructor and also the field in the class. Fields start with an underscore, so this breaks convention. Furthermore, primary class constructors cannot generate fields as readonly.

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
ewoutkramer: I added `para` and `code` tag now!

Cql/CodeGeneration.NET/ExpressionConverter.cs
baseTwo: That's a Java-centric convention :)

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
baseTwo: If you want the example to be in it's own paragraph, then yes.

Cql/CodeGeneration.NET/ExpressionConverter.cs
ewoutkramer: I see them as parameters to the class that can be accessed everywhere - so they are kind of "new" - they are neither fields nor properties.  Agree that it would be better if they were readonly, but I don't think you can say that of any parameter to any function in C#.

Cql/CodeGeneration.NET/ExpressionConverter.cs
ewoutkramer: Hahaha, maybe you can still see my roots here.  I am not against changing that rule (it might well be FS is using pascal case here) - but realize that this is has been a standing convention in the SDK codebase for 10+ years, so it's going to be inconsistent for a while....  In any case, it might be nice to discuss with Kas and Marten and see what they think.

It's not just cosmetic for me - I like to make the distinction between public methods clearer, just like you can see the difference between public properties and private fields easily - I feel I need the same for methods.

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
baseTwo: It keeps class member initialization related logic localized, especially since there are a number of large types in this repo where fields and property initializations are not located within close proximity to each other, making it harder to get a understanding on the initial state of the type.

Cql/CodeGeneration.NET/ExpressionConverter.cs
baseTwo: As long as we have a convention documented somewhere. This repo is currently inconsistent with following any guideline. I'm in favour of solidifying these into the editorconfig

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
baseTwo: It was on line 165. In the code diff it's clearer on which line the comment was placed and it also highlights the whitespace. GitHub should have displayed the whitespace in the summary view on the PR as well

image

Cql/CodeGeneration.NET/ExpressionConverter.cs
ewoutkramer: That's right - because it started out with Evan.  Ok, why don't we take some time this week to find out what we need to do. For now, I'll remove that .editorconfig and stick to Evan's originals.

Cql/CodeGeneration.NET/ExpressionConverter.cs
baseTwo: Solution to making `libraryName` readonly, is to add a field explicitly.

image

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
ewoutkramer: Ok, but that's not true for these simple expression subclasses that have little state and no constructor logic at all. The whole point of the new default constructors is removing boilerplate constructors for these cases. So, if you keep your classes simple (as you should), one should not be "punished" by still having constructs that are necessary for classes that are too big.

Cql/CodeGeneration.NET/ExpressionConverter.cs
ewoutkramer: You wouldn't do that with functions either.... and Rider will do its utmost to refactor that away again. I think it's not worth the cost of adding boilerplate.  We got used to it for functions too after all.

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
ewoutkramer: Weird....that must be a diff artifact!  I would not want an empty line AFTER that default return either...

image

Cql/CodeGeneration.NET/ExpressionConverter.cs
baseTwo: Ok, I'm hoping C# vNext solves this somehow, allowing modifier keywords to be used in primary constructor parameters such as `readonly`. I'll resolve this for now.

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
baseTwo: Ok, I'm happy to resolve as in this case they are all together.

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
baseTwo: The whitespace that I'm referring to is on the line just above `_ => MakeLet...`. That line contains only spaces. (It should be a blank line). Could you add the following line to editorconfig as well? `trim_trailing_whitespace = true`
image

ewoutkramer:
Ok, updated the comment with para, removed SDK-styles from the .editorconfig, renamed private functions to use PascalCase.

Cql/CodeGeneration.NET/Visitors/SimplifyExpressionsVisitor.cs
ewoutkramer: Ah, I cannot see that.  Yes, I'll add that to the .editorconfig

ewoutkramer: I've added it, but this whitespace is invisible to me, and very easy to introduce: if I press enter after a line, the editor will insert whitespace so my cursor aligns with the previous line (which makes sense), but that whitespace will stay there.

Cql/CodeGeneration.NET/ExpressionConverter.cs
ewoutkramer: It's being discussed: https://github.com/dotnet/csharplang/discussions/7699

baseTwo (approved):

PR #289, opened 2024-04-10
develop-2.0-04-10-cql-operator-binder-refactor

baseTwo:
ℹ️Work for issue #292 

# Main Work
## CqlOperatorsBinding
* New `CqlContextExpressions` added as a central cache for expressions and reflection info needed in `CqlOperatorsBinder`.
* Move `LibraryDefinitionsBuilder.ParameterExpression` to `CqlContextExpressions.ParameterExpression`, since it was only ever used in `CqlOperatorsBinder`. The `Bind` method no longer accepts this as a parameter, so the `ExpressionBuilder` is cleaned up from sending over this static instance in many places.
* Add partial file `CqlOperatorsBinding.Bind.cs` which contains the main generic binding logic to `ICqlOperators`
* Add partial file `CqlOperatorsBinding.Specific.cs` and moved the specific binding methods there.

## CqlOperators
* The methods on `ICqlOperators` have been renamed to match the enum values on `CqlOperator'

## ExpressionBuilder
* Removed `BindCqlOperator` method and call `_operatorBinding.BindToMethod` directly. It is no longer necessary to pass the parameter expression, so this helper method is no longer needed.
* Removed these methods after they were simplified into single-lines: `MaxValue`, `MinValue`, `DateFrom`, `DifferenceBetween`, `DurationBetween`, `Expand`, `First`, `Last`, `Descendents`, `ConvertQuality`, `ChangeTypeToBoolean`, `ChangeTypeToConcept`, `ChangeTypeToDate`, `ChangeTypeToDateTime`, `ChangeTypeToDecimal`, `ChangeTypeToLong`, `ChangeTypeToInteger`, `ChangeTypeToQuantity`, `ChangeTypeToString`, `ChangeTypeToTime`, `ChangeTypeToList`, `Quantity`, `BinaryOperator`, `AggregateOperator`, `QueryLetRef`, `AliasRef`, `Null`

# Other Work
## VS Hack
* Temporary Hack: New Visual Studio version discards MSBuild variables from the launchSettings file when running the PackagerCLI. The relative folders to the `$(SolutionDir)` is no longer filled out correctly. Temporary hack at the start of the `Main` method, to find the solution dir and prepend it to those paths if necessary.

## Use `SelectToArray` extension methods
* Added overloads for `SelectToArray` which includes the ordinal in the selector, e.g. `col.SelectToArray( (item, ordinal) => $"{ordinal}: {item}" )`
* Favor calling `col.SelectToArray()` over `col.ToArray()` for any readonly collection with a known `Count` or `Length`. (More efficient having the target array initialized to the correct length).
* Favor calling `enumerable.SelectToArray(count)` over `enumerable.ToArray()` when the `Count()` of the enumerable is known without having to enumerate it first. 

## Do not build Measures* projects
* Remove the `Measures` and `Measures-cms` projects from being debug build in the solution. This should be added back later.


Cql/Cql.Abstractions/Infrastructure/EnumerableExtensions.cs
```diff
@@ -24,29 +24,46 @@ public static void AddRange<T>(this ICollection<T> target, IEnumerable<T> source
         }
     }
 
+    public static T[] SelectToArray<TIn, T>(
```
ewoutkramer, line 27: What kind of speed improvements are we looking for?  If this is to avoid multiple enumerations, I have another option that will cache the contents of an IEnumerable, so the second iteration is much faster.  Otherwise, is this so much faster than IEnumerable + ToArray() that we see a difference at expression build/compile time?

Cql/Cql.Compiler/CqlOperatorsBinding.Bind.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Reflection;
+using Hl7.Cql.Operators;
+using Hl7.Cql.Primitives;
+
+namespace Hl7.Cql.Compiler;
+
+#pragma warning disable CS1591
+partial class CqlOperatorsBinding
+{
+    private static readonly IReadOnlyDictionary<string, MethodInfo[]> ICqlOperators_MethodInfos_By_Name =
```
ewoutkramer, line 14: oh yeah!!!

Cql/Cql.Compiler/CqlOperatorsBinding.Bind.cs
```diff
@@ -0,0 +1,121 @@
...
+        typeof(ICqlOperators)
+        .GetMethods(BindingFlags.Instance | BindingFlags.Public)
+        .GroupBy(m => m.Name)
+        .ToDictionary(m => m.Key, m => m.ToArray());
+
+    private static readonly ConstantExpression NullObject_ConstantExpression = Expression.Constant(null, typeof(object));
```
ewoutkramer: You have one in CqlContextExpressions -- don't you want to re-use that one?

Cql/Cql.Compiler/CqlOperatorsBinding.Bind.cs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    private static MethodCallExpression BindToMethod(
+        string methodName,
+        params Expression[] arguments) =>
+        Expression.Call(CqlContextExpressions.Operators_PropertyExpression, methodName, null, arguments);
+
+    private static MethodCallExpression BindToMethod(
+        string methodName,
+        Type[] typeArguments,
```
ewoutkramer: For readability, can we change the name `typeArguments` to `generic(Type)Arguments`?

Cql/Cql.Compiler/CqlOperatorsBinding.Bind.cs
```diff
@@ -0,0 +1,121 @@
...
+{
+    private static readonly IReadOnlyDictionary<string, MethodInfo[]> ICqlOperators_MethodInfos_By_Name =
+        typeof(ICqlOperators)
+        .GetMethods(BindingFlags.Instance | BindingFlags.Public)
+        .GroupBy(m => m.Name)
+        .ToDictionary(m => m.Key, m => m.ToArray());
+
+    private static readonly ConstantExpression NullObject_ConstantExpression = Expression.Constant(null, typeof(object));
+
+    private MethodCallExpression? BindToMethodConvertArgsOrNull(
```
ewoutkramer: Would you mind adding a bit of comments to document what this method is trying to achieve?

Cql/Cql.Compiler/ExpressionBuilder.ComparisonOperators.cs
```diff
@@ -43,21 +42,15 @@ protected Expression Equal(Expression left, Expression right)
                 }
                 else if (right.Type == typeof(string))
                 {
-                    var call = BindCqlOperator(CqlOperator.EnumEqualsString,
-                        Expression.Convert(left, typeof(object)),
-                        right);
-                    return call;
+                    return _operatorBinding.BindToMethod(CqlOperator.EnumEqualsString, Expression.Convert(left, typeof(object)), right);
```
ewoutkramer, line 45: These conversions should eventually be moved to your god binding method I think.

Cql/Cql.Compiler/ExpressionBuilder.Query.cs
```diff
@@ -203,8 +203,8 @@ protected Expression Query(Query query)
 
     private Expression DemoteSourceListToSingleton(Expression source)
     {
-        source = BindCqlOperator(CqlOperator.Single, source);
-        return source;
+        // Do not inline this method
```
ewoutkramer: How? Why?

Cql/Cql.Operators/CqlOperators.ClinicalOperators.cs
ewoutkramer, line 1: Binary file not shown??

ewoutkramer (changes requested):
I have no issues with this PR. Just a few questions and remarks. Otherwise, this was the work my hands were itching to take on, happy we're taking the first step!

Cql/Cql.Abstractions/Infrastructure/EnumerableExtensions.cs
baseTwo: This is not so much speed optimization, but rather memory optimization. ToArray on an enumerable builds array starting from length=8, then copying it into another array double the size. And eventually copying it into an array the correct size. The .net ToArray does optimize for IListProvider or ICollection<T>, but not for IReadOnlyCollection<T>.

Cql/Cql.Compiler/CqlOperatorsBinding.Bind.cs
baseTwo: I think I will also rename `BindToMethod` to `BindToGenericMethod`

Cql/Cql.Compiler/ExpressionBuilder.ComparisonOperators.cs
baseTwo: I'll create a second issue for this. I'm considering merging the CqlOperatorBinding into ExpressionBuilder.

baseTwo: Issue #297

Cql/Cql.Operators/CqlOperators.ClinicalOperators.cs
baseTwo: Not sure why GitHub thinks this is a binary file. Perhaps it was stored with incorrect encoding before?

Cql/Cql.Compiler/ExpressionBuilder.Query.cs
baseTwo: Provided more detail in the comment

Cql/Cql.Abstractions/Infrastructure/EnumerableExtensions.cs
baseTwo: Let's have a chat around this on Slack

Cql/Cql.Compiler/CqlOperatorsBinding.Bind.cs
baseTwo: Comments added. Also broke up this method into two - one for resolving the method info and converted arguments, and the other for creating the call expression

Cql/Cql.Operators/CqlOperators.ClinicalOperators.cs
ewoutkramer: I cannot even view it, because it is binary!

ewoutkramer (dismissed):

ewoutkramer (approved):

Issue #292, opened 2024-04-15
2.0/Improve conversions handling


baseTwo:
Conversions are scattered everywhere and should preferably only happen in two places, binding to cql operator method when arguments are matched/converted against existing methods, and then again on the return type if necessary

This is a big task, so all work will be done across several PR's and on a side branch [develop-2.0-bindings](https://github.com/FirelyTeam/firely-cql-sdk/tree/develop-2.0-bindings)

## TODO's
- [x] Synchronize names between ICqlOperators methods, and the values on the CqlOperator enum
- [x] There is no need to call methods`...OrNull(...)` e.g. `SelectOrNull(...)`, since nullability hints are available in the result type of the method. (Causes all generated c# output to change)
- [x] Park the Choice type generator in another branch, until needed

Comments from PR #303
- [x] Fix the **unnecessary casts** from a value type to its nullable value type e.g. [var o_ = context.Operators.Quantity((decimal?)40m, "%");](https://github.com/FirelyTeam/firely-cql-sdk/pull/303/files/4f159df9d414193481f52fcde28622f2bd93cd13#diff-b2645f821ef513fac4fbc49440ca1775a70e602e730606efee77c39e4021df5f)
- [x] The generated code that handles **choice elements** is wrong.  These were treated as DateType in the old code, but are now turned into an object. [Instead of calling Convert on them (which invoked specific logic to turn a DateType into a string), this has changed into a as string now](https://github.com/FirelyTeam/firely-cql-sdk/pull/303/files#r1582645244), which will not invoke the specific logic in the TypeConverter.
- [ ] Converting **Enums** to Strings in [CqlOperatorsBinding.Conversions.cs](https://github.com/FirelyTeam/firely-cql-sdk/pull/303/files/4f159df9d414193481f52fcde28622f2bd93cd13#diff-35c58660127ea2de46a7c2d9feb58eaa3266228838112b54435ec5066a26e36a) : This is not good enough for the fhir pocos, and we should involve the model, probably by asking the specific type converter to convert the enum to a string (e.g. in the firely pocos, you need to call GetLiteral() on an enum value to get to the code).
[...]

PR #303, opened 2024-04-16
2.0/Work on CQL bindings part 1

baseTwo:
ℹ️Work for issue #292
⏭️ PR #306

### Cleanup
* Removed unused namespaces
* Removed unused `else` keywords

### Changes
* Renamed `CqlContextExpressions` to `CqlExpressions`, and made it static
* Renamed `OperatorBinding` to `OperatorsBinder`
* Renamed `CqlOperatorsBinding` to `CqlOperatorsBinder` (the filenames still have to be changed as well)
* Changed `CqlOperatorsBinder` dependencies from internal properties to private fields

### Elements with DateTime Precision 
* Added interface `IGetDateTimePrecision`, and apply on all ELM Element types that have this

### `CqlOperatorsBinder`
* Moved the `Equals` method from `ExpressionBuilder` to `CqlOperatorsBinder`. It has changed from a simple method bind to checking conversions before binding.
* Moved methods for converting or casting from the `ExpressionBuilder` to `CqlOperatorsBinder`: `ConvertToType` and `CastToType`
* Removed following methods: `Round`
* Methods are resolved based on whether a match is found on the arguments directly or via conversions. If multiple candidates are found, they are scored based on how closely the arguments were matched, or if there were no arguments, on the result type.
* `CqlOperator`'s that were removed:
  - `IntervalAfterInterval`, `IntervalAfterElement`, `ElementAfterInterval` (all goes through `After`)
  - `IntervalBeforeInterval`, `IntervalBeforeElement`, `ElementBeforeInterval` (all goes through `Before`)

### `ExpressionBuilder`
* Simplified and removed following methods: `Round`, `CalculateAge`, `CalculateAgeAt`, `InValueSet`, `AnyInValueSet`, `ExpandValueSet`,  `Equal` (moved to `CqlOperatorsBinder`), `NotEqual`, `After`, `Before`, `DateTimeComponentFrom`, `UnaryOperator`, `NaryOperator`, `DateTime`, `Date`, `LastPositionOf`, `PositionOf`, `Ratio`, `Round`, `Time`
* Moved `IsInterval` to `TypeExtensions.IsCqlInterval`
* Since many element types implement `IGetDateTimePrecision`, the `Precision` methods was changed to take in the element, instead of the two property components.

### `ContextBinder`
Another dependency was added for the expression builders, `ContextBinder`, which binds method call expressions to `CqlContext`

### Expression Building Dependencies
Since the list of types is getting cumbersome to pass along, they have all been wrapped up inside `ExpressionBuildingDependencies`. This new type is created by the `CqlCompilerFactory` and passed to `LibraryDefinitionsBuilder`, then `LibrarySetExpressionBuilder`, `LibraryExpressionBuilder` and finally `ExpressionBuilder`. Each builder type in created by the one before, and they copy the dependencies they need into private fields. See the new service dependencies diagram below.

### `TypeExtensions`
Some type checking utilities scattered around the codebase were consolidated into 
* `Hl7.Cql.Abstractions.Infrastructure.TypeExtensions`
  - `IsNullableValueType` (I added this back, because it is just a lot more readable than otherwise calling `Nullable.GetUnderlyingType()` then checking whether is returns null or not.
  - `IsNullable` - nullable value types, or reference types
  - `IsEnum`
  - `IsImplementingGenericTypeDefinition`
* and `Hl7.Cql.Compiler.Infrastructure.TypeExtensions`
  - `IsCqlInterval`

## Service Dependencies (excl Logger and Options)

```mermaid
classDiagram
    direction BT

    namespace CSharpCode_Generate_And_Compile {
        class AssemblyCompiler {
%%            Compile(librarySet : LibrarySet,definitions : DefinitionDictionary~LambdaExpression~? = null) IDictionary~string, AssemblyData~
        }

        class CSharpLibrarySetToStreamsWriter {
        }

        class CSharpCodeStreamPostProcessor {
            ProcessStream(name : string, stream : Stream) void
        }

        class WriteToFileCSharpCodeStreamPostProcessor {
        }
    }

    namespace Expression_Building {
        class Expression{
        }

        class Library{
        }

        class LibrarySet{
        }

        class ExpressionBuildingDependencies{
        }

        class LibraryDefinitionBuilderSettings {
        }

        class ExpressionBuilder{
        }

        class LibraryExpressionBuilder{
        }

        class LibrarySetExpressionBuilder{
        }

        class LibraryDefinitionsBuilder {
%%            ProcessLibrarySet(librarySet : LibrarySet) DefinitionDictionary<LambdaExpression>
        }

        class OperatorsBinder {
        }

        class CqlOperatorsBinder {
        }

        class TypeConverter {
        }

        class ModelInspector {
        }
    }

    namespace Fhir_Resource_Building {
        class ResourcePackager {
%%            PackageResources(elmDirectory : DirectoryInfo, cqlDirectory : DirectoryInfo, resourceCanonicalRootUrl : string? = null) IReadOnlyCollection~Resource~
        }

        class FhirResourcePostProcessor {
%%            ProcessResource(resource : Resource) void
        }

        class WriteToFileFhirResourcePostProcessor {
        }
    }

    namespace Cql_To_Resource_Pipeline {
        class CqlToResourcePackagingPipeline {
        }        
    }

    namespace Application {
        class PackagerCliProgram {
        }

        class OptionsConsoleDumper {
        }
    }

%%    namespace Dependencies {
        class TypeManager {
        }

    
        class TypeResolver {
        }
%%    }

    %% Inheritance  

   
    CqlOperatorsBinder --> OperatorsBinder : inherits
    WriteToFileCSharpCodeStreamPostProcessor --> CSharpCodeStreamPostProcessor : inherits
    WriteToFileFhirResourcePostProcessor --> FhirResourcePostProcessor : inherits

    %% Injected Dependencies

    CSharpCodeStreamPostProcessor ..> AssemblyCompiler : injected\n(optional)
    CSharpLibrarySetToStreamsWriter ..> AssemblyCompiler : injected
    
    Expression ..> ExpressionBuilder : processed by

    LibrarySet ..> LibrarySetExpressionBuilder : processed by
    LibraryExpressionBuilder ..> LibrarySetExpressionBuilder : created by

    AssemblyCompiler ..> CqlToResourcePackagingPipeline : injected
    LibraryDefinitionsBuilder ..> CqlToResourcePackagingPipeline : injected
    ResourcePackager ..> CqlToResourcePackagingPipeline : injected 
    
    CqlToResourcePackagingPipeline ..> PackagerCliProgram : injected
      
    TypeResolver ..> CqlOperatorsBinder : injected
    TypeConverter ..> CqlOperatorsBinder : injected

    ExpressionBuildingDependencies ..> LibraryDefinitionsBuilder : injected
    LibrarySetExpressionBuilder ..> LibraryDefinitionsBuilder : created by

    ILoggerFactory ..> ExpressionBuildingDependencies : injected
    OperatorsBinder ..> ExpressionBuildingDependencies : injected
    TypeManager ..> ExpressionBuildingDependencies : injected
    TypeConverter ..> ExpressionBuildingDependencies : injected
    TypeResolver ..> ExpressionBuildingDependencies : injected (via TypeManager)
    LibraryDefinitionBuilderSettings ..> ExpressionBuildingDependencies : injected (static)

    ExpressionBuilder ..> LibraryExpressionBuilder : created by
    Library ..> LibraryExpressionBuilder : processed by

    TypeResolver ..> ResourcePackager : injected
    FhirResourcePostProcessor ..> ResourcePackager : injected\n(optional) 
    
    TypeResolver ..> CSharpLibrarySetToStreamsWriter : injected
```


Cql/Cql.Abstractions/TypeResolver.cs
```diff
@@ -112,22 +112,26 @@ public abstract class TypeResolver
...
-        /// <remarks>
-        /// This is typically used in conjunction with <see cref="GetListElementType(Type, bool)"/>, where
-        /// <paramref name="genericInterfaceTypeDefinition"/> is <see cref="IEnumerable{T}"/>.
-        /// </remarks>
         /// <param name="type">The type to check.</param>
-        /// <param name="genericInterfaceTypeDefinition">The generic type definition.</param>
-        /// <returns><c>true</c> if the type implements the generic interface; otherwise, <c>false</c>.</returns>
-        public abstract bool ImplementsGenericInterface(Type type, Type genericInterfaceTypeDefinition);
+        /// <returns><see langword="true"/> if the type is a list type; otherwise, <see langword="false"/>.</returns>
+        internal virtual bool IsListType(Type type)
```
ewoutkramer, line 127: Why didn't this move to TypeExtensions?

Cql/Cql.Compiler/ContextBinder.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Linq.Expressions;
+
+namespace Hl7.Cql.Compiler;
+
+/// <summary>
+/// Represents an abstract class for binding context information.
```
ewoutkramer, line 6: What is the "context" here? I see no reference to a context in the parameters.

Cql/Cql.Compiler/CqlOperatorsBinding.Bind.cs
```diff
@@ -3,147 +3,194 @@
...
+                                 return score;
+                             })
+                             .SelectToArray(candidates.Length, o => o);
+                return (candidates[0].method, candidates[0].arguments);
+            }
+
+            if (resultTypeHint is { } t)
+            {
+                // Find best match on return type
+                throw new InvalidOperationException("");
```
ewoutkramer, line 94: This is a todo I guess, to see this actually happens?

Cql/Cql.Compiler/CqlOperatorsBinding.Conversions.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                // Null values to a type that can accept nulls e.g. default(string) or default(int?)
+                result = (NullExpression.ForType(to), TypeConversion.ExactType);
+                return true;
+            }
+
+            if (fromConstant.Value is Enum enumValue && to == typeof(string) )
+            {
+                // Enum values to lowercase string e.g DateTimePrecision.Year -> "year"
+                var name = Enum.GetName(enumValue.GetType(), enumValue);
```
ewoutkramer, line 44: This is not good enough for the fhir pocos, and we should involve the model, probably by asking the specific type converter to convert the enum to a string (e.g. in the firely pocos, you need to call GetLiteral() on an enum value to get to the code).

Cql/Cql.Compiler/CqlOperatorsBinding.Specific.cs
```diff
@@ -5,36 +5,37 @@
...
     private Expression Expand(
         Expression argument,
         Expression perQuantity)
     {
         if (perQuantity is ConstantExpression { Value: null })
-            perQuantity = Expression.Constant(null, typeof(CqlQuantity));
+            perQuantity = NullExpression.ForType<CqlQuantity>();
 
-        if (TypeResolver.ImplementsGenericIEnumerable(argument.Type))
+        if (_typeResolver.IsListType(argument.Type))
```
ewoutkramer, line 25: The next few lines look like argument conversion/binding to me, but I guess this is a todo for the next phase?

Cql/Cql.Compiler/ExpressionBuilder.DateAndTimeOperators.cs
```diff
@@ -8,170 +8,48 @@
...
-                if (!IsInterval(right.Type, out var rightElementType))
-                    throw this.NewExpressionBuildingException();
-
-                return _operatorBinding.BindToMethod(CqlOperator.IntervalSameOrBefore, left, right, precision);
-            }
-
-            return _operatorBinding.BindToMethod(CqlOperator.SameOrBefore, left, right, precision);
-        }
-
-        protected Expression Time(Elm.Time e)
```
ewoutkramer, line 168: Time is now handled by the general algorithm?

Cql/Cql.Compiler/ExpressionBuilder.cs
```diff
@@ -95,195 +95,232 @@ partial class ExpressionBuilder
...
+            params object?[] args) =>
+            _operatorsBinder.BindToMethod(@operator, resultTypeHint, TranslateAll(args));
+
+        private Expression BindCqlOperator<T>(
+            CqlOperator @operator,
+            Type? resultTypeHint,
+            params T?[] args) =>
+            _operatorsBinder.BindToMethod(@operator, resultTypeHint, TranslateAll(args));
+
+        private Expression[] TranslateAll(params object?[] args) =>
```
ewoutkramer, line 115: Is this a temporary trick? object here is either a Linq.Expression or an Elm.Element, if I understand it correctly, but in the end, it's probably all Elm.Element, no more translations done on "this side" of the builder?

CqlAndDemo.sln.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
ewoutkramer, line 1: what's this?

Demo/Measures-cms/AHAOverall-2.6.000.g.cs
```diff
@@ -385,7 +384,7 @@ private IEnumerable<object> Moderate_or_Severe_LVSD_Findings_Value()
 		bool? c_(Observation EjectionFraction)
 		{
 			var n_ = FHIRHelpers_4_3_000.ToValue(EjectionFraction?.Value);
-			var o_ = context.Operators.Quantity(40m, "%");
+			var o_ = context.Operators.Quantity((decimal?)40m, "%");
```
ewoutkramer, line 387: That's superfluous, any idea why this happens now?

Demo/Measures-cms/FHIRHelpers-4.3.000.g.cs
```diff
@@ -676,856 +704,1704 @@ object a_()
 		throw (new NotImplementedException("External function FHIRHelpers-4.3.000.htmlChecks is not implemented."));
 
     [CqlDeclaration("ToString")]
-	public Account.AccountStatus? ToString(Code<Account.AccountStatus> value) => 
-		value?.Value;
+	public string ToString(Code<Account.AccountStatus> value)
```
ewoutkramer, line 707: The new algorithm actually translates to a string instead of translating a Code<T> to an enum That should be an improvement, but maybe some parts of the code was dependent on this. Something to keep in mind if we see difference in behaviour when dealing with enums.

Demo/Measures-cms/FHIRHelpers-4.3.000.g.cs
```diff
@@ -676,856 +704,1704 @@ object a_()
...
+	{
```
ewoutkramer, line 708: Ah, maybe not - remember that we intercept "ToString" somewhere in the binding because this ToString here was not correct - maybe now it it!

Demo/Measures-cms/SupplementalDataElements-3.4.000.g.cs
```diff
@@ -153,42 +153,43 @@ Tuples.Tuple_DMgHTLENEHBHWJISQgKZGZVMB f_(Extension E)
...
-			var ai_ = new Tuples.Tuple_DMgHTLENEHBHWJISQgKZGZVMB
+			object ae_(Extension @this) => 
+				@this?.Value;
+			var af_ = context.Operators.SelectOrNull<Extension, object>(ad_, ae_);
+			var ag_ = context.Operators.SingleOrNull<object>(af_);
+			var ah_ = new Tuples.Tuple_DMgHTLENEHBHWJISQgKZGZVMB
 			{
 				codes = ab_,
-				display = ah_,
+				display = (ag_ as string),
```
ewoutkramer, line 189: This requires scrutiny: a DataType has become an object, then in the old code the datatype is converted using Convert to a string, then assigned to display. In the new code, this object is just cast to a string, without call to Convert - which is almost certainly wrong, unless the SelectOrNull converts this - but the lambda ae_ does not seem to do that.

Demo/Measures/ColorectalCancerScreeningsFHIR-0.0.003.g.cs
```diff
@@ -688,19 +689,20 @@ private IEnumerable<Condition> Total_Colectomy_Condition_Value()
 		var k_ = AdvancedIllnessandFrailtyExclusionECQMFHIR4_5_17_000.Advanced_Illness_and_Frailty_Exclusion_Not_Including_Over_Age_80();
 		var l_ = context.Operators.Or(j_, k_);
 		var m_ = this.Patient();
-		var n_ = context.Operators.Convert<CqlDate>(m_?.BirthDateElement?.Value);
+		var n_ = context.Operators.Convert<CqlDateTime>(m_?.BirthDateElement?.Value);
```
ewoutkramer, line 692: Is this expected? A CqlDate becoming a CqlDateTime? Do we chose another overload?

ewoutkramer (approved):
I found one thing in the generated code that is probably wrong, which has to do with the handling of choice elements. These were treated as DateType in the old code, but are now turned into object. Then, instead of calling Convert<string> on them (which invoked specific logic to turn a DateType into a string), this has changed into a `as string` now, which will not invoke the specific logic in the TypeConverter, so I think that is wrong.

Cql/Cql.Abstractions/TypeResolver.cs
baseTwo: Because it's virtual, and what is considered a list can be more specific (eg in the derived `FhirTypeResolver`, types with an `FhirTypeAttribute` are excluded)

Cql/Cql.Compiler/CqlOperatorsBinding.Bind.cs
baseTwo: I discovered I didn't need the result type anyway, so it was removed in the latest pr #308

Cql/Cql.Compiler/ContextBinder.cs
baseTwo: I think I'll remove that abstract class and just have a CqlContextBinder. Likewise, I don't thing we need a base type for CqlOperatorsBinder either.

Demo/Measures-cms/AHAOverall-2.6.000.g.cs
baseTwo: Not sure why it does that yet. Will tidy this up later #292

Demo/Measures/ColorectalCancerScreeningsFHIR-0.0.003.g.cs
baseTwo: Fixed in latest PR
image

Cql/Cql.Compiler/CqlOperatorsBinding.Specific.cs
baseTwo: Completed in the latest PR #308

Cql/Cql.Compiler/ExpressionBuilder.DateAndTimeOperators.cs
baseTwo: Yes, only a few handful of cases still remain in #308

Cql/Cql.Compiler/CqlOperatorsBinding.Conversions.cs
baseTwo: Added to todo's #292

Demo/Measures-cms/FHIRHelpers-4.3.000.g.cs
baseTwo: Added a todo #292

Cql/Cql.Compiler/ContextBinder.cs
baseTwo: Added to todo #292

Cql/Cql.Compiler/ExpressionBuilder.cs
baseTwo: No, it is not just temporary? Ultimately the binder `BindToMethod` expects an array of `Linq.Expression` as input, so the expression builder still need to traverse the `Elm.Element` into those. `object` is directly put into `Linq.ConstantExpression`

CqlAndDemo.sln.DotSettings
baseTwo: ReSharper settings snuck into the repo. I'll remove this #292

Demo/Measures-cms/SupplementalDataElements-3.4.000.g.cs
baseTwo: Added to #292

Demo/Measures-cms/AHAOverall-2.6.000.g.cs
baseTwo: Fixed in PR #308 
image

Demo/Measures-cms/SupplementalDataElements-3.4.000.g.cs
baseTwo: Fixed in #308 
image

CqlAndDemo.sln.DotSettings
baseTwo: Removed in PR #308

Demo/Measures-cms/FHIRHelpers-4.3.000.g.cs
baseTwo: Back to how it was in #308 
image

Demo/Measures-cms/HIVRetentionFHIR-0.1.000.g.cs
```diff
@@ -311,42 +310,43 @@ Tuples.Tuple_DMgHTLENEHBHWJISQgKZGZVMB f_(Extension E)
...
 					? ((E as Element).Extension)
 					: null), ac_);
-			var af_ = context.Operators.SelectOrNull<Extension, DataType>(ad_, p_);
-			var ag_ = context.Operators.SingleOrNull<DataType>(af_);
-			var ah_ = context.Operators.Convert<string>(ag_);
-			var ai_ = new Tuples.Tuple_DMgHTLENEHBHWJISQgKZGZVMB
+			object ae_(Extension @this) => 
+				@this?.Value;
+			var af_ = context.Operators.SelectOrNull<Extension, object>(ad_, ae_);
+			var ag_ = context.Operators.SingleOrNull<object>(af_);
```
baseTwo, line 342: Not going through Operators.Convert<string>

Demo/Measures-cms/FHIRHelpers-4.3.000.g.cs
baseTwo: Fixed #308

Cql/Cql.Compiler/ExpressionBuilder.cs
ewoutkramer: Ok, so I just didn't understand what this does on first sight, I'll have to look deeper ;-)

Issue #309, opened 2024-04-22
2.0/Missing Url in Library Resource


baseTwo:
https://firelyteam.slack.com/archives/C021G7NAN2J/p1713802757993669

Library Resources no longer have the `Url` property assigned.

PR #310, opened 2024-04-22
Fix missing url on library resource

baseTwo:
ℹ️Fix for bug #309 

Somehow writing out the url for a library resource got deleted at some point in the 2.0 branch. Added it back.
Also, extracted the code to create a measure into its own method, just like library has its own method


baseTwo:
Closed, branched off an old version od develop-=2.0

Issue #321, opened 2024-04-30
2.0/Copyright file header


baseTwo:
Add file headers to each C# file. No need to update older dated copyrights, since a copyright date refers to the year in which a creative work, such as a book, music, or artwork, was first published or produced. [It serves as evidence of when the work was made public, aiding in legal matters](https://www.copyrighted.com/blog/copyright-date)

```
/*
 * Copyright (c) 2024, NCQA and contributors
 * See the file CONTRIBUTORS for details.
 *
 * This file is licensed under the BSD 3-Clause license
 * available at https://raw.githubusercontent.com/FirelyTeam/firely-cql-sdk/main/LICENSE
 */
```

PR #323, opened 2024-04-30
2.0/Continue work on cql bindings part 3

baseTwo:
ℹ️Work for issue #292
⏮️PR #321
⏭️PR #332 

* Added new project `MultiPackagerCLI` which will package both librarysets
* Added cleanup support to the factories via a CancellationToken
* Simplify `DefaultExpressions` in the `SimpleExpressionVisitor` (no need for defaults to be put into separate variables, use them as is)
* Disable simplifying `NullConditionalMemberExpression` when using exclicit types in c# output
* Expanded method overload resolution when providing generic type arguments
* Simplified `ToList`, `MinValue`, `MaxValue`, `Expand`, `Width`. Some special handling remains in the binder
* Simplified `ExpandValueSet`, however, the cql operator method was not renamed and kept to `CreateValueSetFacade` 
* Added support to write generic MethodInfo's to C# strings 
* C# String Formatter supports writing to TextWriter and StringBuilder. The default extension is now to write to a string, although it is still possible to write to TextWriter and StringBuilder via the format options.
* C# String Formatter supports writing out generic methods
* FhirTypeConverter adds support for FHIR Enums -> string
* Remove redundent `;` in generated code after a CaseWhenThenExpression  `if {...} else {...};`


Cql/Cql.Compiler/ExpressionBuilder.cs
```diff
@@ -122,6 +122,9 @@ public void ProcessParameterDef(ILibraryExpressionBuilderContext libCtx, Paramet
         public void ProcessExpressionDef(ILibraryExpressionBuilderContext libCtx, ExpressionDef expressionDef)
         {
             ExpressionBuilderContext context = new ExpressionBuilderContext(this, libCtx, new());
+            bool b0 = libCtx.LibraryKey == "SupplementalDataElements-3.4.000";
```
ewoutkramer: Is that remaining debug code?

Cql/Cql.Operators/CqlOperators.ClinicalOperators2.cs
```diff
@@ -165,7 +165,7 @@ internal partial class CqlOperators
         }
         #endregion
 
-        public IEnumerable<CqlCode> ResolveValueSet(CqlValueSet valueSet) => CreateValueSetFacade(valueSet);
+        public IEnumerable<CqlCode> ResolveValueSet(CqlValueSet valueSet) => ExpandValueSet(valueSet);
```
ewoutkramer: The original CreateValueSet|Facade had nothing to do with Expanding valuesets, I'll see further on I guess what has changed here.

Cql/Cql.Operators/CqlOperators.cs
```diff
@@ -171,10 +171,13 @@ public T Message<T>(T source, string code, string severity, string message)
...
 
-        public TAccumulate? Aggregate<TSource, TAccumulate>(IEnumerable<TSource?>? source, TAccumulate? seed, Func<TAccumulate?, TSource?, TAccumulate?> lambda) =>
+        public TAccumulate? Aggregate<TSource, TAccumulate>(
+            IEnumerable<TSource?>? source,
+            Func<TAccumulate?, TSource?, TAccumulate?> lambda,
+            TAccumulate? seed) =>
             source == null ? default : source.Aggregate(seed, lambda);
 
-        public IValueSetFacade CreateValueSetFacade(CqlValueSet valueSet) => ValueSets.GetValueSet(valueSet);
+        public IValueSetFacade ExpandValueSet(CqlValueSet valueSet) => ValueSets.GetValueSet(valueSet);
```
ewoutkramer: I don't think this is the right name - Expanding a valueset is a FHIR operation that turns intensional valuesets into extensional valuesets - it might be that deepdown in the infrastructure this is happening when ValueSets.GetValueSet is called, but that would be an implementation detail, so I don't think this is a good name change.

Demo/Measures-cms/AHAOverall-2.6.000.g.cs
```diff
@@ -285,28 +285,32 @@ private Patient Patient_Value()
...
+			var y_ = context.Operators.ConvertCodeToConcept(x_);
+			var z_ = context.Operators.Equivalent(w_, y_);
+			var aa_ = context.Operators.Or(u_, z_);
+			var ab_ = context.Operators.Not(aa_);
+			var ac_ = context.Operators.And(j_, ab_);
+
+			return ac_;
 		};
-		var c_ = context.Operators.Where<Condition>(a_, b_);
+		var c_ = context.Operators.Where<Condition>((IEnumerable<Condition>)a_, b_);
```
ewoutkramer, line 313: Is this cast expected?

ewoutkramer (changes requested):

Cql/Cql.Operators/CqlOperators.ClinicalOperators2.cs
baseTwo: This must have been done in error. Renamed it back

Cql/Cql.Operators/CqlOperators.cs
baseTwo: Ok, then this would have to be an exception on the name matching between the Elm.ExpandValueSet and the ICqlOperator method for it

Demo/Measures-cms/AHAOverall-2.6.000.g.cs
baseTwo: It isn't required, but there is still breaking changes happening elsewhere if this cast is removed globally

Cql/Cql.Operators/CqlOperators.cs
ewoutkramer: It could be it is a real operation - and since the current implementation of the valueset system *always* supplies expanded valuesets, Evan decided to hardwire the operation to just returning the valueset because of that....

ewoutkramer (approved):

PR #324, opened 2024-04-30
Multi library support

EvanMachusak:
- Refinements to signature resolution:
  - During generic inference, we will only infer `T` to be `System.Any` if there are no other compatible inferences.  Even in cases where inferring `T` to be some type `U` with `U <> System.Any` whose cost is higher than inferring `T` to be `System.Any`, we would infer `T = U`.  This prevents type inference from aggressively picking `Any` in too many cases.
  - Generic inference now considers model conversion functions, e.g. `ToInterval(FHIR.Period) Interval<DateTime>`, so that for functions like  `end of(argument Interval<T>) T`, we can infer `T` to be `System.DateTime` for `end of period` where `period` is a `FHIR.Period`.
    - There are no model conversions that create lists currently, but this pattern would need to be implemented for list types also; that is not done in this PR.
  - Coercion cost computation now considers list and interval conversions immediately after `ExactMatch` analysis.  The cost to coerce `List<U>` to `List<V>` is computed as the cost to coerce `U` to `V`; this same logic applies to interval point types as well.
  - `Compatible` has been split into `LessCompatible` and `MoreCompatible`, where `MoreCompatible` is preferential to `LessCompatible`.  Coercions from `Any` to `List` or `Interval` types are considered `LessCompatible` than coercions to other values.
  - Coercion from `List<U>` to `List<V>` is now implemented through a query pattern.
  - List promotion and demotion now considers list "degree":
    - Non-list types have a degree of 0
    - `List<T>` has a degree of 1
    - `List<List<T>>` has a degree of 2
  - Only where the degree of the list differs by 1 can a promotion or demotion be performed.  While it's possible to recursively use the query pattern to go from `List<List<List<T>>>` to `T` during a `ListDemotion` event, this behavior should be controlled by a currently non-existent option.
  - Model-defined conversion functions (e.g., in `FHIRHelpers` will now be considered during `ImplicitConversion` coercion costs, and those coercions will be implemented by creating a `FunctionRef` to those conversion functions.

- Added multi-library support:
  - Definitions can now be resolved across libraries
  - Library providers (memory, file system) were added; more testing is needed on these
  - A rudimentary `StreamInspector` class that attempts to detect and disambiguate CQL, JSON ELM, and XML ELM was implemented to identify libraries regardless of specific file naming conventions; this would also be useful in networking use cases.
 
- Implemented HL7 XML tests
  - Created a dynamic data test that reads HL7 XML test files from a test folder, compiles the expression & expectation ELM, executes, and compares the results:
    - 1472 passed
    - 50 skipped
  - Removed `HL7Tests` from the `CoreTests` project as they are now redundant.


ewoutkramer:
Sounds like a ton of good improvements.

Cql/Cql.Compiler/ExpressionBuilder.cs
```diff
@@ -2035,26 +2373,59 @@ protected Expression FunctionRef(FunctionRef op, ExpressionBuilderContext ctx)
...
-                9 => typeof(Func<,,,,,,,,>).MakeGenericType(funcTypeParameters),
-                10 => typeof(Func<,,,,,,,,,>).MakeGenericType(funcTypeParameters),
-                11 => typeof(Func<,,,,,,,,,,>).MakeGenericType(funcTypeParameters),
-                12 => typeof(Func<,,,,,,,,,,,>).MakeGenericType(funcTypeParameters),
-                13 => typeof(Func<,,,,,,,,,,,,>).MakeGenericType(funcTypeParameters),
-                14 => typeof(Func<,,,,,,,,,,,,,>).MakeGenericType(funcTypeParameters),
-                15 => typeof(Func<,,,,,,,,,,,,,,>).MakeGenericType(funcTypeParameters),
-                _ => throw new NotSupportedException("Fucntions with more than 15 parameters are not supported."),
-            };
+            Type? funcType;
```
ewoutkramer: What happened here? I guess an unintended refactor from an switch expression to a statement?

Cql/Cql.CqlToElm/LibraryProviders/StreamInspector.cs
```diff
@@ -0,0 +1,157 @@
...
+            }
+            catch { return null; }
+        }
+        private Library? FromXml(Stream reader) => null;
+
+        private Regex LibraryWithVersion = new Regex("^library\\s+(?'name'\\S+)(\\s+version\\s+'(?'version'\\d+(\\.\\d+)*)')$", RegexOptions.Compiled);
+        private Regex LibraryWithoutVersion = new Regex("^library\\s+(?'name'\\S+)$", RegexOptions.Compiled);
+        internal VersionedIdentifier? FromCql(StreamReader reader)
+        {
+            // TODO: use ANTLR?
```
ewoutkramer: No, Pidgin (https://github.com/benjamin-hodgson/Pidgin) or Sprache (https://github.com/sprache/Sprache) would be perfect for this task. Let me know, I have worked with these tools before.

Cql/Cql.CqlToElm/ModelProvider.cs
```diff
@@ -14,10 +14,10 @@ public static class ModelProvider
         /// Returns the <see cref="ModelInfo"/> for a model, given the model's uri and version.
         /// </summary>
         /// <exception cref="ArgumentException">There is no model with the given uri.</exception>
-        public static Model.ModelInfo GetModelFromUri(this IModelProvider provider, string uri, string? version = null) =>
+        public static Model.ModelInfo? GetModelFromUri(this IModelProvider provider, string uri, string? version = null) =>
```
ewoutkramer: In public interfaces, I prefer using `TryXXXX` instead of returning null to signal failure.

Cql/Cql.CqlToElm/ModelProvider.cs
```diff
@@ -176,8 +176,16 @@ public static bool TryGetTypeInfoFor(this ModelInfo model, string name, out Type
         internal static bool TryFindTypeInfoByName(this IModelProvider provider, string uri,
             string typeName, out TypeInfo? typeInfo, out ModelInfo? model)
         {
-            return provider.TryGetModelFromUri(uri, out model) &
-                    model!.TryGetTypeInfoFor(typeName, out typeInfo);
+            if (provider.TryGetModelFromUri(uri, out model)
```
ewoutkramer, line 169: A try function that returns null if it is succesful sounds like a bad design on which we will trip later on. Let me take a look what's going on...

ewoutkramer: No, it cannot happen, maybe you just wanted the nullability warning to go away. We can do that too, by adding [NotNullWhen(true)] to the definition of the TryGetModelFromUri function.

ewoutkramer (approved):
Just a few non-blocking comments, otherwise really nice work.

Cql/Cql.CqlToElm/LibraryProviders/StreamInspector.cs
ewoutkramer: Then again, Antlr already has the syntactic rules for identifiers and versions too, so that's worth trying out indeed.

ewoutkramer: Turned out to be super easy.

Cql/Cql.CqlToElm/ModelProvider.cs
ewoutkramer: Actually....there is already a TryGetModelFromUri that does not throw - so I think this change does not make sense, just call that method instead.

ewoutkramer: And then this method became unused, so I have removed it.

Cql/Cql.Compiler/ExpressionBuilder.cs
EvanMachusak: I pulled in the changes from the `develop` branch with this PR as well, since we were many commits behind.  This must have been made to `develop` some time after we branched the main `cql-to-elm` branch.

PR #334, opened 2024-05-14
2.0/revise build for demo

richfirely:
I attempted to make the demo so it only runs the maven download, cql-elm, and then elm-csharp on demand. It's close, but I didn't figure out elm-csharp so it still runs every time. I left the CS files where they are, but I think they could be moved under a folder in a later PR.


Demo/CLI/Program.cs
```diff
@@ -26,7 +26,9 @@ public static int Main(string[] args)
                 Console.Error.WriteLine("Missing required parameter: --d");
                 return ShowHelp();
             }
+#pragma warning disable CS8604 // Possible null reference argument.
```
baseTwo: Remove the pragmas, and just add an `!` behind the `dArg`

baseTwo:
I'm getting these errors while rebuilding? I already have the latest Java version

```
20>==============================================
20>CMS java -Djakarta.xml.bind.JAXBContextFactory=org.eclipse.persistence.jaxb.XMLBindingContextFactory -classpath C:\Dev\FirelyTeam\firely-cql-sdk\develop-2.0-binding-rich\Demo\Cql\Build\target/dependency/* org.cqframework.cql.cql2elm.cli.Main -input C:\Dev\FirelyTeam\firely-cql-sdk\develop-2.0-binding-rich\LibrarySets/CMS/Cql -f JSON -output C:\Dev\FirelyTeam\firely-cql-sdk\develop-2.0-binding-rich\LibrarySets/CMS/Elm -locators true -result-types true -signatures All
20>==============================================
20>Exception in thread "main" java.lang.UnsupportedClassVersionError: org/cqframework/cql/cql2elm/cli/Main has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
20>	at java.lang.ClassLoader.defineClass1(Native Method)
20>	at java.lang.ClassLoader.defineClass(Unknown Source)
20>	at java.security.SecureClassLoader.defineClass(Unknown Source)
20>	at java.net.URLClassLoader.defineClass(Unknown Source)
20>	at java.net.URLClassLoader.access$100(Unknown Source)
20>	at java.net.URLClassLoader$1.run(Unknown Source)
20>	at java.net.URLClassLoader$1.run(Unknown Source)
```

SOLUTION : Turns out I had eclipse installed and it had an older version of the sdk in the PATH variable.

baseTwo (approved):

Issue #340, opened 2024-05-20
2.0/CqlTypeToFhirTypeMapper for mapping between CqlCode and Coding is wrong


alexzautke:
**Describe the bug**
[CqlTypeToFhirTypeMapper.cs#L122](https://github.com/FirelyTeam/firely-cql-sdk/blob/0e1ff194ccc78dcab1bbf8084ebb2f437fb136b8/Cql/Cql.Packaging/CqlTypeToFhirTypeMapper.cs#L122) should map a CqlCode to a Coding. You can see that the mapping is wrong as a FHIR Coding is mapped to a CqlCode.

baseTwo:
Will fix in 2.0

PR #342, opened 2024-05-24
Map a CqlCode to a Coding and not Code

baseTwo:
ℹ️Work for issue #340 

Fix mapping from Cql Code -> FHIR Coding


alexzautke (changes requested):
I would be in favour of having an accurate mapping:
- If CqlCode only contains a code and not a system, version or display -> FHIRAllTypes.code
- Else -> FHIRAllTypes.Coding

baseTwo:
> I would be in favour of having an accurate mapping:
> 
> * If CqlCode only contains a code and not a system, version or display -> FHIRAllTypes.code
> * Else -> FHIRAllTypes.Coding

@alexzautke This mapper only works on type information, it does not have access to an instance of CqlCode

alexzautke (approved):

Issue #347, opened 2024-06-01
2.0/Equality between Code<T> and string


baseTwo:
**Describe the bug**
In at least one case in CMS measures `ChildandAdolescentMajorDepressiveDisorderMDDSuicideRiskAssessmentFHIR`, there is an equality between different types (Code<T> against string). The cql is included below, see the HERE comment)

```cql
define "Major Depressive Disorder Encounter":
  ( ["Encounter": "Office Visit"]
    union ["Encounter": "Outpatient Consultation"]
    union ["Encounter": "Psych Visit Diagnostic Evaluation"]
    union ["Encounter": "Psych Visit for Family Psychotherapy"]
    union ["Encounter": "Psych Visit Psychotherapy"]
    union ["Encounter": "Psychoanalysis"]
    union ["Encounter": "Group Psychotherapy"]
    union ["Encounter": "Telehealth Services"] ) ValidEncounter
    where ValidEncounter.status = 'finished' // <- HERE
      and ( ValidEncounter.reasonCode in "Major Depressive Disorder Active"
          or exists ( ( CQMCommon."EncounterDiagnosis" ( ValidEncounter ) ) EncounterDiagnosis
              where EncounterDiagnosis.code in "Major Depressive Disorder Active"
          )
      )
      and ValidEncounter.period during "Measurement Period"
```
This generates C# code as follows:
```csharp
Code<Encounter.EncounterStatus> z_ = ValidEncounter?.StatusElement;
Encounter.EncounterStatus? aa_ = z_?.Value;
Code<Encounter.EncounterStatus> ab_ = context.Operators.Convert<Code<Encounter.EncounterStatus>>(aa_);
bool? ac_ = context.Operators.Equal(ab_, "finished");
```

Unfortunately, the result is always `false` at runtime. There is no current way to compare between different types.

**To Reproduce**
A unit test was created for this scenario: `FhirCodeEqualityTests.Equal_FhirCode_vs_String` and `Equal_String_vs_FhirCode`

**Expected behavior**
The outcome must be `true`. The FHIR Code always encapsulates a FHIR enum, the the values of it are attributed with EnumLiteral like so:
```csharp
/// <summary>
/// Current state of the encounter.
/// (url: http://hl7.org/fhir/ValueSet/encounter-status)
/// (system: http://hl7.org/fhir/encounter-status)
/// </summary>
[...]

PR #348, opened 2024-06-01
2.0/347 fix comparison between code t and string

baseTwo:
ℹ️Fix for #347 

## Fix
* Added tests: `CqlContextOperatorTests`
  * Test equality between Code<T> and string and vice versa
  * Test conversion  between Code<T> and string and vice versa
* Added a decorator base type for `ICqlComparer`
* Updated `CqlComparers` to prioritize different types such that a type from the System namespace (such as `string`) is always on the right. This will allow us to get to the comparer on `Code<T>` first
* `CqlComparerExtensions.AddFhirComparers` will register a decorator type `CodeStringComparer` to handle `Code<T>` It will check if the right type is a `string` and handle it itself, otherwise it falls back to the `CodeComparer<>` type

## Other
* Remove following PackagerCLI args: `-d` and `-f`, since they were not implemented in 2.0
* Added the following argument `--log-debug` to force debug level logging
* Close/Flush the log after application exit
* Update `Cql.Targets.xml` to build up the PackagerCLIArgs property from other supplied properties.

### DIFF
--- a/Cql/CodeGeneration.NET/CSharpCodeWriterOptions.cs
+++ b/Cql/CodeGeneration.NET/CSharpCodeWriterOptions.cs
@@ -22,12 +22,13 @@ public class CSharpCodeWriterOptions
     /// </summary>
     internal const string ConfigSection = "CSharpCodeWriter";
 
+    internal const string ArgNameOutDirectory = "--cs";
+
     /// <summary>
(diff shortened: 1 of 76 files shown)


richfirely (approved):
Thanks for this quick work! CMS measures should be closer now with this fix in our test runner.

PR #352, opened 2024-06-06
480 Fix equivalence between CQL concepts

### LINKED ISSUE #480: Concept equivalence fix + unit test
According to the CQL spec: https://cql.hl7.org/02-authorsguide.html#comparison-operators

Quote: For Concept values, equivalence means the values have at least one equivalent code.

I created a test measure that used the exact examples in this unit test and it failed. I also tested this with java and it works.

The part that is unclear is if code should be equivalent to concept when the code matches at least one code in the concept.

richfirely:
Work for #480 

According to the CQL spec:

[https://cql.hl7.org/02-authorsguide.html#comparison-operators](https://cql.hl7.org/02-authorsguide.html#comparison-operators)

Quote: For Concept values, equivalence means the values have at least one equivalent code.

I created a test measure that used the exact examples in this unit test and it failed. I also tested this with java and it works.

The part that is unclear is if code should be equivalent to concept when the code matches at least one code in the concept. I couldn't find that in the spec. @EvanMachusak  do you know?


EvanMachusakNCQA:
Yes, equivalence is a commutative operator.  (x ~ y) ~ (y ~ x) for all x and y.

baseTwo:
@baseTwo note to self - see if this is still relevant, and test the unit test if it passes. If so, approve

baseTwo:
The logic has been fixed. I also added one additional unit test where at least one CqlCode in both sets is null. This fix does not affect the cms integration tests pass rate

baseTwo:
!!! Do not review yet! Checking if more integration tests are passing as a result of this fix !!!

baseTwo (changes requested):
I will add a fix for the unit tests to work

baseTwo (approved):

Issue #206, opened 2024-02-26
2.0/Cleanup - Remove Graph library to load FHIR Resources


baseTwo:
A follow up on issue #202 . 

Eliminate the Graph project entirely. The only thing we really need is topological ordering. We could probably do that better with a simple adjacency matrix or something. There could be some lean NuGet packages out there that might give us what we need.

Some work was done already for ELM libraries by moving functionality into a `LibrarySet` (see PR #238 ). FHIR resources still remain.

baseTwo:
Here is some code that can traverse a graph given a starting node and a function to get to the next items. It is assumed there are no circular dependencies. It could be improved by making DepthFirst non-recursive.

```csharp
public readonly record struct TraversalCallbacks<T>(
	Func<(T? Previous, T Current, T Next), bool>? AllowEnter = null,
	Action<(T? Previous, T Current)>? Enter = null,
	Action<(T? Previous, T Current)>? Exit = null)
{
	public static TraversalCallbacks<T> CreatePreventRevisits()
	{
		HashSet<T> visited = new();
		return new(AllowEnter: t => {
			var allow = visited.Add(t.Next);
			return allow;
		});
	}
}

public static class Traversal
{
	public static IEnumerable<T> DepthFirst<T>(
		T current,
		Func<T, IEnumerable<T>> getNextItems,
		TraversalCallbacks<T> callbacks = default,
		T? previous = default)
	{
		callbacks.Enter?.Invoke((previous, current));
		yield return current;

		foreach (var next in getNextItems(current))
		{
			if (callbacks.AllowEnter?.Invoke((previous, current, next)) ?? true)
			{
				foreach (var grandChild in DepthFirst(next, getNextItems, callbacks, current))
					yield return grandChild;
			}
		}		
		callbacks.Exit?.Invoke((previous, current));
	}

	public static IEnumerable<T> TopologicalSort<T>(
		T current,
		Func<T, IEnumerable<T>> getNextItems,
		TraversalCallbacks<T> callbacks = default,
		T? previous = default)
	{
		// Get the whole graph first (assuming everything is reachable from the current node)
		HashSet<T> unvisited = new();
		DepthFirst(current, getNextItems)
			.ForEach(n => unvisited.Add(n));

		HashSet<T> visited = new();
		Queue<T> results = new();
		TraversalCallbacks<T> callbacks2 =
			new(
				Enter: t =>
				{
					visited.Add(t.Current);
[...]

PR #355, opened 2024-06-12
2.0/206 remove graph library

baseTwo:
ℹ️Work for itssue #206 

* Removed Graph library from solution files, since there are no more usages of DirectedGraph
* Removed unused usings
* Updated solution filter Cql.slnf (added MultiPackagerCLI)


alexzautke (dismissed):

alexzautke:
Changes look good, build is failing though

alexzautke (approved):

Issue #290, opened 2024-04-12
Cleanup


baseTwo:
TODO's. 
(Some items were done and have PR's assigned to them, but after some major merges, they need to be done again)

- [ ] Merge `LibraryExpressionBuillder` with `ExpressionBuilder`
- [ ] Pick more meaningful names for `LibrarySet`-, `Library`- and `ExpressionBuilder`. The term `Expression` is very overloaded, should perhaps refer to Definitions instead, since that is what is being build See [PR comment](https://github.com/FirelyTeam/firely-cql-sdk/pull/306#discussion_r1582849334)
- [ ] Match namespace names to their folders
- [ ] Match class names to their files
- [ ] Correct usage of access modifiers. (private, protected, public, etc)
- [ ] Seal classes that are not going to be subclassed
- [ ] Small code cleanup comment in PR #418
- [x] Remove unused HL7 CQL files in CoreTests
- [x] Split *ExpressionBuilder/Context classes into separate files (PR #345)
- [ ] Remove all trailing whitespace (PR #345, #356)
- [ ] Refactor code to use collection initializer e.g. (PR #301)
  * `List<int> list = [];`, `int[] arr = [1];`
  * `[]` over `Types.Empty` or `Array.Empty<T>()` etc.
  * `new string([..charList])` over `new string(charList.ToArray())`
  * `[..arr1, ..arr2]` over `arr1.Concat(arr2).ToArray()`

baseTwo:
Since two big branches `cql-to-elm` and `scopes` have been merged into into `develop-2.0`, the sub tasks are re-opened

baseTwo:
Closing as obsolete: the code this checklist targets no longer exists.

The Expression-based pipeline was deleted in f7a3fa18 (`feat!: delete the Expression-based pipeline - the typed IR is the pipeline (phase 6)`) and the surviving builders were renamed the next day in 10481e88 (`refactor: rename ExpressionBuilder pipeline types to CodeBuilder`). `ExpressionBuilder`, `LibraryExpressionBuilder` and `ExpressionBuilderContext` return no hits anywhere in the repo today.

Item by item:

- **Merge `LibraryExpressionBuilder` with `ExpressionBuilder`** - not actionable; neither type exists. The modern equivalents are `CodeBuilder` / `LibraryCodeBuilder` / `LibrarySetCodeBuilder`, each with its own context type by design.
[...]

PR #356, opened 2024-06-12
2.0/290 Remove trailing whitespace, dead code. Add tests to check it.

baseTwo:
ℹ️Work for #290 

* Add a test to check that there are no trailing  whitespace in the code files.
* Fixed the files where that was case
* Removed files in dead folder

Failing test looks like this 
```
The following files have lines with trailing whitespace:
\Demo\CLI\LibraryRunner.cs
   6:        private readonly StreamWriter? _writer;     <eol>
   6:        <eol>
   6:                    <eol>
   6:        #region Writing Output     <eol>
   6:    }   <eol>
\Cql\Cql.Compiler\Expressions\NullConditionalMemberExpression.cs
   6:            // Nothing	<eol>
\Demo\CLI\Helpers\ResourceHelper.cs
   6:        };     <eol>
   6:            parametersConverted.Add(parameter.Name, converted!);         <eol>
```


baseTwo:
When running tests on the build pipeline, the tests are isolated from the source code, so it is not possible to get access to the solution dir from the test runner agent. The PR is therefore closed

PR #363, opened 2024-06-14
2.0/Fix - demo unit test library was still using DirectedGraph.

baseTwo:
ℹ️Work for #206 

The demo unit test library was still using DirectedGraph. Replaced LibraryExtensions with the one used in CLI via a file link

(Demo test projects are not run during the build pipeline)


richfirely (approved):
I tested this with https://github.com/FirelyTeam/firely-cql-sdk/pull/366 to be sure we have a good full compile.

Issue #365, opened 2024-06-18
2.0 Using Tuples is added to C# even when there are no tuples, causing compiler issue


richfirely:
**Describe the bug**
"Using Tuples" is added to the C# even when there are no tuples. The C# will not compile as a result.

I found this when starting the new empty project for DevDays and had to add a fake Tuples CQL file to work around.

**To Reproduce**
1. In the Measures.Authoring project, remove all the CQL files except for FHIRHelpers and ParametersExample.
2. Clean and build the project
3. Notice Packager CLI compile exceptions in the build.log

**Expected behavior**
If current files does not contain Tuples then the "using Tuples" should not be added. It appears to be added generically at line 71, BuildUsings within CodeGeneration.Net file CSharpLibrarySetToStreamsWriter.CS

The using section could be more accurate in general. When you highlight over certain usings the C# analyzer shows a few that are unnecessary. For example, FHIR Helpers does not need Hl7.Cql.ValueSets

For now maybe handle tuples only and consider a refactor of the logic in a future ticket to maybe do all the using more intelligently.

**Version used:**
2.0 Dev branch

baseTwo:
https://github.com/FirelyTeam/firely-cql-sdk/pull/366

PR #366, opened 2024-06-19
2.0/Fix emiting 'using Tuples' - only include when there are tuples

baseTwo:
ℹ️Fix for #365


richfirely (approved):
Looks good. Thanks for the quick turnaround

PR #381, opened 2024-07-02
Develop 2.0 merge from cql to elm (and develop)

baseTwo:
ℹ️Work for #371 and #386

### Background
Prior to this PR, develop was merged into `develop-2.0`, and `develop` was merged into `cql-to-elm`.

This PR merges `cql-to-elm` and 'scopes' into `develop-2.0`. These branches deviated significantly over the past 4 or more months and merging them all back into one, is quite a challenge.

### Remarks
* Updated the solution filter file `Cql.slnf`, to include the three new projects related to cql-to-elm
* Removed the folder `Cql.Firely.Test`, since it only had one unit test without any assertions
* The `CqlOperatorBinder` method binding logic [had to be expanded](https://github.com/FirelyTeam/firely-cql-sdk/pull/381/commits/09906a5fb086af3ab986ea6ff227b1b4af677d5a#diff-52d505647a01e7b310c340b9bdd73389c22fc27a0f42989b35b70bc265ba6bf8). Previously only the inner generic argument was matched to the generic parameter, but now the argument itself is also tried. Meaning, previously when the input argument is CqlInterval<CqlDate> and the parameter is T, then only CqlDate was tried on T. Now, CqlInterval<CqlDate> is also tried.
* Added [missing closing](https://github.com/FirelyTeam/firely-cql-sdk/pull/381/commits/09906a5fb086af3ab986ea6ff227b1b4af677d5a#diff-0b853ddebe409674734de7e63d444b4686c942f3f8fe122ed1862c5018281256) on CqlInterval operator
* `ElmAsExpression` must also take an [argument for strictness](https://github.com/FirelyTeam/firely-cql-sdk/pull/381/commits/1c8917e20c4d73f5cb045ebfd7b887ba6ca0ad2f#diff-fd8ee1480c41b6ef51fbb7284bbc7389b18e8a4728f74d320fa5a81aed6deb86)
* `Base` test in CqlToElmTests must use `LibraryExpressionBuilder` [instead of ExpressionBuilder](https://github.com/FirelyTeam/firely-cql-sdk/pull/381/commits/def59cd061d1e7e30c3cae4c5597048b1ce6aaa1#diff-0d2eaef45771775544eead1f9ffc247c0b8c990f0276c03e4809d618a4dbd430)
* The `Lambda` method was removed on `ExpressionBuilder`, but the CqlToElmTests still use it, so had to put an [extension method](https://github.com/FirelyTeam/firely-cql-sdk/pull/381/commits/fdd0b9390e53bbb1d93fbec240c13c8d43a5adca#diff-d5ca8a4b89777db901a2150431f5e85431533dd6e3f42c0bd57e48718358ffe7) for it there. Also, the `ExpressionBuilderContext.TranslateArg` method had to be [opened from private to internal](https://github.com/FirelyTeam/firely-cql-sdk/pull/381/commits/fdd0b9390e53bbb1d93fbec240c13c8d43a5adca#diff-87d951321fb28d7802b9c267bda6f2ada35d0429d792d916fca8c420e7bb8a38), to make this work
* The generated C# has many places where [null is cast to a type](https://github.com/FirelyTeam/firely-cql-sdk/pull/381/commits/ebac8a104b192dcb4dab0667109b47b72cb9d4bf). Must be fixed in issue #387 387.
* Always convert to result type when translating elements in `ExpressionBuilderContext.TranslateElement`. This is especially important when two arguments are passed to Equals, since it expects the argument types to be the same
* Fix ExpressionBuilderContext.ConceptRef by returning the type of the actual conceptRef, and not hard-coded CqlCode[] (#386)
* Small fix in CqlOperatorsBinder. 
* Regenerate all C#

### CMS Integration Tests
Running the integration tests ([a separate repo from this](https://github.com/FirelyTeam/Firely.Cql.Sdk.Integration.Runner)) returns the following results:
* Tests passing before merge: 902 / 1822 (47%)
* Tests passing after merge: 1247 / 1822 (68%)

### Questions
* Why is the length of a null string, null, but the length of a null list is zero? See tests `LengthNullString` and `LengthNullList`. The logic for `Length` can be found in `CqlOperators.Length(string?)` and `CqlOperators.Length<T>(IEnumerable<T>?)`. I added the LengthNullString test in code and in the xml tests as well
* There are 2 `XmlTests`, one in CoreTests, the other in CqlToElmTests. The one in CoreTests is empty? My guess is the CqlToElmTests.XmlTests is replacing CoreTests.XmlTests and thus the latter must be removed?
image


baseTwo:
Since this is a big merge between the three branches, @ewoutkramer suggested @EvanMachusak also reviewing

Cql/Cql.Primitives/CqlInterval.cs
```diff
@@ -33,9 +33,6 @@ public class CqlInterval<T>
         /// <exception cref="ArgumentNullException">If both <paramref name="low"/> and <paramref name="high"/> are null.</exception>
```
ewoutkramer: This is no longer true then.....

Cql/Elm/ElmTreeWalker.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Hl7.Cql.Elm
+{
+    /// <summary>
+    /// A walker that does a depth-first traversal of a tree and calls a visitor function for each node.
+    /// </summary>
+    public class ElmTreeWalker
```
ewoutkramer: Not sure we should make this public.  It's useful for others, that's true. What do we think?

ewoutkramer (dismissed):

EvanMachusakNCQA:
I removed the CoreTests version because the XmlTests in CqlToElm supplanted it.

Cql/Cql.Primitives/CqlInterval.cs
baseTwo: Fixed

Cql/Elm/ElmTreeWalker.cs
baseTwo: Let's keep things internal, until we review the public api later?

ewoutkramer (dismissed):

ewoutkramer:
All the warnings about new public API have to be resolved, otherwise it will not build....

baseTwo:
The PR will be changed back to draft. The scopes branch will be merged as a separate PR, and the work before that will move over to another PR

ewoutkramer:
Since all work has moved to the scopes branch, and the second version of this PR has been merged, I am closing this one.

PR #382, opened 2024-07-02
Improve json serialization support (on develop-1.0)

ewoutkramer:
I managed to get it to work with System.Text.Json after all, which did allow me to clean up the code quite a bit. 

* The deserializer now has a pre-processing stage that makes sure that the type discriminators are always first & that old-style "type" properties are upgraded to "resultTypeSpecifier". 
* Narrative is now fully supported in the format that is currently in the ELM produced by the Java stack
* Added write support to the custom converters.


ewoutkramer:
Replaced by #411.

PR #396, opened 2024-07-11
Add support for DataRequirements to the packager

### LINKED ISSUE #395: Add minimal DataRequirement support
Currently, our SDK packager produces Library resources without DataRequirements. This means that servers running such a library must be able to supply all data for the context (most often, a Patient) to the engine. This is wasteful, since Libraries will never use all data for a patient.

By generating basic DataRequirements specifying the resource type used in the `Retrieve` statements in the ELM, we can limit the set of data needed to fetch and pass on to the engine.

ewoutkramer:
Fixes #395.

Note that this really is a basic first version, just adding the names of the resource types as datarequirements to the Library when packaging.


ewoutkramer:
Not that basic anymore, as I ported (part of) the public Java analyzer to C#, so we at least have codes to limit the retrieves as well.

ewoutkramer:
This is an example of a DataRequirement that we now add to our Library resources:

```json
  {
      "type": "Encounter",
      "profile": [
        "http://hl7.org/fhir/StructureDefinition/Encounter"
      ],
      "mustSupport": [
        "type"
      ],
      "codeFilter": [
        {
          "path": "type",
          "valueSet": "https://www.ncqa.org/fhir/valueset/2.16.840.1.113883.3.464.1004.1761"
        }
      ]
    }
```

PR #405, opened 2024-07-15
Add support for DataRequirements to the packager

### LINKED ISSUE #395: Add minimal DataRequirement support

ewoutkramer:
Fixes https://github.com/FirelyTeam/firely-cql-sdk/issues/395.

Note that this really is a basic first version, just adding the names of the resource types as datarequirements to the Library when packaging.


ewoutkramer:
Not that basic anymore, as I ported (part of) the public Java analyzer to C#, so we at least have codes to limit the retrieves as well.

ewoutkramer:
This is an example of a DataRequirement that we now add to our Library resources:

```json
  {
      "type": "Encounter",
      "profile": [
        "http://hl7.org/fhir/StructureDefinition/Encounter"
      ],
      "mustSupport": [
        "type"
      ],
      "codeFilter": [
        {
          "path": "type",
          "valueSet": "https://www.ncqa.org/fhir/valueset/2.16.840.1.113883.3.464.1004.1761"
        }
      ]
    }
```

Cql/Cql.Compiler/LibrarySet.cs
```diff
@@ -213,11 +217,14 @@ private void RecalculateStateIfNecessary()
     /// <param name="lib">The name of the library to load.</param>
     /// <param name="version">The version of the library to load.</param>
     /// <returns>A collection of loaded libraries, including the specified library and its dependencies.</returns>
+    /// <remarks>Supply the name of the library in <paramref name="lib"/>, not the name of the file.</remarks>
     public IReadOnlyCollection<Library> LoadLibraryAndDependencies(
         DirectoryInfo elmDirectory,
         string lib,
         string version = "")
     {
+        if(lib.EndsWith(".cql")) lib = lib[..^4];
```
baseTwo: This should be case-insensitive comparison

Cql/Cql.Compiler/LibrarySetDefinitionCache.cs
```diff
@@ -0,0 +1,108 @@
+using System;
```
baseTwo, line 9: Missing file header

Cql/Elm/IDefinitionElement.cs
```diff
@@ -296,9 +308,10 @@ public partial class UsingDef : IDefinitionElement
     partial class IdentifierRef : IDefinitionElement
     {
         [JsonIgnore]
-        public AccessModifier Access => AccessModifier.Private;
+        AccessModifier IDefinitionElement.Access => AccessModifier.Private;
```
baseTwo, line 311: Is JsonIgnore still necessary on explicit interface properties?

baseTwo (changes requested):

Cql/CoreTests/Elm/TestDataRequirementsAnalyzer.cs
```diff
@@ -0,0 +1,41 @@
+using System.IO;
```
baseTwo, line 1: NEeds file header

Cql/CoreTests/Elm/TestElmResolution.cs
```diff
@@ -0,0 +1,65 @@
+using System.IO;
```
baseTwo, line 1: NEeds file header

Cql/Cql.Compiler/Errors.cs
```diff
@@ -17,16 +17,38 @@ namespace Hl7.Cql.Compiler;
...
+    public string GetMessage() => $"Library has an {Def.GetType()} element without a name. Library Identifier: '{Library}', Name: '{Def.Name}'.";
+}
+
 internal readonly record struct LibraryIdentifierMustBeUniqueError(Library Library) : ILibraryError
 {
     public string GetMessage() => $"Library did not have a unique name and version in the set. Duplicate Identifier: '{Library}'";
 }
 
-internal readonly record struct LibraryAliasUnresolvedError(Library Library) : ILibraryError
+internal readonly record struct LibraryDoesNotAppearAsADependencyError(Library Library) : ILibraryError
```
baseTwo: The message returned doesn't match with the name of this method

Cql/Cql.Compiler/LibrarySet.cs
```diff
@@ -210,16 +214,20 @@ private void RecalculateStateIfNecessary()
...
     /// <param name="elmDirectory">The directory containing the ELM files.</param>
-    /// <param name="lib">The name of the library to load.</param>
+    /// <param name="libraryPath">The name of the library to load.</param>
     /// <param name="version">The version of the library to load.</param>
     /// <returns>A collection of loaded libraries, including the specified library and its dependencies.</returns>
+    /// <remarks>Supply the name of the library in <paramref name="libraryPath"/>, not the name of the file.</remarks>
     public IReadOnlyCollection<Library> LoadLibraryAndDependencies(
         DirectoryInfo elmDirectory,
-        string lib,
+        string libraryPath,
```
baseTwo: libraryName could be a better name for this parameter. It is not the whole path of it that is passed in here

Cql/Cql.Compiler/LibrarySetDefinitionCache.cs
```diff
@@ -0,0 +1,116 @@
...
+        private LibraryCache()
+        {
+            // Nothing
+        }
+
+        private readonly Dictionary<Type, object> _cache = new();
+
+        private IDictionary<string, T> GetIndexForType<T>()
+        {
+            if (_cache.TryGetValue(typeof(T), out var index)) return (IDictionary<string, T>)index;
```
baseTwo: Put return on a separate line, then add another blank line after it. This improves the readability of the method exit point

Cql/Cql.Compiler/LibrarySetDefinitionCache.cs
```diff
@@ -0,0 +1,116 @@
...
+            _cache.Add(typeof(T), newIndex);
+            return newIndex;
+        }
+
+        private static void Index<T>(
+            IDictionary<string, T> result,
+            Library library,
+            T[]? definitions) where T : IDefinitionElement
+        {
+            if (definitions is not { Length: > 0 }) return;
```
baseTwo: Return on its own line

Cql/Cql.Packaging/DataRequirementsAnalyzer.cs
```diff
@@ -0,0 +1,214 @@
...
+    {
+        // Function toDataRequirement in the above Java file
+        if (node is not Elm.Retrieve retrieve) return false;
+
+        var dr = new DataRequirement();
+
+        // Get the resource type the requirement is for.
+        var (_, resourceType) = retrieve.dataType;
+        var knownType = Hl7.Fhir.Model.ModelInfo.FhirTypeNameToFhirType(resourceType);
+        if (knownType is null) return true; // Not a known FHIR type, so we can't do anything with it.
```
baseTwo: REturn on own line, followed by blank line

baseTwo (changes requested):

Cql/Cql.Compiler/LibrarySetDefinitionCache.cs
ewoutkramer: This is going to be hard, I have never done that in 30 years....

Cql/CoreTests/Elm/TestDataRequirementsAnalyzer.cs
ewoutkramer: Sigh...one thing I miss about VS.....splitting a class off into another file did copy the header in VS, but not in Rider. Oh well.....

ewoutkramer: I noticed *none* (well, except one or two) of our test files have headers, so I won't fix that here. If we want our tests with headers, we'll need to make a PR to fix it for all of them, it does not make sense to just fix up a few.

baseTwo: I guess it's just easier to remove the header from the unit tests, I don't think they would be useful to anyone as-is. [I'll create a ticket for that ](https://github.com/FirelyTeam/firely-cql-sdk/issues/436)

baseTwo (approved):

Issue #397, opened 2024-07-15
Solve to fix ExpressionBuilder when definitions are not in-order


baseTwo:
Predecessor issue #422 

* `ExpressionBuilderContext.InvokeDefinedFunctionThroughRuntimeContext` when definitions are out of order
* Fix unit tests - Find the `[Ignore("Will fix in https://github.com/FirelyTeam/firely-cql-sdk/issues/397")]` attributes from 4 unit tests and fix them

Check additional todo's  in Slack discussion : https://firelyteam.slack.com/archives/C05GW1J1HFV/p1721036253203929

ewoutkramer:
We will implement this by visiting the ELM before we start processing it. Using the visitor + the recent LibrarySet resolver functions we can easily add `resultTypeSpecifiers` to all FunctionRefs.

baseTwo:
Demo projects didn't build, work rolled back in PR #437

baseTwo:
Reopened ticket as it is not merged yet #468

baseTwo:
Added in the description that 4 unit tests must be un-ignored to truly test if this feature is working.

PR #408, opened 2024-07-15
Merge scopes branch into develop-2.0

baseTwo:
ℹ️Work for #398 

* Merged the `scopes` branch into `develop-2.0` (which is now the new default branch as of 15 July 2024)
* The significant changes are in the following folders:
  * `\Cql\Cql.CqlToElm`
  * `\Cql\CqlToElmTests`
  * `\Cql\Elm`
* Some Microsoft Nuget package versions went back to a previous major version, so they had to be changed back to the latest
* The ExpressionBuilderContext.InvokeDefinedFunctionThroughRuntimeContext() had to be reverted to the original 2.0 behavior. Some new unit tests are failing because of that and is ignored for now. An issue was created to fix this in #397


baseTwo:
Replaced by PR #419

Issue #440, opened 2024-08-01
Better integration between `CQL SDK` and  `Integration Tests`


baseTwo:
The [Firely CQL SDK](https://github.com/FirelyTeam/firely-cql-sdk) and the [Firely CQL Integration Test Runner](https://github.com/FirelyTeam/Firely.Cql.Sdk.Integration.Runner) are two separate repos.

For the most part this is fine, since the Integration Test Runner includes the CQL SDK via NuGet packages.

However, to properly debug failing tests, it is desirable to step through the generated C# demo libraries. Currently these demo projects are included in the CQL SDK repo. They may move to a separate repo in the future.

A second requirement is that the integration tests must be run as part of every PR in the CQL SDK. (see link to successor issue below)

A folder structure is required so that the Integration Test Runner may include project references to the CQL SDK and the Demo projects which will have consistent relative paths no matter who checks out the repos on their local environment.

Two ways exists to do this, Git submodules or Git subtrees. See my notes on submodules/subtrees: https://firely.atlassian.net/wiki/x/EAD5Yg. Also should the Integration Test Runner be a submodule of the CQL SDK, or the other way around?

Since the CQL SDK is the main repo, it was chose that the Integration Test Runner would be a submodule into the CQL SDK repo. A submodule is not automatically retrieved when the main repo is cloned, and for most developers on the CQL SDK, they would not need the Integration Test Runner. It is however beneficial for the build pipeline of the CQL SDK to have submodule access to the Integration Test Runner, so that when the Demo projects are rebuild, the Integration Test can run on each PR in the CQL SDK.

An alterative which could also be considered if this workflow doesn't work well, is to have the CQL SDK as a submodule/subtree in the Integration Test Runner. The Integration Test Runner would have to be NuGet packaged and then included and run on a PR for the CQL SDK. 

Successor Issue : #428

baseTwo:
Added wiki page https://firely.atlassian.net/wiki/x/EAD5Yg

PR #447, opened 2024-08-05
Add submodule to git repo Firely.Cql.Sdk.Integration.Runner

baseTwo:
ℹ️Work for #440 

* Create a folder containing all submodules, named `submodules`
* Added new submodule `Firely.Cql.Sdk.Integration.Runner` (https://github.com/FirelyTeam/Firely.Cql.Sdk.Integration.Runner)
* Solution file renamed to `Cql-Sdk-All.sln` and updated to include the projects of the submodule `Firely.Cql.Sdk.Integration.Runner`
* Solution filter files to select parts from the main solution file above `Cql-Sdk.slnf` and `Cql-Sdk-and-Demo.slnf`
* Updated build pipeline to
  * Enabled submodules
  * Enabled git long path support
  * Note: Cannot enable a rebuild of the demo projects yet, since that requires the Java tooling and dependencies in the pipeline
* Updated getting started documentation


richfirely:
My only comment is that I am weary of the plus sign in the naming convention and also to avoid renaming the solution file, and therefore the pipeline, all the time maybe we should just make it generic and make solution filters for each module combination. Cql-Sdk.sln ?? If neither of these concern you right now, we can merge.

richfirely (dismissed):

cheng13231 (approved):

PR #446, opened 2024-08-05
Submodules for CqlToElmTests and Firely.Cql.Sdk.Integration.Runner

baseTwo:
ℹ️Work for #448

* Create a folder containing all submodules, named `submodules`
* Move existing submodule from `Cql/CqlToElmTests` → `submodules/CqlToElmTests` (https://github.com/cqframework/cql-tests)
* Update build pipeline to include tests under `submodules` folder

PR #469, opened 2024-08-19
Fix #466

### LINKED ISSUE #466: Update CQL Conversion Mapping (Id / FhirString)
Bug created from this [discussion on Slack](https://firelyteam.slack.com/archives/C05GW1J1HFV/p1723811123444779).

There seems to be a mismatch between the element types in the [spec](https://hl7.org/fhir/R4B/datatypes.html#id) (see diagram) and the fhir model in the sdk.

Image

According to the spec, and Id inherits from FhirString, however this is not the case. They both inherit from the same type:
FhirString : PrimitiveType : DataType : Element : ...
Id : PrimitiveType : DataType : Element : ...

**Note: this applies to all the primitive types on the diagram, they do not go through the intermediate types as on the diagram above! Actual type hierarchy below**
Image

The generated C# assumes Id can be safely cast to FhirString. (The example below equates to properties, on of type Id and the other of type FhirString). This is not compliable.
Image

After[ discussing with the firely-net-sdk team](https://firelyteam.slack.com/archives/C011XD1TKL0/p1724053272836459), they are convinced it is the spec that is incorrect/unclear.

So, the conversion mappings in the ELM-to-C# engine would have to be updated, so that types are converted instead of casted.

EvanMachusak:
- Adds an out parameter to `ChangeType` to indicate the result of the conversion.  This prevents us from needing to have a `CanConvert` or similar function, which would result in duplicative work.  In all cases where `ChangeType` would result in `NoMatch`, an exception is thrown.
- In the `As` operator, if there is a custom conversion defined between two types, we invoke it.  If `ChangeType` identifies a `NoMatch` condition, we return `default`.  Otherwise, we produce a standard C# `as` operator.


baseTwo:
@EvanMachusak I continued to a new branch and PR #473

I picked up that:
* The conversion code in Negate and NegateLiteral was duplicate
* The Negate method is called in TranslateElement, which already does result type conversion for each element type, thereby, the conversion logic for Negate would be run twice

I created a new branch to:
* Consolidate the conversion code between Negate, NegateLiteral and TranslateElement into a local function in TranslateElement
* To see the outcome of this, the demo projects was rebuild so that the effect of the changes on the generated code can be visible

If you happy with the changes there, this PR can be closed and that one merged instead.

baseTwo (commented):
@EvanMachusak I continued to a new branch and PR #473

I picked up that:
* The conversion code in Negate and NegateLiteral was duplicate
* The Negate method is called in TranslateElement, which already does result type conversion for each element type, thereby, the conversion logic for Negate would be run twice

I created a new branch to:
* Consolidate the conversion code between Negate, NegateLiteral and TranslateElement into a local function in TranslateElement.ConvertToResultType(). <br>*Note:* There was a hard fail if conversion did not succeed in the Negate methods, but when this is done in TranslateElement, 15 unit tests fail (e.g. no conversion from Interval<CqlDateType> to Interval<any>). We should probably look into this in another issue.
* The  last bit in the `ChangeType` logic threw an error when binding to Convert isn't possible. Changed it to TryConvert.
* To see the outcome of this, the demo projects was rebuild so that the effect of the changes on the generated code can be visible
* The correct issue number is #466, not #446  (updated in the new PR description)

If you happy with the changes there, this PR can be closed and that one merged instead.

PR #473, opened 2024-08-20
Conversions

### LINKED ISSUE #466: Update CQL Conversion Mapping (Id / FhirString)

baseTwo:
Fix for issue #466 
Continued from PR #469 

- Adds an out parameter to `ChangeType` to indicate the result of the conversion.  This prevents us from needing to have a `CanConvert` or similar function, which would result in duplicative work.  In all cases where `ChangeType` would result in `NoMatch`, an exception is thrown.
- In the `As` operator, if there is a custom conversion defined between two types, we invoke it.  If `ChangeType` identifies a `NoMatch` condition, we return `default`.  Otherwise, we produce a standard C# `as` operator.
- `ChangeType` throws an error when binding to Convert fails. Instead, use TryConvert to the binder.


EvanMachusak:
Tested by:
- `CqlToElm/AsTest/FhirId_As_FhirString`
-  `CqlToElm/AsTest/Choice_As`

EvanMachusak (approved):

PR #500, opened 2024-08-30
Fix 499 - generating parameter names for external functions

EvanMachusak:
When generating an automatic parameter name for a generic type, we strip the `@` character if it was added normalizing the generic type argument name (e.g., because it's a C# keyword like `decimal`).  Example of a bad name was `nullable_@decimal` which is now produced as `nullable_decimal`.

Also made the `definitions` parameter of `AssemblyCompiler.Compile` no longer optional.  When not specified, the function does nothing.

Tested by CqlToElmTests/FunctionDefinitionTest/External_Function

### DIFF
--- a/Cql/CodeGeneration.NET/AssemblyCompiler.cs
+++ b/Cql/CodeGeneration.NET/AssemblyCompiler.cs
@@ -71,9 +71,10 @@ namespace Hl7.Cql.CodeGeneration.NET
 
         public IReadOnlyDictionary<string, AssemblyData> Compile(
             LibrarySet librarySet,
-            DefinitionDictionary<LambdaExpression>? definitions = null)
+            DefinitionDictionary<LambdaExpression> definitions)
         {
-            definitions ??= new();
+            if (definitions is null)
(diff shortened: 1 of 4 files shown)


Cql/CodeGeneration.NET/AssemblyCompiler.cs
```diff
@@ -71,9 +71,10 @@ public AssemblyCompiler(
 
         public IReadOnlyDictionary<string, AssemblyData> Compile(
             LibrarySet librarySet,
-            DefinitionDictionary<LambdaExpression>? definitions = null)
+            DefinitionDictionary<LambdaExpression> definitions)
```
ewoutkramer, line 74: This will require the user to create an empty dictionary themselves, why is that necessary?

ewoutkramer (approved):

Cql/CodeGeneration.NET/AssemblyCompiler.cs
baseTwo: Resolved, since the usages to this method create them anyway

PR #501, opened 2024-09-02
406 standarize on dependency injection strategy

### LINKED ISSUE #406: Standarize on dependency injection strategy
Currently there are two dependency injection strategies: The .net way with the host builder, and the factories way currently in the 2.0 repo (types derived from `CqlFactoryBase`).

Only one should be used.

CQL to ELM testing uses ServiceProvider and built in the BaseTest class.

baseTwo:
Work for #406 

### Removed the old factory types : 
* `CqlFactoryBase` (in Hl7.Cql.Abstractions)
* `CqlAbstractionsFactory` (in Hl7.Cql.Abstractions)
* `CqlCompilerFactory` (in Hl7.Cql.Compiler)
* `CqlPackagerFactory` (in Hl7.Cql.Packaging)
* `ProgramCqlPackagerFactory` (in Hl7.Cql.Packager)

### New types to simplify service construction

The following projects (see diagram below) contains services that can be added to a service collection, typically as part of a hosting environment. These projects all have a new sub-namespace called `Hosting`. 

```mermaid
classDiagram
    direction BT
    Cql_Compiler ..> Cql_Runtime
    Cql_CqlToElm ..> Cql_Runtime
    CodeGeneration_NET ..> Cql_Compiler
    Cql_Packager ..> CodeGeneration_NET
```
*<small>Diagram of project hierarchy containing CQL services</small>*

The `Cql.Runtime` project doesn't have any services of its own that is used for the elm to c# pipeline, so its `Hosting` namespace only contains utility extensions needed by the other projects, such as `ServiceCollectionExtensions` and `ServiceProviderExtensions`. 

For the other projects, their `Hosting` namespace have two types each, one to construct services into a `IServiceCollection`, and another to access services from `IServiceProvider` (which is primarily only used for unit testing).  A consistent naming convention is used for types between all these projects: `Cql*ServicesInitializer`  and `Cql*Services`.

`Cql*ServicesInitialzer` contains methods such as `AddCql*Services` and `GetCql*Services`.

## Runtime Usage
To add services to `IServiceCollection` at runtime, just call any of the following 
* `services.AddCqlPackagerServices()` - adds services for packager, code generation and compiler
* `services.AddCqlCodeGenerationServices()`  - adds services for code generation and compiler
* `services.AddCqlCompilerServices()` - adds services for compiler
* `services.AddCqlToElmServices()` - adds services for cql to elm

## Test Usage

The `CqlServicesInitializer` is a helper to create services for the compiler or packager services. 

```csharp
// Create a context to which services will be registered to for disposal
using DisposeContext disposeContext = new DisposeContext();

// Get all the services for CQL code generation
CqlCodeGenerationServices cqlCodeGenerationServices = CqlServicesInitializer.CreateCqlCodeGenerationServices(disposeContext.Token);

// Get the CSharpLibrarySetToStreamsWriter 
CSharpLibrarySetToStreamsWriter writer = cqlCodeGenerationServices.CSharpLibrarySetToStreamsWriter;

// Direct access to services from dependency libraries possible with GetCqlCompilerServices()
LibraryExpressionBuilder libraryExpressionBuilder = cqlCodeGenerationServices.GetCqlCompilerServices().LibraryExpressionBuilder;
```

## Cql-to-Elm

The same dependency injection pattern is used for cql to elm.


baseTwo:
Discussed with @ewoutkramer that we add the same pattern to initialize and group services for the cql to elm project the same way

Cql/CodeGeneration.NET/Hosting/CqlCodeGenerationServices.cs
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+using System;
+using Hl7.Cql.CodeGeneration.NET.PostProcessors;
+using Hl7.Cql.Compiler.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Hl7.Cql.CodeGeneration.NET.Hosting;
+
+internal readonly struct CqlCodeGenerationServices(IServiceProvider serviceProvider)
```
ewoutkramer, line 16: Ah, if only we had extension properties ;-)

Cql/CoreTests/ExpressionBuilderTests.cs
```diff
@@ -1,48 +1,48 @@
...
-        private static ILoggerFactory LoggerFactory { get; } =
-            Microsoft.Extensions.Logging.LoggerFactory
-                .Create(logging => logging.AddDebug());
-
-        private static readonly CqlCompilerFactory Factory = new(LoggerFactory);
-
         [TestMethod]
         public void AggregateQueries_1_0_0()
         {
+            using var disposeContext = new DisposeContext();
```
ewoutkramer, line 17: Do we really need to repeat this boilerplate for every test?

ewoutkramer (approved):

PR #510, opened 2024-09-05
509 Fix equivalence

### LINKED ISSUE #509: Fix Equivalence checking - should never return null
Currently our cql operators for equivalence sometimes return null. Unlike with equal, this should never be the case.

For equal:
The Equal operator returns true if the arguments are equal; false if the arguments are known unequal, and null otherwise. Equality semantics are defined to be value-based.
see: https://cql.hl7.org/04-logicalspecification.html#equal

For equivalence :
"The Equivalent operator returns true if the arguments are the same value, or if they are both null; and false otherwise.
**With the exception of null behavior and the semantics for specific types defined below, equivalence is the same as equality.**" 
see: https://cql.hl7.org/04-logicalspecification.html#equivalent

baseTwo:
Fix for #509

Unlike with equality, equivalence returns `true` instead of `null`, if both inputs are `null`. It also never returns `null`
This logic is placed in a common method `CqlComparers.EquivalentOnNullsOnly()` 
Updated equivalence methods, to first go through this logic, reducing repetitive code.

Also,
* Moved all equality and equivalence from `CqlOperators` into a file `CqlOperators.EqualityAndEquivalence.cs`
* `CqlComparerBase` allowed derived types to avoid this duplication by implementing `Equivalent` through the nullability checks, then passing on to the non-null abstract method `EquivalentImpl`
* Some types had incorrect nullability for `GetHashCode`. Since these are only compile-time hints, the changes won't change at runtime

### DIFF
--- a/Cql/Cql.Abstractions/ICqlComparer.cs
+++ b/Cql/Cql.Abstractions/ICqlComparer.cs
@@ -46,7 +46,7 @@ namespace Hl7.Cql.Abstractions
         /// </summary>
         /// <param name="x">The object whose hash code to compute.</param>
         /// <returns>The hash code for <paramref name="x"/>.</returns>
-        int GetHashCode(object x);
+        int GetHashCode(object? x);
     }
 
     internal abstract class CqlComparerDecorator(ICqlComparer inner) : ICqlComparer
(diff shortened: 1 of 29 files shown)


richfirely:
I tested this and it fixes the two unit tests in question. I'd prefer @EvanMachusak or @ewoutkramer review the code though.

ewoutkramer (approved):

PR #521, opened 2024-09-09
388 Make Retrieve accept a parameters argument

ewoutkramer:
Fixes #388.

Note that we have not yet extended BundleDataSource to actually filter based on the templateId. For this, I have added a separate issue, #520.

PR #522, opened 2024-09-09
regen c#

baseTwo:
When generating c# I noticed small differences in parameter naming for FHIRHelpers. This must have been missed from a previous PR.


ewoutkramer:
Yes, I noticed it too, it is also part of #521, but let's pull this separately.

ewoutkramer (approved):

PR #530, opened 2024-09-11
526 Remove IServiceProvider adapters. Rename Hosting namespace to DependencyInjection

### LINKED ISSUE #526: Remove IServiceProvider encapsulation and use GetRequiredService<T> only
* Remove encapsulating service retrieval into a struct
* Remove DisposeContext in unit tests and improve lifetime cycle
* Rename Hosting namespaces

baseTwo:
Work for #526 

* Renamed `*.*.Hosting` namespaces to `Microsoft.Extensions.DependencyInjection`. (Most API's on Github uses this convention, to enhance service initialization discoverability to developers)
* Replace the `IServiceProvider` adapters `{ProjectName}Services`, with extension methods `{ProjectName}ServiceProviderExtensions`. These are used in unit tests and provide an explicit way to access internal services. In production use, these services must be accessed via dependency injection.
* Renamed the `{ProjectName}Initializor` extension method type to `{ProjectName}ServiceCollectionExtensions`
* Removed `DisposeContext`, `DisposeContextToken` and `CqlServicesInitializer` which were used in unit testing, but now replaced with the changes above
* Moved `GetProperty` method from `ExpressionBuilder` to `ExpressionBuilderContext`, because that's where it was used. Moved the accompanying `Get_Property_Uses_TypeResolver` unit test into `ExpressionBuilderContextTests` as well
* Note some changes to adding options in cql to elm. Previously it went through registering options through a verbose way (create options, copy into in-memory config, then binding it again to a new options). If the use case is to create the options manually, like with testing, its much simpler to just configure it directly see `CqlToElmServiceCollectionExtensions.AddCqlToElmOptions()`. At some point indeed it needs to be able to load from `IConfiguration` as well


baseTwo:
@ewoutkramer what is your opinion on having dependency injection types all in the `Microsoft.Extensions.DependencyInjection` namespace? This makes it easier to discover extensions when users include our libraries.

ewoutkramer (approved):

PR #531, opened 2024-09-11
516 Prepend resource type (eg Library or Measure) to resource file extension

### LINKED ISSUE #516: Measure library creation overwriting the library resource file. Measure output incomplete.
**Describe the bug**
When compiling libraries with the packager, if the library is also a measure, it will overwrite the FHIR library resource file. Measures should be written to a different filename or folder. ResourcePackager.cs CreateMeasureResource. 

We should also produce the full and correct measure FHIR resource.

**To Reproduce**
Compile elm that is both a library and a measure. Notice only the measure is created.

**Version used:**
2.0 branches.

⚠️ Must be discussed at September WGM as this is not per spec.

baseTwo:
Work for #516 

Prepend resource type (eg Library or Measure) to resource file extension, since there could be overwrites on the same file when the resource is both

eg.
```
Library-LibraryName-Version.json
Measure-MeasureName-Version.json
```

A new date type, [ResourceFileName](https://github.com/FirelyTeam/firely-cql-sdk/pull/531/files#diff-02c7ed21d8fa1bbcee52cf023679147a19b0a0a8c6a99431516f3b129f4775ab) contains the logic for parsing and formatting:
* Two or three non-empty parts separated by a hyphen '-' : Type, Identifier and optional Version
* Because hyphen is used as separator, it cannot be used in a part. 
* "To avoid issues with characters between web ids and names, library names SHALL NOT have underscores." see [spec](https://build.fhir.org/ig/HL7/cql-ig/conformance.html#library-name-and-url:~:text=To%20avoid%20issues,the%20FHIR%20specification.) 
* Assumption: For simplicity, each part is not further checked for whitespace or other illegal characters see [spec](https://build.fhir.org/ig/HL7/Content-Management-Infrastructure-IG/using-cql.html)

Unit Tests added for ResourceFileName.


baseTwo (commented):
Think about other formats e.g. Having the resource type name as part of the file extension, eg. Reporting.Library.json.
Consider that this file name change effects resource loading of dependencies.

baseTwo:
Must be discussed at September WGM. Must have settings to enable this or not

richfirely:
@baseTwo Based on this doc:
https://build.fhir.org/ig/HL7/cql-ig/conformance.html
CQL library source files SHOULD be named CQLLibraryName-Version.cql

I think that we should go with the format when writing to file system
Library-LibraryName-Version.json
Measure-MeasureName-Version.json

This way it mimics more closely the source cql recommendation.

I think we should merge this change with the modification stated above, so that the measure building isn't overwriting the library files. We agreed at WGM that building measures from CQL annotations was a good value add to the SDK. For now, just fixing the "bug" of overwriting is more important.

The settings part and output still needs design. I have some code started for this which I will submit in a future PR.

baseTwo:
Once this is completed, built a new release and uncommented the pipeline tests that were commented out here https://github.com/FirelyTeam/Vonk/pull/2090

baseTwo:
@richfirely - Please have a look at the validations I put in for ResourceFileName. Each part (e.g. identifier and version) cannot take an underscore as part of the spec. This means the cql file names in the ncqa build artifacts are not valid
image

I removed this rule for now

richfirely (approved):
Looks good. This covers most.. We may need to handle dashes in library names and any dot notation for namespace in the future.

PR #535, opened 2024-09-12
533 Quote strings to c#, escaping as necessary

### LINKED ISSUE #533: Bug - CQL Tag strings must be properly string escaped in C#
The following cql generates invalid c#
```cql
/*
@description: Conditions of type 'Injury due to falling rock' within the measurement period
@fhirquery: akin to Condition?code:in=http://moh.alpha.alp/ValueSet/DiagnosisInjuryDueToFallingRock&onset-date=sa[Period-start]&onset-date=eb[Period-end]
@datarequirement: Condition http://hl7.org/fhir/StructureDefinition/Encounter ["code","onset.ofType(DateTime)","subject.ofType(Patient)"]
@coderequirement: Condition.code http://moh.alpha.alp/ValueSet/DiagnosisInjuryDueToFallingRock
*/
define "Injury due to falling rock within measurement period":
    [Condition: "Injury due to falling rock"] C
       where (C.onset.value as DateTime) during "Measurement Period"
```

Notice that the double-quotes are not escaped 
Image

[test cql files.zip](https://github.com/user-attachments/files/16975614/test.cql.files.zip)

Other issues created on this dataset:  #533, #534

baseTwo:
Work for #533 
Escape strings when writing out c# literals.

e.g. before
image

after
image

### DIFF
--- a/Cql/CodeGeneration.NET/CSharpLibrarySetToStreamsWriter.cs
+++ b/Cql/CodeGeneration.NET/CSharpLibrarySetToStreamsWriter.cs
@@ -20,6 +20,7 @@ using System.Linq.Expressions;
 using System.Reflection;
 using System.Text;
 using Hl7.Cql.Compiler;
+using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.Extensions.Options;
 
 namespace Hl7.Cql.CodeGeneration.NET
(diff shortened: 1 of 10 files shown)


EvanMachusak (changes requested):
Please add unit tests for this functionality, both for `QuoteString` and for an actual CQL snippet.  You can add a test to CqlToElmTests so you can use the `MakeLibrary` function to write a small snippet of CQL demonstrating the error.

EvanMachusak (approved):

PR #536, opened 2024-09-12
534 bug c member access on casted variable needs additional paranthesis

### LINKED ISSUE #534: Bug - C# member access on casted variable needs additional parenthesis
The follow cql generates invalid c#

```cql
define "Injury due to falling rock within measurement period":
    [Condition: "Injury due to falling rock"] C
       where (C.onset.value as DateTime) during "Measurement Period"
```

Notice the missing parantheses around the casts
image

And this is the correct c#
image

[test cql files.zip](https://github.com/user-attachments/files/16975614/test.cql.files.zip)

Other issues created on this dataset:  #533, #534

baseTwo:
Work for #534 

* Improved handling of parentheses when generating c#. See `ExpressionToCSharpConverter.ParanthesizeIfNeeded(...)` (previously called `Paranthesize(..)`. One extra rule was added: if the term starts with `(` xor term ends with `)`, then parantheses are added
* Removed unnecessary ()'s. Use the method above instead of hardcoding them in strings.
* Regen C# 
* No unit tests, but `RR24.cql` was updated in CoreTests which tests the code generation during build


ewoutkramer (dismissed):

Cql/CodeGeneration.NET/ExpressionToCSharpConverter.cs
```diff
@@ -670,13 +672,23 @@ private string PrettyMethodName(MethodBase method)
...
 
-        private static string Parenthesize(string term)
+        private string ParenthesizeIfNeeded(string term)
         {
             term = term.Trim();
-            if (term.StartsWith('(') && term.EndsWith(')')) return term;
 
-            // if (term.ToCharArray().Any(char.IsWhiteSpace))
-            //     return $"({term})";
+            if (term.StartsWith('(') && term.EndsWith(')'))
```
ewoutkramer: When I tried to restructure this a while ago, this bit me, since an expression like `(expr) opr (expr)` has a beginning and ending bracket, but not in a way that we mean it here, so the precedence of opr in such an expression still matters. Well, if this currently gives the right code, we are somehow not producing these kind of expressions anymore, but it's fragile.

ewoutkramer (changes requested):

baseTwo (commented):
Looks good enough, but I can't approve 🤡

ewoutkramer (approved):

PR #546, opened 2024-09-17
262 refactor   rename nameandversion/librarykey to versionedidentifer 2

baseTwo:
Work for 
* #262 

Make the term `VersionIdentifier` the standard when referring to an ELM library by its name and/or version. Renamed any case of `NameAndVersion` to `VersionedIdentifier` to make it all consistent.

Keep the `GetVersionedIdentifier()` method as internal. I'd like to use the Result pattern for the future instead of passing in a boolean to throw an error or not.


richfirely:
Since your doing this shouldn't we take this opportunity to make this also remove the need for getting parts and building strings? This is off the top of my head so please forgive.
VersionedIdentifier .ToFileName <-- which always has the dash naming convention that we agreed on.
VersionedIdentifier .Name
VersionedIdentifier .Version
VersionedIdentifier .ToString() -- which results in the proper pipe format, but no pipe if version is null.
VersionedIdentifier .ToUri or ToCanonical("https...") which results in the full uri

baseTwo:
> Since your doing this shouldn't we take this opportunity to make this also remove the need for getting parts and building strings? This is off the top of my head so please forgive. VersionedIdentifier .ToFileName <-- which always has the dash naming convention that we agreed on. VersionedIdentifier .Name VersionedIdentifier .Version VersionedIdentifier .ToString() -- which results in the proper pipe format, but no pipe if version is null. VersionedIdentifier .ToUri or ToCanonical("https...") which results in the full uri

Sounds like a good idea, can we do this as part of another issue? 
* #550

Cql/Cql.Compiler/LibraryExpressionBuilderContext.cs
```diff
@@ -13,27 +13,40 @@
...
+        LibraryVersionedIdentifier = Library.GetVersionedIdentifier()!;
+    }
+
     private static readonly AmbiguousOverloadCorrector AmbiguousOverloadCorrector = new AmbiguousOverloadCorrector();
 
     /// <summary>
     /// Gets the library associated with the expression builder context.
     /// </summary>
-    public Library Library => library;
+    public Library Library { get; }
```
ewoutkramer, line 43: Re-using the class constructor parameters is exactly the pattern microsoft wants us to use, why are you taking the long route here?

ewoutkramer:
My main remark is the same as @richfirely 's - there is actually a `VersionedIdentifier` type, so I always thought the name "NameAndVersion" meant a string with both components, and we'd use "VersionedIdentifier" for when we talk about the type.  So, this is a necessary first stap (I am happy we ditch the unnecessary LibraryKey term), but it's confusing the `GetVersionedIdentifier()` function does not return a `VersionedIdentifier`.

ewoutkramer (approved):

Cql/Cql.Compiler/LibraryExpressionBuilderContext.cs
baseTwo: This is public property is mapping to a primary constructor parameter. This is a class, which is not the same as a record when it also becomes the property

baseTwo:
> * VersionedIdentifier enhancement #550



> My main remark is the same as @richfirely 's - there is actually a `VersionedIdentifier` type, so I always thought the name "NameAndVersion" meant a string with both components, and we'd use "VersionedIdentifier" for when we talk about the type. So, this is a necessary first stap (I am happy we ditch the unnecessary LibraryKey term), but it's confusing the `GetVersionedIdentifier()` function does not return a `VersionedIdentifier`.

Will be addressed in 

* #550

PR #549, opened 2024-09-17
575 Investigate static libraries on 2.0

### LINKED ISSUE #575: Performance hit with numerous class initializations (2.0)
This is the same ticket as below, only for the 2.0 branch:
- #547

baseTwo:
Work for 
- #575

This PR is to look at the applied changes the PR below and the effect it has on the generated C# libraries. These libraries are all static now. Will be discussed internally before accepting.
* PR #548

Additional changes applied from the PR:
* Make libraries non-static
* Inject library dependencies via primary constructor
* Add extensions to build libraries on IServiceCollection


baseTwo:
This was a conceptual PR. The 1.0 implementation is to use singleton libraries, so it would be best to get 2.0 consistent with the current 1.0 implementation

PR #552, opened 2024-09-18
547 Prototype to make library types singleton instead of static on 1.0

### LINKED ISSUE #547: Performance hit with numerous class initializations (1.0)

baseTwo:
DO NOT MERGE THIS. Performance testing needs to happen to compare to static library types, and if found to be comparable enough, can be cherry picked back into PR #548


DamienMosier:
Ran a performance test on a few measures with a few large decks and performance is comparable between static and singleton.

EvanMachusak:
Part of our public API could be something like this:
```
public static IServiceProvider AddLibraries(this IServiceProvider isp, LibrarySet libraries)
{
  ... compile the library set to assemblies, and then register their types as singletons
}

baseTwo:
Closed, the work was cherry picked into the merged PR #548

PR #577, opened 2024-09-26
570 bug on handling conversion from a choice type 2

### LINKED ISSUE #570: Bug on handling conversion from a choice type

baseTwo:
Work in Progress for #570
Continuation for PR #571 
This should be considered conceptual and used for discussion only at this stage

While developing I found a lot of conversions don't exist, for example in `MATGlobalCommonFunctionsFHIR4.Patient.Normalize Interval`

```
Could not resolve on or more case expressions for Choice type Choice<{http://hl7.org/fhir}dateTime, {http://hl7.org/fhir}Period, {http://hl7.org/fhir}Timing, {http://hl7.org/fhir}instant, {http://hl7.org/fhir}string, {http://hl7.org/fhir}Age, {http://hl7.org/fhir}Range>.
These conversions could not be resolved:
- Hl7.Fhir.Model.Period to Hl7.Fhir.Model.FhirDateTime
- Hl7.Fhir.Model.Timing to Hl7.Fhir.Model.FhirDateTime
- Hl7.Fhir.Model.Instant to Hl7.Fhir.Model.FhirDateTime
- Hl7.Fhir.Model.FhirString to Hl7.Fhir.Model.FhirDateTime
- Hl7.Fhir.Model.Age to Hl7.Fhir.Model.FhirDateTime
- Hl7.Fhir.Model.Range to Hl7.Fhir.Model.FhirDateTime
```

A setting was added in ExpressionBuilderOptions to either throw or ignore these cases. If it is ignored, the C# code will show in a comment which types were missing.
```csharp
CqlDateTime g_ = f_ switch { null => null , CqlDateTime a => a/* , FhirDecimal => ??? */, _ => throw new System.Diagnostics.UnreachableException() };
```


baseTwo:
Work continues in #608

PR #585, opened 2024-10-07
575 performance hit with numerous class initializations 2.0

### LINKED ISSUE #575: Performance hit with numerous class initializations (2.0)

baseTwo:
Work for 
* #575 

Updating libraries to only be accessible via the singleton pattern (single static instance property, and private constructor)

Introduced a static interface for this called `ISingleton<TSelf>` which contains the property `static TSelf Instance { get; }`

Introduced another interface `ILibrary` which has three properties: `string Name { get; }`, `string Version { get; }` and `ILibrary[] Dependencies { get; }`. Having this interface makes it possible to avoid getting to dependencies via reflection. (We could probably add additional members here to make it easier to get to valuesets, etc)


baseTwo:
Additional requirements were added to the issue, so changing back to draft

baseTwo:
Opened again, might be worthwhile to split this issue into two parts with the comments that were added to the original issue.

ewoutkramer (approved):

PR #590, opened 2024-10-08
 Performance update for 1.0 

### LINKED ISSUE #588: Measure performance issues 1.0
Unfortunately with the previous singleton change, I only ran a handful of measures and noticed an ~60% gain. When running all of our measures I noticed some measures that relied heavily on the caching took a huge performance hit. These are the execution time on all of our measures/decks on the same build machine for our nightly runs.

Before the singleton updates when we still had Lazy variables.

ADD-E: Duration: 57m 4s
CRE: Duration: 1h 41m 27s
PND-E: Duration: 44m 15s
SPD: Duration: 9h 24m 18s

After singleton

ADD-E: Duration: 3h 32m 35s
CRE: Duration: 2h 46m 1s
PND-E: Duration: 10h 48m 39s
SPD: Duration: 3h 10m 21s

Here is the total run time of all measures on our build machine.

Latest 1.0 with singleton changes

Image

I realized we needed an update to use scoped measures and singleton libraries. Here is the same run on our build machine with local changes in the engine to have both. No CQL changes were made to have a proper apples to apples comparison of the CQL performance.

Image

ADD-E: Duration: 38m 0s
CRE: Duration: 57m 4s
PND-E: Duration: 53m 39s
SPD: Duration: 1h 36m 22s

As you can see, huge performance gains using singleton libraries to reduce the constructor initialization of all supporting libraries as well as improved performance to use caching/lazy variables inside the measure CQL to improve the performance when you write 1 CQL define that is used in multiple places, like a list of episode dates.

DamienMosier:
- restore the lazy loading and constructors but only for measures with context defined
- add lazy null handling
- addresses #588

Current 1.0

ADD-E: Duration: 3h 32m 35s
CRE: Duration: 2h 46m 1s
PND-E: Duration: 10h 48m 39s
SPD: Duration: 3h 10m 21s

Time to run 80 HEDIS measures with multiple configurations using current develop 1.0 branch
image

With these updates, same build pipeline

ADD-E: Duration: 38m 0s
CRE: Duration: 57m 4s
PND-E: Duration: 53m 39s
SPD: Duration: 1h 36m 22s

image

### DIFF
--- a/Cql/CodeGeneration.NET/CSharpSourceCodeWriter.cs
+++ b/Cql/CodeGeneration.NET/CSharpSourceCodeWriter.cs
@@ -208,14 +208,15 @@ namespace Hl7.Cql.CodeGeneration.NET
 
             var hasContext = false;
             var node = dependencyGraph.Nodes[libraryName];
-            if (node.Properties != null && node.Properties.ContainsKey("Library"))
+            if (node.Properties != null && node.Properties.TryGetValue("Library", out var nodeLibrary))
             {
-                hasContext = ((Elm.Library)node.Properties["Library"])?.contexts != null;
+                hasContext = ((Elm.Library)nodeLibrary)?.contexts != null;
             }
 
             // Class
             {
                 // singleton for libraries, constructors for measures
+                // files that have a context defined will get the constructor with lazy for caching
                 if (!hasContext)
                 {
                     writer.WriteLine(indentLevel, $"public static {className} Instance {{ get; }}  = new();");
@@ -435,7 +436,14 @@ namespace Hl7.Cql.CodeGeneration.NET
 
                 if (useLazy)
                 {
-                    writer.Write(expressionConverter.ConvertTopLevelFunctionDefinition(indentLevel, valueFunc, methodName!, "public"));
+                    // add null handler because, in some rare cases the order of the Lazy matters and a null exception is thrown in the constructor
+                    // but having it as lazy?.Value handles this without degrading performance or affecting result of execution 
+                    var valueFunction =
+                        expressionConverter.ConvertTopLevelFunctionDefinition(indentLevel, valueFunc, methodName!,
+                            "public");
+
+                    valueFunction = valueFunction.Replace(".Value", "?.Value");
+                    writer.Write(valueFunction);
                 }
                 else
                 {
@@ -447,7 +455,6 @@ namespace Hl7.Cql.CodeGeneration.NET
                 writer.WriteLine(indentLevel, $"[CqlDeclaration(\"{cqlName}\")]");
                 WriteTags(writer, indentLevel, tags);
                 writer.Write(expressionConverter.ConvertTopLevelFunctionDefinition(indentLevel, overload, methodName!, "public"));
-                //      writer.WriteLine();
             }
         }
 
--- a/Cql/CodeGeneration.NET/CSharpSourceCodeWriter.cs
+++ b/Cql/CodeGeneration.NET/CSharpSourceCodeWriter.cs
@@ -94,6 +94,8 @@ namespace Hl7.Cql.CodeGeneration.NET
 
         internal ILogger<CSharpSourceCodeWriter> Log { get; }
 
+        internal IList<string> ContextLibraries { get; set; } = new List<string>();
+
         /// <summary>
         /// Writes C# source code from inputs.
         /// </summary>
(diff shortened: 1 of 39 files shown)


DamienMosier:
I could use some feedback on why the CQL Test exclusion is failing. The CQL and demo project build and the unit tests pass in the Demo project.

ewoutkramer:
@baseTwo - is this something we need in your equivalent 2.0 port as well?

baseTwo:
It looks like the generated c# code in the Demo project reverted to the old way? Could these be regenerated as part of this PR to demonstrate how they ought to look like now?

baseTwo:
> @baseTwo - is this something we need in your equivalent 2.0 port as well?

We have an issue to add memoization as part of the context https://github.com/FirelyTeam/firely-cql-sdk/issues/600, however, it looks like 1.0 changed back to the old way where most libraries contain state for the context and the Lazy<>'s. 

@DamienMosier , which measures do not have context defined, I'd like to compare the two types, but GitHub collapses too many files for me to go through each one until I find it.

DamienMosier:
@ewoutkramer this is something you would definitely want in the 2.0 branch for the performance gains.

@baseTwo The way the logic work is if the file or one of the includes has a context then it creates it the old way with a constructor and utilizes Lazy<T>. If the file does not, aka a library, then it's a singleton. 

For our use case only our actual measures have context Patient and all of our libraries are either functions or fluent functions with the resources passed in. In your Demo project, context is added to almost all files which is why it looks like the old way.

The main performance issue with the original 1.0 branch was with the constructors. If you have LibraryA that has 5 includes and each of those subsequent libraries each have their own set of includes. You could end up instantiating the same library multiple times. Having it scoped allowed us to use Lazy<T> so we gained the benefit of some caching but at the expense of calling the constructors a lot. Switching to singleton removed that constructor lift but removed the ability to use Lazy<T>. I thought about it and realized we only need Lazy<T> where there is context and all subsequent libraries did not need it and we could have the best of both worlds.

Here are some partial examples of measure AAB from DQIC if you want to see the CQL.

```
[System.CodeDom.Compiler.GeneratedCode(".NET Code Generation", "1.0.0.0")]
[CqlLibrary("AAB_Concepts", "2024.1.0")]
public class AAB_Concepts_2024_1_0
{

    public static AAB_Concepts_2024_1_0 Instance { get; }  = new();

    #region Dependencies
    #endregion

    [CqlDeclaration("Acute Bronchitis")]
    [CqlValueSet("https://www.ncqa.org/fhir/valueset/2.16.840.1.113883.3.464.1004.1016")]
	public CqlValueSet Acute_Bronchitis(CqlContext context) => 
		new CqlValueSet("https://www.ncqa.org/fhir/valueset/2.16.840.1.113883.3.464.1004.1016", null);

    [CqlDeclaration("Competing Diagnosis")]
    [CqlValueSet("https://www.ncqa.org/fhir/valueset/2.16.840.1.113883.3.464.1004.1067")]
	public CqlValueSet Competing_Diagnosis(CqlContext context) => 
		new CqlValueSet("https://www.ncqa.org/fhir/valueset/2.16.840.1.113883.3.464.1004.1067", null);

```
and the actual file with context is reverted to the old way, the same way you see files in the DQIC repository.

```
[System.CodeDom.Compiler.GeneratedCode(".NET Code Generation", "1.0.0.0")]
[CqlLibrary("AAB_Reporting", "2024.1.0")]
public class AAB_Reporting_2024_1_0
{
    internal CqlContext context;

    #region Cached values

    internal Lazy<Tuples.Tuple_BOFQdLfCiMegGCeQZaSIJRaC> __Measure;
    internal Lazy<Patient> __Patient;
    internal Lazy<IEnumerable<ExplanationOfBenefit>> __Eobs;
    internal Lazy<IEnumerable<ExplanationOfBenefit>> __Admin_Eobs;
    
    #endregion
    public AAB_Reporting_2024_1_0(CqlContext context)
    {
        this.context = context ?? throw new ArgumentNullException("context");
        
        __Measure = new Lazy<Tuples.Tuple_BOFQdLfCiMegGCeQZaSIJRaC>(this.Measure_Value(context));
        __Patient = new Lazy<Patient>(this.Patient_Value(context));
        __Eobs = new Lazy<IEnumerable<ExplanationOfBenefit>>(this.Eobs_Value(context));
        __Admin_Eobs = new Lazy<IEnumerable<ExplanationOfBenefit>>(this.Admin_Eobs_Value(context));
```

DamienMosier:
For reference, I generated this for Evan after doing testing. The number are based off our build machine times. The "1.0 DCS" is the original 1.0 prior to singleton. The "1.0 Develop" is current develop with only singleton's and the final are these changes that are pending.

image

ewoutkramer (approved):

PR #594, opened 2024-10-09
Idea to fix issue 592 by using ExpandoObject and dynamic instead of ValueTuples

### LINKED ISSUE #592: ValueTuple objects created from CQL Tuple types are missing field names
**Describe the bug**

CQL Tuple types are currently mapped to C# ValueTuples.
Unfortunately, this has the effect that the field names of the tuple are not available at runtime as the compiler replaces all custom names with default names like `Item1`, `Item2` etc. The custom names are not part of the runtime type information.

This is documented [here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples#tuple-field-names).

This means that a CQL tuple like

```
Tuple{
    status: 'success',
    result: 'some result'
  }
```

is mapped to a C# ValueTuple like

```csharp
('success', 'some result')
```

or

```csharp
(Item1: 'success', Item2: 'some result')
```

The field names (`status` and `result`) are lost in the conversion.

**To Reproduce**

1. Create a simple CQL file that defines a tuple like `Tuple{status: 'success', result: 'some result'}`
2. Compile the file into a C# assembly, e.g. using the cql-packager tool.
3. Load the assembly and access the tuple
4. Inspect the tuple object to check for the existing fieldnames. Field names are only available as `Item1` or `Item2`. The custom field names `status` or `result` are unknown.

~~I will provide a PR to demonstrate the issue in some unit tests.~~
This PR demonstrates the issue in some unit tests: #593 

**Expected behavior**

The C# object created from the CQL tuple should contain the field names.

**Screenshots**

N/A

**Version used:**

Tested with Hl7.Cql.Packager Version 2.0.4-alpha and HEAD of develop-2.0 branch (09-Oct-2024, cf842e95).

**Additional context**

It seems as if `ValueTuple` is not a suitable C# type to represent the CQL Tuple Type.
One alternative might be a combination of `ExpandoObject` and `dynamic`. 

~~I will provide a PR to demonstrate this idea.~~
This PR demonstrates this idea and might serve as a basis for further discussion: #594

and-hus:
Demonstrating an idea/possible approach to fix issue #592 

Refactors the *Expression to C# conversion* code especially for Tuples to use `ExpandoObject` and `dynamic` instead of `ValueTuples`.

Adds some testdata and  two unit-tests to demonstrate how the new code might be an approach to fix #592.
See #592 for a detailed description of the issue.

> [!Note]
This PR is probably not ready to be merged. It just demonstrates an idea and might serve as a basis for discussions or to build a proper solution.

# Summary

## General cleanup in `ExpressionToCsharpConverter.cs`

* move the only public method (`ConvertTopLevelFunctionDefinition(…)`) to the top to make it more prominent and easier to spot
* make `ConvertExpression()` method privat as it's not used outside of this class

## Using dynamic/ExpandoObject as dotnet type for CQL Tuples

* `CSharpLibrarySetStreamsWriter`
  * Add `System.Dynamic` to list of using statements for generated code to enable the use of `ExpandoObject`

* `ExpressionToCsharpConverter`
  * Catch special case of tuple initialization in `ConvertBinaryExpression()` and handle it via new method `ConvertTupleInitExpression()`, it's necessary to handle this here and not in `ConvertMemberInititExpression()` as we need the name of the parameter, which is not available in `ConvertMemberInitExpression()`
  * Created private method `ConvertTupleInitExpression()` to generate code like

    ```csharp
    dynamic paramName = new ExpandoObject();
    paramName.member1 = value1;
    paramName.member2 = value2;
    ```
  
  * Remove handling the special case of tuples in `ConvertMemberInitExpression()`, as the special case is now already handled in `ConvertBinaryExpression()`
  
* `TypeToCSharpConverter`
  * Refactor `FormatTypeNameAsTuple()` to return "`dynamic`" as the C# typename for CQL tuples

## Tests

* Adding test data to `Input/ELM/HL7/`:
  * `CqlTupleTest.cql`: CQL file that returns a simple tuple
  * `CqlTupleTest.json`: ELM representation of CqlTupleTest.cql
  * `CqlNestedTupleTest.cql`: CQL file that returns a nested tuple
  * `CqlNestedTupleTest.json`: ELM representation of CqlNestedTupleTest.cql

* `Demo/Cql/Build/ElmToSharp.Targets.xml`:
  * Add additional packager cli argument to export the assemblies to the configured directory (`DllDirectory`)

* `Cql/CoreTests/CoreTests.csproj`:
  * Define `CSharpDirectory` and `DllDirectory` property to let the packager export C# assemblies to new `Assemblies` directory
  * Copy new `Assemblies` directory to the build output, to be able to load the generated assembly in tests

* Generated files in `Cql/CoreTest/CSharp`:
  * `CqlTupleTest-1.0.0.g.cs`: auto-generated from `Input/ELM/HL7/CqlTupleTest.json` during build step
  * `CqlNestedTupleTest-1.0.0.g.cs`: auto-generated from `Input/ELM/HL7/CqlNestedTupleTest.json` during build step
  * other files are changed for the additional using statement

* New tests in `Cql/CoreTests/CodeGenerationNet`:
  * `CqlTupleTest.cs`: Demonstrating that the Tuples from `CqlTupleTest.cql` and `CqlNestedTupleTest.cql` are properly converted into dynamic ExpandoObjects and all properties/fields can be accessed easily, even across assembly boundaries
  * `ExpressionToCSharpConverterTest.cs`: Small unit test that helps to debug the process of the code generation, especially to debug the new/changed code


baseTwo:
Thank you for your ideas.

I'd like to provide some background context to the C# implementation of the CQL SDK. Many of the other language implementations use dynamic quite extensively, however a very conscious design decision when we started the C# implementation, was to avoid dynamic completely. So while value tuples lose their property names at runtime, we chose this to be acceptable. 

I've also taken note that a file was reformatted from block to file scoped namespaces. This causes quite a lot of noise when viewing file history, so we only use file scoped namespaces for new files. We have a task #290 which outlines the cleanup work we plan to do, however, this is not one of the items on that list.

As for minor refactorings, changes committed to a PR must be surgical to the associated issue at hand.

We are always open for ideas, feel free share them via the [ideas](https://github.com/FirelyTeam/firely-cql-sdk/discussions/ideas) discussion page.

cc @ewoutkramer and @EvanMachusak who may be able to provide additional context

and-hus:
Thanks a lot for the quick response and for providing the background information. 
I agree that adding the general cleanup stuff (and especially the change from block to file scoped namespaces) into this PR makes it hard to review and spot the relevant changes. I just pushed an update to undo some of these changes. 

I also fully agree with your design decision to avoid `dynamic`. It's funny, because usually I am the one convincing others not to use it ;-) Actually, I was hesitating for quite some time and checking other alternatives, but finally decided to bite the bullet ...

It's interesting that you actively decided to accept the fact that the value tuples lose their property names at runtime, because this causes a huge issue for our use case. I would love to discuss this further. Unfortunately, the ideas page that you link to is not accessible for me - I am getting a 404. It's probably a team-internal page...

What do you think about moving the discussion to the issue that I created ( #592 ) ?

baseTwo:
> Thanks a lot for the quick response and for providing the background information. I agree that adding the general cleanup stuff (and especially the change from block to file scoped namespaces) into this PR makes it hard to review and spot the relevant changes. I just pushed an update to undo some of these changes.
> 
> I also fully agree with your design decision to avoid `dynamic`. It's funny, because usually I am the one convincing others not to use it ;-) Actually, I was hesitating for quite some time and checking other alternatives, but finally decided to bite the bullet ...
> 
> It's interesting that you actively decided to accept the fact that the value tuples lose their property names at runtime, because this causes a huge issue for our use case. I would love to discuss this further. Unfortunately, the ideas page that you link to is not accessible for me - I am getting a 404. It's probably a team-internal page...
> 
> What do you think about moving the discussion to the issue that I created ( #592 ) ?

I will continue the conversation in #592 :)

baseTwo:
Closed this PR since we want to avoid dynamic code. The issue was moved into the current sprint, there are 3 proposals under discussion https://github.com/FirelyTeam/firely-cql-sdk/pull/599

PR #599, opened 2024-10-13
592 Three proposals to fix broken serialization of generated C# cql tuples

### LINKED ISSUE #592: ValueTuple objects created from CQL Tuple types are missing field names

baseTwo:
Proposals for 
- #592

⚠️Discussion Point PR - Actual Work will follow from outcome
Also see discussion from #592
There are currently three proposals. The PR contains all of them, please read here first to see which code contains which proposal

## 1. Structural Values

Associated code: see `StructuralValueTupleTests.cs` `Hl7.Cql.Runtime` : `IStructuralValue`, `IStructuralValue<T1, ....>`, `IStructuralValueMetadataProvider`, `StructuralValueMetadata`

Unit test `StructuralValueTupleBasicTests` tests equality, formatting, hashcode and cloning. The second test `StructuralValueTupleSerializationTests`, tests serialization. Note that no special converters are necessary. 

The C# generated tuples that would go into the libraries are `TupleA`, `TupleB`, `TupleAddress`, `TuplePerson`. The type names would be deterministically generated from the properties like the 1.0 implementation. Each library would have these as nested types, and output from one method would clone into an equivalent type as input on another library

Pros:
* Value types are fast to use and clone from other equivalent types
* No need for a special JSON/XML Serializer
* Supports equality, string formatting, hashcodes as this is generated in each tuple

Cons:
* A lot of infrastructure code required
* For every n-th item in the tuple, and interface is needed, so some upper limit will be needed for `IStructurableValue<T1,T2,.... Tn>
* Needs a more complex refactoring in the C# code generator
* A lot of additional types generated in the C#, each tuple type used in a library will be a nested type to that library

----
## 2. CqlTuple<T> containing properties referencing static metadata and a value tuple, and custom serialization
There is also a second idea called a CqlTuple<T> which takes a static list of property names, and a value tuple. 

Associated code:
* All code in file `CqlTupleTests`

Pros:
* Very little infrastructure code required. (CqlTuple<> type)
* Needs a some refactoring in the C# code generator
* Supports equality, string formatting, hashcodes as this is generated in each CqlTuple

Cons:
* Ugly code mixing generic type with value tuples
* Requires special JSON/XML Serializer
* Equality and hashcode may differ on equivalent tuples if the first value does not contain the same `string[]` reference

---
## 3. Plain ValueTuples with additional value referencing static metadata, and custom serialization
And lastly a value tuple-only approach, which uses value tuples where the first value points to some metadata with property names, and would be a static property on the library.

Associated code:
* All code in `ValueTupleTests`

Pros:
* Easy and quick to implement: we keep the current code mostly as it is.
* Very little infrastructure code required. (`TupleMetadata` type)
* Simple code usage

Cons:
* Requires special JSON/XML Serializer
* Equality and hashcode may differ on equivalent tuples if the first value does not contain the same `TupleMetadata` reference, so between libraries this won't work.

---
## Summary
The key difference between proposal 1 and the rest are that 1 doesn't need a special serializer
If the need to use a special serializer is not an issue, then proposal 3 is the simplest and quickest solution


baseTwo:
Moved the discussion to the repo's [Discussion page](https://github.com/FirelyTeam/firely-cql-sdk/discussions/605)

baseTwo:
Closing. Proposal 3 selected from [discussion](https://github.com/FirelyTeam/firely-cql-sdk/discussions/605). PR #603 will implement this.

PR #603, opened 2024-10-14
592 Fix CQL Value Tuples so that they are serializable with custom Json Converters

baseTwo:
Fix for 
* #592 

Also see discussion
* #605

## Value Tuples must have first item of type `CqlTupleMetadata`

Fixes value tuples by including its metadata as first item in the tuple. The metadata is stored as static property on the library and it contains the item names and types. At this stage only the item names are required for serialization, but it doesn't hurt to include the item types as well for future possibilities. (Also, it always adds value to see types in the generated c# code)

New types added to `Hl7.Cql.Runtime`:
* `class Hl7.Cql.Runtime.CqlTupleMetadata`
* `class Hl7.Cql.Runtime.Serialization.CqlValueTupleJsonConverter`
* `class Hl7.Cql.Runtime.Serialization.CqlValueTupleJsonConverterFactory`

Libraries will have private static properties for the tuple metadata as follows:
```csharp
    private static CqlTupleMetadata CqlTupleMetadata_FPVdJEEOUKFVhQPTfKhShUNO = new([typeof(IEnumerable<Practitioner>)], ["Practitioners"]);
```

and usage in the library is as follows:
```csharp
// Note that CqlTupleMetadata is always the first type in a value tuple
public (CqlTupleMetadata, int? IdentifierCount)? Get_Prescriber_NPI_from_Claims(CqlContext context, IEnumerable<Claim> claim) {

   // Declaring a new value tuple
   (CqlTupleMetadata, IEnumerable<Practitioner> Practitioners)? aw_ = (CqlTupleMetadata_FPVdJEEOUKFVhQPTfKhShUNO, av_);

}
```

## Serialization

A unit test was added, see `CqlTupleTests` in CoreTests project. The `CqlValueTupleJsonConverterFactory` must be added as a converter in the `JsonSerializationOptions`

## Versioning
Version bumped to 2.0.6-alpha

## Refactoring

The `ExpressionToCSharpConverter` was cleaned up, so that indentation parameters passed between methods are now contained in a context type. The context also needed an additional parameter to build tuple property names. The `TupleMetadataBuilder` is carried as a dependency 

The same cleanup was appied on `CSharpLibrarySetToStreamsWriter`, to reduce the parameter count being passed around, instead wrapping them up into contexts. The `TupleMetadataBuilder` is created here and passed along the contents until needed. 

I noticed libraries were indented with a mix of tabs and spaces. This was fixed to use spaces only.


baseTwo:
cc @and-hus

baseTwo:
Will fix compiler error soon

baseTwo:
The SDK version need to be incremented when the generated c# changes significantly...

and-hus:
> cc @and-hus

Thanks for the ping - looks interesting. I am just taking a closer look at it...

and-hus:
@baseTwo :
I am not sure if it's related to my own setup, but I am getting a weird error message like
Error

when trying to setup a simple unit test to verify our use-case...

Just added some CQL example for a nested tuple and compiled to a dll. 
I am getting the error when loading the dll and trying to run the code like this:
```csharp
[TestMethod]
public void ExpressionReturningNestedTuples_ResultCanBeSerialized()
{
    var file = @"Dlls/CqlNestedTupleTest-1.0.0.dll";
    var filePath = Path.GetFullPath(file);

    var asm = new AssemblyLoadContext("CqlNestedTupleTest");
    asm.LoadFromAssemblyPath(filePath);

    var ctx = FhirCqlContext.ForBundle();
    var result = asm.Run("CqlNestedTupleTest", "1.0.0", ctx);
    Assert.IsNotNull(result);

    result.TryGetValue("Result", out var obj);
    Assert.IsNotNull(obj);
    Assert.IsInstanceOfType(obj, typeof(ValueTuple));

    // serialization works out of the box
    //var str = JsonSerializer.Serialize(obj);
    //Assert.AreEqual("{\"status\":\"success\",\"result\":\"some result\"}", str);
}
```

This is the patch file to add the test:
[0001-Add-unit-test-to-retrieve-tuple-across-assembly-boun.patch](https://github.com/user-attachments/files/17393999/0001-Add-unit-test-to-retrieve-tuple-across-assembly-boun.patch)

and-hus:
@baseTwo 
It seems as if the code-generation creates only a private default constructor:

```csharp
[System.CodeDom.Compiler.GeneratedCode(".NET Code Generation", "2.0.6.0")]
[CqlLibrary("CqlNestedTupleTest", "1.0.0")]
public partial class CqlNestedTupleTest_1_0_0 : ILibrary, ISingleton<CqlNestedTupleTest_1_0_0>
{
    private CqlNestedTupleTest_1_0_0() {}
```

baseTwo:
> @baseTwo It seems as if the code-generation creates only a private default constructor:
> 
> ```cs
> [System.CodeDom.Compiler.GeneratedCode(".NET Code Generation", "2.0.6.0")]
> [CqlLibrary("CqlNestedTupleTest", "1.0.0")]
> public partial class CqlNestedTupleTest_1_0_0 : ILibrary, ISingleton<CqlNestedTupleTest_1_0_0>
> {
>     private CqlNestedTupleTest_1_0_0() {}
> ```

That's correct. For performance reasons, libraries are accessible through a singleton property `Instance` on the library type. (see PR# 585. This has been applied on v1.0 too.

It seems the AssemblyLoadContext.Run method still needs to be updated for this. I'll add a ticket to fix that, and add some unit tests for it as well.

EvanMachusak:
Why `CqlTupleMetadata_FPVdJEEOUKFVhQPTfKhShUNO`?  Are you still creating new types programmatically?

The purpose of shifting away from having tuple types dynamically generated was to prevent the awkwardness of having a separate `Tuples` assembly that was scoped to a particular build.  It created dependency management problems and was hard to operationalize in a system with many different libraries.

Shouldn't `CqlTupleMetadata` literally just be a `string[]`?  Tuples have `Item1`, `Item2`, and so on.  The only data you need is what the actual member name of `Item1` was, so that `define tuple: Tuple { x: 1, y: 2 }` when translated into C#'s `(1,2)` tuple can remember that the names of those two `Item` props were `x` and `y` so they can be serialized at runtime.

Put another way - if you are still generating types that will by definition have to exist in a separate cross-cutting assembly, then the existing solution of generating the Tuple classes themselves was sufficient already and the entire CQL `Tuple` to C# `Tuple` should just be rolled back entirely.

and-hus:
@baseTwo

I am really sorry. I completely missed the discussion on #605 and I feel bad to throw in another idea so late in the process and after you already spent so much time to implement the current approach. :-(

But have you considered to represent the CQL Tuples in C# as a string-object dictionary (`Dictionary<string, object>`) instead of ValueTuples with the additional CqlMetadata entry for the field names?

Something along these lines:

```csharp
Dictionary<string, object> tuple = new()
{
    { "status", "success" },
    { "result", "some result" },
    { "isOK", true },
    { "someNumber", 3045 },
    { "nestedTuple", new Dictionary<string, object>
    {
        { "nestedValue", "some nested Value"},
        { "2ndNestedValue", "another nested Value"}
    }}
};
```

From my perspective this would have multiple advantages over the current approach:

- using C# native type in a standard way vs. ValueTuple with extra convention for the first element
- doesn't require any extra types like `CqlTupleMetadata`
- serializes directly to JSON, without additional serializer options
- reduced complexity and therefore improved maintainability
- easier to handle in consuming code
- possible to define a type alias via `using CqlTuple = System.Collections.Generic.Dictionary<string, object>;`

I am not sure about the cons, especially with respect to performance or if it would conflict with general design decisions for the SDK.

If you're interested, I am happy to give it a try and provide some sample implementation via a separate PR - what do you think?

ewoutkramer:
> 
> But have you considered to represent the CQL Tuples in C# as a string-object dictionary (`Dictionary<string, object>`) instead of ValueTuples with the additional CqlMetadata entry for the field names?
> 

@EvanMachusak took this approach in an early version of this SDK but found that is was too slow. Many measures, by absence of a data definition language in CQL, produce and work with these anonymous tuples, so the use of a Dictionary versus direct properties becomes a burden. Evan might chime in if he wants to add more about his experience on using this approach.

ewoutkramer:
> Why `CqlTupleMetadata_FPVdJEEOUKFVhQPTfKhShUNO`? Are you still creating new types programmatically?
> 

No, we luckily don't need to generate types anymore! But Paul needed a unique name for a shared instance of the metadata per library. It does not matter if other libraries use the same type of tuple with another instance of the same metadata however, the equality on Tuples takes care of that. It's just making sure you only need to create this metadata once for a type of tuple.

ewoutkramer:
For the reviewers of the generated code: since this PR changes tabs into spaces in the generated code, you need to tell Github to ignore whitespace while making diffs like so:

image

and-hus:
> > But have you considered to represent the CQL Tuples in C# as a string-object dictionary (`Dictionary<string, object>`) instead of ValueTuples with the additional CqlMetadata entry for the field names?
> 
> @EvanMachusak took this approach in an early version of this SDK but found that is was too slow. Many measures, by absence of a data definition language in CQL, produce and work with these anonymous tuples, so the use of a Dictionary versus direct properties becomes a burden. Evan might chime in if he wants to add more about his experience on using this approach.

@ewoutkramer Great - thanks for the background info!

ewoutkramer (approved):

ewoutkramer:
Mmmm....the auto-merge got me here. I reviewed this because I think this is a nice solution, but I think I really wanted @EvanMachusak to take a look at it too.  Seeing Evan's question above on whether this needs generated types (no), then if that was the only hesitancy Evan had, I think it is ok to proceed, but just want to make sure....

Issue #609, opened 2024-10-16
AssemblyLoadContext.Run fails, because libraries are singletons now.


baseTwo:
As a result of #575 changing libraries to singletons, an oversight was to update the `AssembleLoadContextExtensions` which run these libraries.

> @baseTwo It seems as if the code-generation creates only a private default constructor:
> 
> ```cs
> [System.CodeDom.Compiler.GeneratedCode(".NET Code Generation", "2.0.6.0")]
> [CqlLibrary("CqlNestedTupleTest", "1.0.0")]
> public partial class CqlNestedTupleTest_1_0_0 : ILibrary, ISingleton<CqlNestedTupleTest_1_0_0>
> {
>     private CqlNestedTupleTest_1_0_0() {}
> ```
> 
> That's correct. For performance reasons, libraries are accessible through a singleton property `Instance` on the library type. (see PR# 585. > This has been applied on v1.0 too.
> 
> It seems the AssemblyLoadContext.Run method still needs to be updated for this. I'll add a ticket to fix that, and add some unit tests for it as well.
_Originally posted by @baseTwo in https://github.com/FirelyTeam/firely-cql-sdk/issues/603#issuecomment-2416418619_

PR #610, opened 2024-10-16
609 Fix AssemblyLoadContext.Run

baseTwo:
Fix for 
* #609 

⚠️Previous PR #603  must be merged first

Loads and executes singleton libraries correctly in `AssemblyLoadContextExtensions.Run`
Unit test added `AssemblyLoadContextTest`


baseTwo:
cc @and-hus I added your unit tests for running cql nested tuples from an assembly load context and serializing it in this PR

and-hus:
> cc @and-hus I added your unit tests for running cql nested tuples from an assembly load context and serializing it in this PR

Great - thanks!
Just one remark. I see that in serialization test you are loading the tuple from the same assembly like:

```csharp
var obj = CqlNestedTupleTest_1_0_0.Instance.Result(ctx);
```

Be aware that I deliberately loaded and executed from an external assembly to ensure that whole information is passed across assembly boundaries - as this was the main issue with the initial implementation...

0001-Add-unit-test-to-retrieve-tuple-across-assembly-boun.patch
```diff
@@ -0,0 +1,351 @@
+From e430d95c96041334f7c17c64487d9b50697210bc Mon Sep 17 00:00:00 2001
```
and-hus: I don't think that this patch file should be commited to the repo...
I assume it was added by accident.

baseTwo: Well spotted

ewoutkramer (dismissed):

baseTwo:
> > cc @and-hus I added your unit tests for running cql nested tuples from an assembly load context and serializing it in this PR
> 
> Great - thanks! Just one remark. I see that in serialization test you are loading the tuple from the same assembly like:
> 
> ```cs
> var obj = CqlNestedTupleTest_1_0_0.Instance.Result(ctx);
> ```
> 
> Be aware that I deliberately loaded and executed from an external assembly to ensure that whole information is passed across assembly boundaries - as this was the main issue with the initial implementation...

Noted, I could add the unit test to load from external assembly to the one loading directly from a library

ewoutkramer (approved):

PR #614, opened 2024-10-18
613 Fix CqlComparer for tuples

### LINKED ISSUE #613: Fix Integration test - System.ArgumentException: 'Cannot generate a hash code for ValueTuple`3 Arg_ParamName_Name'
## Error
The Integration Test running is failing with the following exception:

```
System.ArgumentException
  HResult=0x80070057
  Message=Cannot generate a hash code for ValueTuple`3 Arg_ParamName_Name
  Source=Hl7.Cql.Comparers
  StackTrace:
   at Hl7.Cql.Comparers.CqlComparers.GetHashCode(Object x) in C:\Dev\firely-cql-sdk\Cql\Cql.Comparers\CqlComparers.cs:line 321
```

at

```
Measures.CMS.dll!UrinarySymptomScoreChangeAfterBenignProstaticHyperplasiaFHIR_1_3_000.Urinary_Symptom_Score_Assessment(Hl7.Cql.Runtime.CqlContext context) Line 707
	at C:\Dev\firely-cql-sdk\Demo\Measures.CMS\CSharp\UrinarySymptomScoreChangeAfterBenignProstaticHyperplasiaFHIR-1.3.000.g.cs(707)
```

while running this test
```
UrinarySymptomScoreChangeAfterBenignProstaticHyperplasiaFHIR.Population_FromResource(testCaseId: "051c5977-9f2c-4e8b-8e02-ac3ec0c718d6")
```

## Background
Tuples emitted to C# have recently changed from being derived from `TupleBaseType` to just value tuples (`ValueTuple<,,,...>` at runtime).

## Solution
* The `CqlComparer` must be updated to support value tuples
* Add unit tests 
* `TupleBaseType` is part of the Cql.Primitives project. It is no longer used at runtime, consider making this internal. It is only used to generate the expression LINQ tree as a temporary placeholder until C# generates value tuples.

baseTwo:
Fix for 
* #613 

## Work
* Replaced `TupleComparer` with `CqlTupleTypeComparer`. The new comparer compares value tuples where the first parameter is of type  `CqlTupleMetadata` and uses the metadata to compare items.
* The legacy `TupleBaseTypeComparer` cannot be removed yet, since some unit tests still execute against directly compiled LambdaExpressions, instead of going through the C# code writer and assembly compiled.
* `CqlTupleMetadata` had to be moved to a more base project and was moved from `Hl7.Cql.Runtime' to `Hl7.Cql.Primitives`
* `TypeExtensions` had to move to a more base project too, and was moved from `Hl7.Cql.Compiler` to `Hl7.Cql.Primitives`. The extensions for this particular `TypeExtensions` is for types related to CQL.
* Added unit test `TypeExtensionsTests`
* Relevant unit tests from CoreTests were moved into subfolders so it is clear from which project they belong to
* A few places in the repo uses a property-name/property-type tuple. Sometimes the type is placed first, other times the name. This was made consistent so the type is always first. 

## Integration Runner
While working on this PR, I also looked at the integration tests for CMS measure. I found they were flaky,  due to the libraries sometimes failing to create. This was due to the CqlTupleMetadata calculating the hashcode in the constructor, and that used Hasher.Instance which uses MD5 to hash a string. For some strange reason, a NullReferenceException is sometimes thrown deep inside the stack trace. I changed the `CqlTupleMetadata` to do hash calculation lazily.

I also discovered that the `KeyValuePairComparer<TKey,TValue>` comparer fails when `TValue` is an `object` when doing a compare. An error is thrown stating that 'object' is not comparable. I fixed this by using the ICqlComparer on key and value instead.

Test pass rate increased from 68% to 72%


Cql/CoreTests/Fhir/DataSourceTests.cs
ewoutkramer, line 1: We'll get a merge conflict here with the PR that reinstates the unit tests.

Cql/Cql.Primitives/Cql.Primitives.csproj
```diff
@@ -16,6 +16,14 @@
 		<PackageReference Remove="Microsoft.CodeAnalysis.PublicApiAnalyzers" />
 	</ItemGroup>
 
+	<ItemGroup>
+		<InternalsVisibleTo Include="TemporaryTupleAssembly" Key="$(LibraryPKHash)" />
```
ewoutkramer: What's this?

Cql/Cql.Comparers/CqlComparers.cs
```diff
@@ -239,14 +232,11 @@ public CqlComparers Unregister(Type type)
                     comparer = enumerableComparer;
                 }
             }
-            else if (x is TupleBaseType && Comparers.TryGetValue(typeof(TupleBaseType), out ICqlComparer? tupleComparer))
```
ewoutkramer, line 242: Why is checking for TupleBaseType not necessary anymore? Will Comparers.TryGetValue() now return it? Or did it actually always already do that?

Cql/Cql.Comparers/KeyValuePairComparer.cs
```diff
@@ -30,9 +30,9 @@ internal class KeyValuePairComparer<TKey,TValue> : ICqlComparer, ICqlComparer<Ke
         KeyValuePair<TKey, TValue> x,
         KeyValuePair<TKey, TValue> y,
         string? precision = null) =>
-        Comparer<TKey>.Default.Compare(x.Key, y.Key) switch
+        cqlComparer.Compare(x.Key, y.Key, precision) switch
         {
-            0     => Comparer<TValue>.Default.Compare(x.Value, y.Value),
+            0     => cqlComparer.Compare(x.Value, y.Value, precision),
```
ewoutkramer, line 35: Was this a bug, the fact that precision wasn't passed on to the compare?

ewoutkramer (commented):
I just have a few questions to double check. Could the change made by passing on precision create regressions for HEDIS?

Cql/Cql.Primitives/Cql.Primitives.csproj
baseTwo: This was needed to get the generated expressions to build, since TupleBaseType was changed to internal, but this only worked in Debug build, not Release. I changed this type back to public, so this line is no longer needed

Cql/Cql.Comparers/CqlComparers.cs
baseTwo: The `GetType` method handles this above. And this is the `GetType` method:

```csharp
 private static Type GetType(object? x)
 {
     var type = x switch
     {
         TupleBaseType => typeof(TupleBaseType),
         ITuple        => typeof(ITuple),
         _             => x!.GetType()
     };
     return type;
 }
```

Cql/Cql.Comparers/KeyValuePairComparer.cs
baseTwo: The .NET Comparer wasn't capable of taking a precision. This was replaced with the ICqlComparer, it would make sense to use precision on KeyValuePair's Value property.

cc @EvanMachusak

Cql/CoreTests/Fhir/DataSourceTests.cs
baseTwo: Fixed it

ewoutkramer (approved):

PR #616, opened 2024-10-21
615 Run tests through expressions compiled by AssemblyCompiler

### LINKED ISSUE #615: Change CqlToElm unit tests to compile and run via AssemblyCompiler
Some unit tests are executing compiled lambda expressions. This should no longer be the case, as it completely misses additional steps when generating C# and then compiling that.

The unit tests must be changed to run expressions all the way through the AssemblyCompiler.

baseTwo:
Work for 
* #615 

In the CqlToElmTests, the `Base.Run(..)` method was changed to accept a library as well. This will be passed to `Extensions.RunLambda`, which takes an `AssemblyCompiler` on its first param. This new way will run tests through an AssemblyLoadContext which has all the additional fixes in the C# generation step. 

This change requires packaging services to be included returned from `Base.ServiceCollection(..)`

Note, two unit tests failed, which will be ignored and fixed in the next PR #614 . When searching for "`[Ignore("Will be fixed in PR 614")]`" these are the two tests:
* `DataSourceTests.FiltersOnSpecificProp`. I'm not sure why, but this test does throw an exception as asserted anymore. Instead it successfully retrieves an array of zero Patients. Is this the intended behavior?
* `NotEqualTest.Tuple_Equal_Tuple_Null_Equals_Null`

Also, a few other unit tests had to have they cql changed so that it can compile in the `AssemblyCompiler`. In the cql the The `null`'s were changed to  `null as Integer`:
* InTest.TestInNullBoundaries
* IntervalTest.Interval_Properly_Included_in_Interval_Null
* IntevalTest.Interval_Null_Starts_Interval

⚠️I'm having trouble running the `XmlTests` through the `AssemblyCompiler`, so these tests are still compiling the lambdas directly

### DIFF
--- a/Cql/CodeGeneration.NET/ExpressionToCSharpConverter.cs
+++ b/Cql/CodeGeneration.NET/ExpressionToCSharpConverter.cs
@@ -52,14 +52,6 @@ namespace Hl7.Cql.CodeGeneration.NET
 
             public string GetTupleMetadataPropertyName(IReadOnlyCollection<(string Name, Type Type)> tupleProperties) =>
                 _tupleMetadataBuilder.GetTupleMetadataPropertyName(tupleProperties);
-
-            public void Deconstruct(
-                out int Indent,
-                out bool UseIndent)
-            {
-                Indent = this.Indent;
-                UseIndent = this.UseIndent;
-            }
         }
 
         public Context NewContext(
--- a/Cql/CodeGeneration.NET/ExpressionToCSharpConverter.cs
+++ b/Cql/CodeGeneration.NET/ExpressionToCSharpConverter.cs
@@ -710,7 +710,14 @@ namespace Hl7.Cql.CodeGeneration.NET
                 var leftCode =  ConvertExpression(left, nextArgs);
                 leftCode = leftCode.ParenthesizeIfNeeded();
                 var rightCode = ConvertExpression(right, nextArgs);
-                var binaryString = $"{context.IndentString}{leftCode} {@operator} {rightCode}";
+                string binaryString = @operator switch
+                {
+                    // (constant value is null) --> false
+                    "is" when rightCode == "null" && left is ConstantExpression { Value: ValueType } => "false",
+                    // (null is null) --> true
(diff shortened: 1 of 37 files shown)


baseTwo:
@ewoutkramer  please check DataSourceTests.FiltersOnSpecificProp 
* #621

baseTwo:
@baseTwo create separate issue to fix XmlTests 
* #618

Cql/CodeGeneration.NET/ExpressionToCSharpConverter.cs
```diff
@@ -718,7 +710,14 @@ private string ConvertBinaryExpression(
                 var leftCode =  ConvertExpression(left, nextArgs);
                 leftCode = leftCode.ParenthesizeIfNeeded();
                 var rightCode = ConvertExpression(right, nextArgs);
-                var binaryString = $"{context.IndentString}{leftCode} {@operator} {rightCode}";
+                string binaryString = @operator switch
```
baseTwo, line 713: Create an issue to move this to a  pre-processor traverser
* #620

Cql/CqlToElmTests/Base.cs
```diff
@@ -115,29 +117,40 @@ internal static LibrarySetCSharp GenerateCSharp(LibrarySetDefinitions librarySet
...
+            Expression expression,
+            Library library,
+            CqlContext? ctx = null)
         {
             var lambda = LibraryExpressionBuilder.Lambda(expression);
-            var dg = lambda.Compile();
-            var result = dg.DynamicInvoke(ctx ?? FhirCqlContext.ForBundle());
+            var result = AssemblyCompiler.RunLambda(lambda, library, ctx);
             return result;
+            // var dg = lambda.Compile();
```
baseTwo: remove dead lines

ewoutkramer:
>@ewoutkramer please check DataSourceTests.FiltersOnSpecificProp
>https://github.com/FirelyTeam/firely-cql-sdk/issues/621

It's broken on develop-2.0, it breaks in my totally unrelated branch too 😨

ewoutkramer (approved):

Issue #634, opened 2024-11-19
Adopt standard extension for output parameters that are a list


alexzautke:
**Describe the bug**
Currently, if the output parameter of a Library is list (repeating parameter), an extension with url `https://ncqa.org/fhir/StructureDefinition/ext-parameter.list-element-type` is added on top of the corresponding parameter. It indicates the concrete type of the list. The type of the parameter is set to "List" in these cases.

The international [Using CQL with FHIR implementation guide](https://hl7.org/fhir/uv/cql/conformance.html) defines the handling of these cases differently. The type of the parameter should be the concrete type. Additionally, the http://hl7.org/fhir/StructureDefinition/cqf-cqlType should be used to indicate that it's a list. 

Note: The Java tooling uses List<T> instead of List for the cqlType. In our implementation we don't know about the type parameter, so it will just be List. Remove the NCQA extension.

Additional: Note also the `"max" : "1"` field in the before sample below, is this correct for a list? 

**Example**
Before:
![Image](https://github.com/user-attachments/assets/0af52b49-1ea4-4016-86ad-f3dd4268612c)

After:
![Image](https://github.com/user-attachments/assets/bae8f3ea-5b00-4054-bda0-f160e069d44d)

baseTwo:
@baseTwo  Reach out to Alexander to walk through the bug

baseTwo:
Meeting [video](https://fhirly-my.sharepoint.com/personal/alexander_fire_ly/_layouts/15/stream.aspx?id=%2Fpersonal%2Falexander%5Ffire%5Fly%2FDocuments%2FRecordings%2FMeeting%20with%20Alexander%20Zautke%2D20241202%5F114614%2DMeeting%20Recording%2Emp4&referrer=StreamWebApp%2EWeb&referrerScenario=AddressBarCopied%2Eview%2E71486431%2Dbbd6%2D4b47%2D9f78%2D121253b6843a&ga=1) with Alexander

PR #647, opened 2024-12-04
635 ensure that cqltype extension is added to all library input parameters

### LINKED ISSUE #635: Ensure that cqlType extension is added to all Library input parameters
**Describe the bug**
Currently the https://ncqa.org/fhir/StructureDefinition/ext-parameter.list-element-type is added also on non-repeating parameters. Make sure to add a [cqlType extension](https://hl7.org/fhir/uv/cql/conformance.html) to all **input and output** parameters of a Library.

Image

baseTwo:
Work for 
* #635 

# Work
* Add cqltype extension on all parameters (boxed in red below). Notice that the previous issue #634 added the same extension but only for list types (boxed in cyan). 
image
* Also include the element type if provided, except for List
* Split the library and measure resource packaging into separate types

### DIFF
--- a/Cql/Cql.Packaging/ResourcePackager.cs
+++ b/Cql/Cql.Packaging/ResourcePackager.cs
@@ -493,6 +493,7 @@ file static class LibraryPackager
     private static void AddParameterCqlTypeExtension(CqlTypeToFhirMapping type, ParameterDefinition parameterDefinition)
     {
         var cqlType = type.CqlType;
+        var cqlElementType = type.ElementType?.CqlType;
         switch (cqlType)
         {
             case null:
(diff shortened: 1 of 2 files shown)


alexzautke (changes requested):
Haven't reviewed the concrete code changes, however for input parameter it is necessary to the type parameter, especially for Intervals. Consumers need to know if it's an Interval<Date> or Interval<DateTime>. This should be done for lists as well, but there it's less critical.

baseTwo:
> Haven't reviewed the concrete code changes, however for input parameter it is necessary to the type parameter, especially for Intervals. Consumers need to know if it's an Interval or Interval. This should be done for lists as well, but there it's less critical.

I added the element type for "genericness". So Interval<T> and List<T> will work now where T is the Cql type

baseTwo:
I removed the "genericness" for Lists, as discussed with @alexzautke

alexzautke (approved):

PR #658, opened 2024-12-16
670 20 Begin CQL, ELM and Runtime APIs

### LINKED ISSUE #670: Public API | Begin CQL, ELM and Runtime APIs
## Target Consumers of API:

- Basic: Developers using API as-is. We will write stable versioned API for this group.
- Advanced: Developers wanting to extend the API. This group may extend functionality of the API writing extension methods on the interfaces provided under namespaces ending with ".Extensibility". (May have to pick a better name at a later stage for this). These interfaces are not considered part of the basic API and not considered stable.
- CLI (Command Line Interface) Tooling Users. Those want to generate artifacts only using the command line packaging tool. 

## APIs
Split into multiple APIs: 
- CQL API 
  - Core
    - add CQL strings
    - convert CQL to ELM
    - access to generated ELM
  - Extensions
    - add CQL from files, directories, etc.
    - save ELM to directories
    - invoke runtime definitions (via ELM, then Runtime API)
- ELM API
  - Core
    - add ELM libraries
    - convert ELM -> C# -> .NET assemblies + debug symbols -> FHIR Resources (library, measure, etc.))

      At no point between ELM and the ultimate outputs is it allowed to modify or inject the intermediate artifacts, e.g. changing the C#, or adding additional DLLs into a FHIR resource is not allowed
    - access to generated C#, assemblies, debug symbols, FHIR Resources
    - ❓ signing assemblies
    - ❓ access to warnings and errors via property
  - Extensions
    - add ELM from CQL API output
    - add ELM from files, directories, etc.
    - save C# to directories
    - save assemblies to directories
    - save debug symbols to directories
    - save FHIR Resources to directories
    - invoke runtime definitions (via Runtime API)
- Runtime API
  - extract assemblies from FHIR Resources (Extension?)
  - add assembly bytes with optional debug symbols bytes
  - provide access to assemblies
  - allow dynamically invoking definitions given a library name/version and the definition name
  - DEBUGGING SUPPORT - The ability to step into the generated c# code when the assemblies contain debug symbols

## Considerations:
- Keeping it very simple for users of the API
- API divided into basic/public/stable and advanced/"internal"/beta
- ~Error handling will be available through an Errors property instead of exceptions~
- Error handling - allow best effort processing, ignoring errors.  Each stage 
- Single-threaded, synchronous operations
- Choosing specialized types (e.g. `DirectoryInfo`, `CqlVersionedIdentifier`) over strings
- ~Immutable compilation units~
- Using files, strings and byte arrays as input and output. No streams (considered advanced)
- Only focused on translating, not on introspection of libraries
- Extending the API possible with extension methods using "advanced" interfaces
- Advanced features available through interfaces and implemented as explicit in derived types
  - Introspection of cql or elm
  - Streams?
  - Internal access to services?

baseTwo:
Work for #670 

ℹ️ For the design, and considerations, please ready the description on the issue above.
ℹ️ Since this is a huge prototype, merges will be into a new side branch `develop-2.1`  which will be merged back into `develop-2.0` later.

This PR is a prototype for the CQL public API. To jump straight into the example usage of the API, go to the `Cql\CqlSdkPrototype\App\Program.cs` file.

## Bulky Update

This PR contains many changes to the internals of the current repo, that were necessary, e.g. removing C# code generation out of `AssemblyCompiler` and shifting that solely to `CSharpCodeGenerator`, and shifting options out of the constructors and moving them as arguments to the main methods in each service. Many of the tests have been updated to use the new API.

Many types were removed in the repo, reducing complexity. 73000 lines of code removed and 7000 new lines added.

Summary of changes:
- General
  - Removed classes: *Callbacks, *PostProcessor, *ServiceCollectionExtensions, *ServiceProviderExtensions and many *Options and *Options.Validator

- Test Projects
  - Use the *Api's as much as possible instead of accessing services directly. Still much to do and clean up here

- CqlSdkPrototype project 
  - This is a new project where the new design is prototyped. The code here will be distributed amongst the existing projects before merging back into develop-2.0

- Compiler
  - LibrarySet is now internal

- CodeGeneration.NET
  - AssemblyCompiler: Simplified by removing the concern to generate source code. Only compiles assemblies. Now capable of generating debug symbols embedded into assembies, or as separate pdb byte[]
  - LibrarySetCSharpCodeGenerator: Simplified by removing callback.

- PackagerCLI
  - Almost all InternalsVisibleTo the PackagerCLI project were removed

- Test.Measures.Demo project
  - Some tests didn't pass before, they are now fixed
  - This was not a problem before since the tests were not included in the build pipeline, however, now they are

...
 
## CQL SDK Service Dependencies
The dependency injection graph shows the simpler design than before

Simpler design
```mermaid
%%{init: {
    'themeVariables':{  
      'lineColor': '#888',
      'lineWidth': 4
}}}%%

classDiagram

    direction LR

    namespace Compiler {
        class LibrarySetExpressionBuilder { }    
        class LibraryExpressionBuilder { }    
        class ExpressionBuilder { }    
        class ExpressionBuilderSettings { }    
        class TupleBuilderCache { }
        class CqlContextBinder { }
        class CqlOperatorsBinder { }       
    }

    namespace CodeGeneration {
        class TypeToCSharpConverter { }
        class CSharpCodeGenerator { }
        class AssemblyCompiler { }
    }

    namespace Packaging {
        class ResourcePackager { }
    }

    namespace Abstraction {
        class TypeResolver { }
    }

    namespace Runtime {
        class BaseTypeResolver { }
    }

    namespace Fhir {
        class FhirTypeResolver { }
        class ModelInspector { }
    }

    namespace Conversion {
        class TypeConverter { }
    }

    %% Style Scoped Types as Cyan

    style LibrarySetExpressionBuilder fill:#055
    style LibraryExpressionBuilder fill:#055
    style ExpressionBuilder fill:#055
    style TupleBuilderCache fill:#055
    
    %% Inheritance  

    BaseTypeResolver --> TypeResolver : inherits
    FhirTypeResolver --> BaseTypeResolver : inherits
    
    %% Dependencies                                                 

    LibraryExpressionBuilder ..> LibrarySetExpressionBuilder : injected

    TypeResolver ..> ExpressionBuilder : injected
    CqlOperatorsBinder ..> ExpressionBuilder : injected
    TupleBuilderCache ..> ExpressionBuilder : injected
    CqlContextBinder ..> ExpressionBuilder : injected
    ExpressionBuilderSettings ..> ExpressionBuilder : injected

    TypeResolver ..> CqlOperatorsBinder : injected
    TypeConverter ..> CqlOperatorsBinder : injected

    TypeToCSharpConverter ..> CSharpCodeGenerator : injected
    TypeResolver ..> CSharpCodeGenerator : injected

    TypeResolver ..> AssemblyCompiler : injected

    TypeResolver ..> ResourcePackager : injected
```


and-hus:
Just some general thoughts about the considerations, which I am not sure if they are implicitly covered (I haven't checked the code yet...):

I assume that the term "API" covers both: Libraries (nuget packages) to add functionality to larger projects or services and command line tools (e.g. nuget packages installed as a dotnet tools). Where the command line tool is basically a wrapper around the library. I assume that it's planned to create both - right?

Working on files is definitively the most important use case.
But what do you think about an abstraction layer to enable support for different sources or sinks than just files, e.g. a database? It might be worth the effort to add this abstraction layer early on to be prepared for different use cases than just file based actions. This might become handy if you want to used the SDK in a larger project - e.g. integrate it into a FHIR server or another service.

Nice idea to design a fluent interface - I like the approach.
It probably already serves the purpose of the abstraction layer I mentioned above...

One thought about execution of libraries. I think it would be nice to have both options:
- Execute the whole library, so run all expressions and return a dictionary with the results  
  Basically what the `Run()` method in the AssemblyLoadContextExtensions implements
- Execute just a single expression and return just this result  
  So something like `object? asm.RunExpression(library, version, context, expression)`

Cql/CqlSdkPrototype/Program.cs
```diff
@@ -0,0 +1,387 @@
...
+            var fileName = Path.Combine(directory.FullName, $"{libraryName}.dll");
+            File.WriteAllBytes(fileName, assemblyBinary);
+            Console.WriteLine($"Saved assembly binary to file: {fileName}");
+        }
+        return this;
+    }
+}
+
+
+public readonly record struct ElmVersionedIdentifier
```
and-hus: I think this struct is generic enough to be used in other contexts then just for the Elm, e.g. for the FHIR library resources etc. So what do you think about renaming this to just `VersionedIdentifier`?

Cql/CqlSdkPrototype/Program.cs
```diff
@@ -0,0 +1,387 @@
...
+
+    public string? Version { get; init; }
+
+    public void Deconstruct(out string name, out string? version)
+    {
+        name = Name;
+        version = Version;
+    }
+
+    public static ElmVersionedIdentifier ParseString(string text)
```
and-hus: To be consistend with `FromNameAndVersion()` and `FromVersionedIdentifier()` I would propose to rename this method to `FromString()` - what do you think?

and-hus:
Regarding the API for the ElmCompiler, what do you think about using the builder pattern, so something along these lines:

```csharp
var elmCompiler = 
          ElmCompilerBuilder
                .New
                .WithInputFile(elmInDir, ElmVersionedIdentifier.FromNameAndVersion("FHIRHelpers"))
                .WithInputFolder(elmInDir, new EnumerationOptions() { RecurseSubdirectories = false })
                .WithCSharpOutputFolder(csharpOutDir)
                .WithAssemblyOutputFolder(dllOutDir)
                .Build();
                
elmCompiler.Compile();
```

Cql/CqlSdkPrototype/Program.cs
baseTwo: There is already a VersionedIdentifier as part of the ELM xml types. I didn't want to change it just yet for the prototype, might do that later

Cql/CqlSdkPrototype/Program.cs
baseTwo: I changed this to just Parse

baseTwo:
> Regarding the API for the ElmCompiler, what do you think about using the builder pattern, so something along these lines:
> 
> ```cs
> var elmCompiler = 
>           ElmCompilerBuilder
>                 .New
>                 .WithInputFile(elmInDir, ElmVersionedIdentifier.FromNameAndVersion("FHIRHelpers"))
>                 .WithInputFolder(elmInDir, new EnumerationOptions() { RecurseSubdirectories = false })
>                 .WithCSharpOutputFolder(csharpOutDir)
>                 .WithAssemblyOutputFolder(dllOutDir)
>                 .Build();
>                 
> elmCompiler.Compile();
> ```

I have considered a builder pattern, but I really want to be able to modify the "ElmCompilation" even after compilation. (Think about running this for the Visual Studio plugin which does incremental builds, adding/removing files even after a compile).

We want to be able to add/remove files, compile, and repeat

baseTwo:
Since this PR size is getting big, I'm splitting up the work into more sub tasks

Cql/CqlSdkPrototype/Cql/CqlApi.cs
```diff
@@ -0,0 +1,159 @@
...
+using CqlTranslationEntriesMap = System.Collections.Immutable.ImmutableDictionary<
+    CqlVersionedLibraryIdentifier,
+    CqlTranslationEntry>;
+
+public class CqlApi(CqlApiOptions options) :
+    ICqlApiExtensible<CqlApi>
+{
+    internal ICqlApiExtensible<CqlApi> AsExtensible() => this;
+    TResult ICqlApiExtensible<CqlApi>.UseLogger<TResult>(Func<CqlApi, ILogger<CqlApi>, TResult> action) => action(this, _state.Logger);
+    public static CqlApi Create(CqlApiOptions? options = null) => new(options ?? CqlApiOptions.Default);
```
and-hus: The class has both, a public primary constructor and this static factory method. 
So there are basically two ways to create an instance of this class - is this intentionally? 
To have just one controlled way for instantiation, I would propose to make the constructor private which leaves the static factory method as the only way to create an instance. What do you think?

Cql/CqlSdkPrototype/App/Program.cs
```diff
@@ -0,0 +1,265 @@
...
+namespace CqlSdkPrototype.App;
+
+internal class Program
+{
+    static void Main(string[] args)
+    {
+        // CONSIDERATIONS:
+        // - Encapsulate the creation of the service provider
+        // - Move any options such as the AssembliesDebugMode and Models to the Cql- or ElmApi instead of via the hosted IOptions<T>
+        // - Should we keep the dependency on Microsoft's ILogger or should we create our own logging abstraction?
```
and-hus: Out of curiosity: 
What's the problem with Microsoft's ILogger<> ? 
Why are you considering to create your own logging abstraction? What are you trying to achieve that is not possible with ILogger?

baseTwo: I use Microsoft's logging framework in the current PR.

baseTwo: I should move out my thoughts from the code though,. They're all in the parent issue's description  #670

baseTwo: I am using ILogProvider[] in the LoggingOptions, but I think it should be changed to ILoggerFactory instead.

and-hus: I agree - I am also more used to `ILogger<>` or `ILoggerFactory`...

From the perspective of a user/consumer of the API, I would expect the API to accept either a `ILogger<TCategoryName>` parameter or an `ILoggerFactory`. So something like

```csharp
 public class CqlApi(){

    public static CqlApi Create(ILogger<CqlApi> logger = null);
 }
```

or

```csharp
 public class CqlApi(){

    public static CqlApi Create(ILoggerFactory loggerFactory = null);
 }
```

See e.g. [here](https://learn.microsoft.com/en-us/dotnet/core/extensions/logging-library-authors) for some guidance on when to use which approach.

But I assume that I'm just not getting the point why the current setup is required  ;-)

baseTwo: The API's create their own ServiceProvider which will create ILogger for all the internal services required including the API's. When setting up the ServiceProvider, it needs to know how to set up the LoggingBuilder. This is where the LoggingOptions comes in.

baseTwo: I'm looking now if it's possible to only use ILoggerFactory in LoggingOptions, instead of passing in ILoggerProvider[] and the log filtering delegate.

and-hus: Yes - I think that would be good and would allow the consumers of the API to use standard Microsoft Logging setups or 3rd party log providers like Serilog which typically provide dependency injection for `ILogger<>` or `ILoggerFactory` - I am not sure how this would work with `ILoggerProvider[]` ...

baseTwo: Ok, ILoggerFactory was the correct choice. Please see changes in the commit.

baseTwo: Setting up up logging now looks like this

```csharp
        // This would be set up in the consumers startup logic
        using var serviceProvider =
            new ServiceCollection()
                .AddLogging(lb => lb
                  .ClearProviders()
                  .AddProvider(new ColorConsoleLoggerProvider())
                  .AddFilter((category, _) => category?.Contains(value: nameof(CqlSdkPrototype)) ?? false))
                .BuildServiceProvider();

        // ILoggerFactory would be available through constructor dependency injection, and not directly from IServiceProvider
        var loggingOptions = new LoggingOptions(LoggerFactory: serviceProvider.GetRequiredService<ILoggerFactory>());

        var cqlApiOptions = new CqlApiOptions(
            LoggingOptions: loggingOptions,
            Models: [Models.ElmR1, Models.Fhir401]);

        var cqlApi = new CqlApi(cqlApiOptions);

```

Cql/CoreTests/PrimitiveTests.cs
```diff
@@ -3405,29 +3407,18 @@ public void Collapse_Containing_Null()
...
+
+            var loggerFactory = new ServiceCollection()
+                                .AddDebugLogging()
+                                .BuildServiceProvider()
+                                .GetRequiredService<ILoggerFactory>();
+            new ElmApi(loggerFactory)
+                .AddElmLibraries(librarySet)
+                .Compile();
+
+            Assert.IsTrue(true);
```
ewoutkramer, line 3421: Does this test that the Compile() does not throw? Can we somehow make that more explicit (maybe FluentAssertions DoesNotThrow?)

Cql/Cql.Abstractions/ISingleton.cs
```diff
@@ -13,4 +13,15 @@ public interface ISingleton<out TSelf>
     /// Gets the instance of the implementing class.
     /// </summary>
     static virtual TSelf Instance { get; } = Activator.CreateInstance<TSelf>();
+}
+
+/// <summary>
+/// Represents an interface for a singleton pattern.
+/// </summary>
+public interface ISingleton
```
ewoutkramer: Is this useful?

Cql/Cql.CqlToElm/Visitors/TimingExpressionVisitor.cs
```diff
@@ -13,6 +13,8 @@ namespace Hl7.Cql.CqlToElm.Visitors
 {
     internal partial class ExpressionVisitor
     {
+        internal static bool EnableDebugAssertions = true;
```
ewoutkramer, line 16: Why is this on TimingExpressionVisitor specifically?

Cql/Cql.Packaging/FhirResourceWriterOptions.cs
```diff
@@ -1,13 +1,13 @@
-﻿/*
+﻿/*/*
  * Copyright (c) 2024, NCQA and contributors
  * See the file CONTRIBUTORS for details.
  *
  * This file is licensed under the BSD 3-Clause license
  * available at https://raw.githubusercontent.com/FirelyTeam/firely-cql-sdk/main/LICENSE
- */
+ #1#
```
ewoutkramer, line 7: Would you not want to delete this file instead of commenting it out completely?

Cql/Cql.Runtime/EmptyDisposable.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Hl7.Cql.Runtime;
+
+internal readonly record struct EmptyDisposable : IDisposable
```
ewoutkramer: Curious to learn later what you are using this for....

Cql/CqlSdkPrototype/Cql/CqlApi.cs
```diff
@@ -0,0 +1,135 @@
...
+using CqlSdkPrototype.Internal;
+using Hl7.Cql.CqlToElm;
+using Hl7.Cql.Runtime;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace CqlSdkPrototype.Cql;
+
+using CqlApiStateEntryDictionary = ImmutableDictionary<CqlVersionedLibraryIdentifier, CqlApiStateEntry>;
+
+public class CqlApi :
```
ewoutkramer, line 11: I am not in love with the name "CqlApi".  Just this morning I confused it with "a library to work with Cql", but it is instead a library to get from Cql to Elm, and the name kind of does not show that. Also, mentioning the term "Api" explicitly feels superfluous. Would CqlToElm or CqlCompiler (well, I guess we have that in the lower-level API already) be something?  CqlConverter, CqlTransformer, CqlProcessor..... CqlSomething ;-)

Cql/CqlSdkPrototype/Cql/CqlApi.cs
```diff
@@ -0,0 +1,135 @@
...
+    internal CqlApi(CqlApiState state) => _state = state; // Might make this public later. Used for testing now.
+
+    public CqlApi(
+        ILoggerFactory? loggerFactory = null,
+        CqlApiOptions? options = null) : this(CqlApiState.Create(loggerFactory ?? NullLoggerFactory.Instance, options ?? CqlApiOptions.Default))
+    {}
+
+    #region State
+
+    private CqlApiState _state;
```
ewoutkramer: I know you initially had CqlApi be read-only, and now it's a big state object in CqlApi, that's almost 1-on-1 with the CqlApi. What was the advantage again of keeping the CqlApi mutable, but the state inside immutable?

Cql/CqlSdkPrototype/Cql/CqlApiState.cs
```diff
@@ -0,0 +1,142 @@
...
+    {
+        get => _options;
+        init
+        {
+            if (ReferenceEquals(_options, value))
+                return;
+
+            ServiceProvider?.Dispose();
+
+            _options = value;
```
ewoutkramer: With this many side-effects, I'd lean towards `SetOptions()`. Which you probably avoided to be able to use `with`?

Cql/CqlSdkPrototype/Cql/CqlApiState.cs
```diff
@@ -0,0 +1,142 @@
...
+                foreach (var modelInfo in modelInfos)
+                    modelProvider.Add(modelInfo);
+            };
+        }
+    }
+
+    private static void SuppressCqlDebugAssertions()
+    {
+        // This is really annoying in debug mode
+        ExpressionVisitor.EnableDebugAssertions = false;
```
ewoutkramer: I think these were left-over debug statements that we might be able to just remove. But, yes, we'd need to double check with Evan.

Cql/CqlSdkPrototype/Cql/CqlApiState.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        if (EntriesBuilder.TryGetValue(libVer, out var entry))
+        {
+            if (entry.ElmLibraryBuilder is {} lb)
+            {
+                libraryBuilder = lb;
+                return true;
+            }
+
+            ;
```
ewoutkramer: I've seen a few empty colons like these, is that on purpose?

Cql/CqlSdkPrototype/Cql/Extensibility/CqlApiStateEntry.cs
```diff
@@ -0,0 +1,9 @@
+﻿using CqlSdkPrototype.Infrastructure;
+using Hl7.Cql.CqlToElm;
+using Hl7.Cql.Elm;
+
+namespace CqlSdkPrototype.Cql.Extensibility;
+
+public readonly record struct CqlApiStateEntry(CqlLibraryString CqlLibraryString, Library? ElmLibrary = null) {
```
ewoutkramer, line 7: CqlApiState is internal, does this need to be public?

ewoutkramer: If this is a public facing struct, then we should give it a sexier name, something that says what it is, as the current name is too abstract to tell the user about its function.

Cql/CqlSdkPrototype/Cql/Internal/ICqlApiInternal.cs
```diff
@@ -0,0 +1,9 @@
+﻿using CqlSdkPrototype.Cql.Extensibility;
+
+namespace CqlSdkPrototype.Cql.Internal;
+
+internal interface ICqlApiInternal<TCqlApi> : ICqlApiExtendable<TCqlApi>
```
ewoutkramer, line 8: Maybe explain why there is an internal version with a few lines of comments.

Cql/CqlSdkPrototype/Elm/ElmApi.cs
```diff
@@ -0,0 +1,194 @@
...
+
+namespace CqlSdkPrototype.Elm;
+
+using ElmApiStateEntryDictionary = ImmutableDictionary<CqlVersionedLibraryIdentifier, ElmApiStateEntry>;
+
+public class ElmApi :
+    IElmApiExtendable<ElmApi>,
+    IElmApiInternal<ElmApi>
+{
+    public ElmApi(
```
ewoutkramer, line 18: Same mixed feelings as for "CqlApi".

Cql/CqlSdkPrototype/Elm/ElmApi.cs
```diff
@@ -0,0 +1,194 @@
...
+        return hasChanged
+                   ? WithEntries(entries: entries.ToImmutable())
+                   : this;
+    }
+
+    #endregion
+
+    #region Processing
+
+    public ElmApi Compile()
```
ewoutkramer, line 85: That's a big function!

Cql/CqlSdkPrototype/Elm/ElmApiScopedState.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Hl7.Cql.Compiler;
+
+namespace CqlSdkPrototype.Elm;
+
+internal class ElmApiScopedState(IServiceScope scope) : IDisposable
```
ewoutkramer, line 5: What are we using scoped state for?

Cql/CqlSdkPrototype/Elm/ElmApiState.cs
```diff
@@ -0,0 +1,100 @@
...
+        init
+        {
+            if (ReferenceEquals(_options, value))
+                return;
+
+            ServiceProvider?.Dispose();
+
+            _options = value;
+
+            var services = new ServiceCollection();
```
ewoutkramer, line 50: See CqlApiOptions.Options

Cql/CqlSdkPrototype/Elm/Extensions/ElmApiExtensions.Cql.cs
```diff
@@ -0,0 +1,17 @@
...
+using CqlSdkPrototype.Runtime.Extensions;
+
+namespace CqlSdkPrototype.Elm.Extensions;
+
+public static partial class ElmApiExtensions
+{
+    public static ElmApi Compile(
+        this CqlApi cqlApi)
+    {
+        return cqlApi
```
ewoutkramer, line 11: Nice!

Cql/CqlSdkPrototype/GlobalUsings.cs
```diff
@@ -0,0 +1,10 @@
+// Global using directives
+
+global using System.Collections.Immutable;
```
ewoutkramer, line 3: Yeah, add System!

Cql/CqlSdkPrototype/Infrastructure/CqlLibraryString.cs
```diff
@@ -0,0 +1,58 @@
...
+    public static CqlLibraryString Empty { get; } = new(CqlVersionedLibraryIdentifier.Empty, string.Empty);
+
+    public static CqlLibraryString FromIdentifierAndString(
+        CqlVersionedLibraryIdentifier versionedLibraryIdentifier,
+        string cqlContent)
+    {
+        return new CqlLibraryString(versionedLibraryIdentifier, cqlContent);
+    }
+
+    [GeneratedRegex("""
```
ewoutkramer, line 21: Good to have a comment with a reference to where this regex can be found in the CQL documentation here

Cql/CqlSdkPrototype/Internal/Directories.cs
```diff
@@ -0,0 +1,56 @@
...
+            current = current.Parent;
+        }
+
+        throw new InvalidOperationException("Could not find ancestor directory from search criteria.");
+    }
+
+    public static DirectoryInfo CurrentDirectory { get; } = new(Directory.GetCurrentDirectory());
+    public static DirectoryInfo CurrentSolutionDirectory { get; } = FindAncestor(CurrentDirectory, d => d.EnumerateFiles("*.sln?").Any());
+    public static DirectoryInfo CurrentProjectDirectory { get; } = FindAncestor(CurrentDirectory, d => d.EnumerateFiles("*.csproj").Any());
+    public static DirectoryInfo AuthoringProjectDirectory => CurrentSolutionDirectory.CreateSubdirectory("Demo").CreateSubdirectory("Measures.Authoring");
```
ewoutkramer, line 32: It's a bit too much side effect for my taste to create a directory when reading a property, can this property be named GetAuthoringProjectDirectory()?  This applies to the other props here as well of course!

Cql/CqlSdkPrototype/Internal/InternalExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    public static Assembly LoadFromBytes(this AssemblyLoadContext assemblyLoadContext, byte[] assembly, byte[]? symbols = null)
+    {
+        using var assemblyStream = new MemoryStream(assembly);
+        var symbolsStream = symbols is {Length:>0} ? new MemoryStream(symbols) : null;
+        using var symbolsStreamDisposable = symbolsStream as IDisposable ?? new EmptyDisposable();
+        return assemblyLoadContext.LoadFromStream(assemblyStream, symbolsStream);
+    }
+
+    public static IEnumerable<TR> SelectWhereNotNull<T, TR>(
```
ewoutkramer, line 33: Would it be far-fetched to have a company-wide library with this stuff? I need it too.  You have done this in your previous job, right?

Cql/CqlSdkPrototype/Internal/Maybe.cs
```diff
@@ -0,0 +1,77 @@
+﻿namespace CqlSdkPrototype.Internal;
+
+internal readonly record struct Maybe
+{
+    public static Maybe NoValue => default;
+    public static Maybe<T> Of<T>(T value) => new(value);
+}
+
+internal readonly record struct Maybe<T>(T Value)
```
ewoutkramer, line 9: I have just removed a similar `Result<T>` from the FHIR Plugin. I love the concept, but it's just not common C# enough to have it in the public interface. And I think it will remain a curiosity within your code. Alas, I'd love to use it.

Cql/CqlSdkPrototype/Logging/ColorConsoleLogger.cs
```diff
@@ -0,0 +1,184 @@
+﻿using Microsoft.Extensions.Logging.Abstractions;
+using System.Collections.Concurrent;
+using System.Globalization;
+using System.Text.RegularExpressions;
+using EnumerableExtensions = Hl7.Cql.Abstractions.Infrastructure.EnumerableExtensions;
+
+namespace CqlSdkPrototype.Logging;
+
+public sealed partial class ColorConsoleLogger(string categoryName, ColorConsoleLoggerProvider provider) : ILogger
```
ewoutkramer, line 9: Does this need to be public?

Cql/CqlSdkPrototype/Runtime/Extensions/RuntimeApiExtensions.Cql.cs
```diff
@@ -0,0 +1,33 @@
...
+        where TCqlApi : ICqlApiExtendable<TCqlApi>
+    {
+        var elmApiOptions = new ElmApiOptions(ProcessBatchItemExceptionHandling: cqlApi.Options.ProcessBatchItemExceptionHandling);
+        if (configureOptions is not null) elmApiOptions = configureOptions(elmApiOptions);
+        var elmApi = new ElmApi(cqlApi.LoggerFactory, elmApiOptions).AddElmFromCqlApi(cqlApi);
+        return elmApi;
+    }
+
+#pragma warning disable RS0026
+    public static RuntimeScope CreateRuntimeScope(
```
ewoutkramer, line 22: As a "new" user, the first question that comes up is: "What's a RuntimeScope and why do I need it?".  When would I call CreateRuntimeScope?

Cql/CqlSdkPrototype/Runtime/Extensions/RuntimeApiExtensions.RuntimeScope.cs
```diff
@@ -0,0 +1,79 @@
...
+
+namespace CqlSdkPrototype.Runtime.Extensions;
+
+public static partial class RuntimeApiExtensions
+{
+    public static IEnumerable<(CqlVersionedLibraryIdentifier library, string declarationName, Func<object?> getResult)> EnumerateLibrarySetDefinitionsResults(
+        this RuntimeScope scope,
+        CqlContext cqlContext)
+    {
+        foreach (var (libId, lib) in scope.Libraries)
```
ewoutkramer, line 15: Shouldn't this function call `EnumerateLibraryDefinitionsResults` in the foreach, instead of duplicating its code?

Cql/CqlSdkPrototype/Runtime/Invokers/LibraryInvoker.2_0_8_0.cs
```diff
@@ -0,0 +1,81 @@
...
+            return false;
+        }
+
+        libraryInvoker = new LibraryInvoker_2_0_8_0(asILibrary);
+        return true;
+    }
+
+    public static bool SupportsVersion(Version cqlToolVersion)
+    {
+        return cqlToolVersion.IsBetween(new Version(2, 0, 8), new Version(2, 1));
```
ewoutkramer: Do we make a newer version of this class when the generated libraries change, and in the mean time keep updating the upper boundary for every new release?

Cql/PackagerCLI/DependencyInjection/PackagerCLiLoggingBuilderExtensions.cs
```diff
@@ -80,6 +75,6 @@ public static ILoggingBuilder AddPackagerCLiLogging(
             /* 6 */
             LogLevel.None => /* n/a */ null,
             // @formatter: on
-            _ => throw new UnsupportedSwitchCaseError(logLevel, typeof(LogLevel).FullName).ToException(),
+            _ => throw new ArgumentOutOfRangeException(nameof(logLevel), logLevel, "Invalid LogLevel."),
```
ewoutkramer: Or use the new standard .NET SwitchExpressionException?

Demo/Test.Measures.Demo/MeasuresTest.cs
```diff
@@ -102,70 +106,63 @@ public void BCSEHEDIS2022_Numerator_FromElm()
...
+            CqlContext context)
+        {
+            return scope
+                   .EnumerateLibraryDefinitionsResults(context, CqlVersionedLibraryIdentifier.Parse($"{lib}-{version}"))
+                   .ToDictionary(t => t.definition, t => t.getResult());
+        }
+
         [UsedImplicitly]
-        public static AssemblyLoadContext LoadResources(
+        public static RuntimeScope CreateRuntimeScopeFromFhirResourceFile(
```
ewoutkramer, line 120: What did the RuntimeScope add to AssemblyLoadContext that requires this extra concept of a scope?  (sorry, it's hard to look back in such a big PR into files that I have already reviewed....)

Demo/Test.Measures.Demo/TestBase.cs
```diff
@@ -1,51 +1,4 @@
...
-                    valueSets.AddRange(bundle.Entry.Select(e => e.Resource).OfType<ValueSet>());
-                else if (resource is ValueSet valueSet)
-                    valueSets.Add(valueSet);
-            }
-            var vsd = valueSets.ToValueSetDictionary(false);
-            return vsd;
-        }
- 
-    }
+    public abstract class TestBase;
```
ewoutkramer: Just there to avoid compilation errors?

Cql/CqlSdkPrototype/Cql/Extensibility/ICqlApiExtendable.cs
```diff
@@ -0,0 +1,14 @@
+﻿using CqlSdkPrototype.Infrastructure;
+
+namespace CqlSdkPrototype.Cql.Extensibility;
+
+public interface ICqlApiExtendable<TCqlApi>
```
ewoutkramer, line 5: What's the advantage of doing this over just writing extensions again the CqlApi class?

Cql/CqlSdkPrototype/Cql/CqlApi.cs
ewoutkramer, line 11: Now that the SDK comes closer to being finzalized, we should enable the check again that all public members need documentation, since there's tons of public members without it now!

ewoutkramer (changes requested):

Cql/CoreTests/PrimitiveTests.cs
baseTwo: It does throw by default. The options has a flag to set the exception handling. We can write an extension method to make it easier to change, but these convenience methods can come later. Issue #697

Cql/Cql.Abstractions/ISingleton.cs
baseTwo: I'll remove the non-generic one, since it's not being used

Cql/Cql.CqlToElm/Visitors/TimingExpressionVisitor.cs
baseTwo: There are two classes in the cql to elm package that has many debug assertions, of which this type is one. Not sure why asserts were used in the original PR, but i needed a quick way to disable them globally without changing existing behaviour. I suppose these assertions need to be cleaned up. We have to speak to @EvanMachusak about it. I added it as an item to #674

Cql/Cql.Runtime/EmptyDisposable.cs
baseTwo: Example usage below, shows the use of an IDisposable that might be null based on a condition, but that would not be allowed in a using inline scope. 

```csharp
 using var codeStream = new MemoryStream();
 MemoryStream? pdbStream = debugInformationFormat == AssemblyCompilerDebugInformationFormat.PortablePdb ? new MemoryStream() : null;
 using var pdbStreamDisposable = pdbStream as IDisposable ?? new EmptyDisposable();
```

Demo/Test.Measures.Demo/MeasuresTest.cs
baseTwo: RuntimeScope encapsulates AssemblyLoadContext, which is disposable, so the RuntimeScope must be disposed when invocation is completed on it, so that the assemblies may unload..

Demo/Test.Measures.Demo/TestBase.cs
baseTwo: I'll remove the base type, since it no longer contains any class members, and there is only one test that derived from it

Cql/CqlSdkPrototype/Cql/CqlApi.cs
baseTwo: - #675

Cql/CqlSdkPrototype/Cql/Extensibility/ICqlApiExtendable.cs
baseTwo: At the moment no advantage, because there is no derived type on CqlApi. I have already thought of another way to do the fluent style as discussed. Issue #697

Cql/Cql.Packaging/FhirResourceWriterOptions.cs
baseTwo: I want to keep this logic readily available when I'm writing the PackagerAPI. After that I'll deleted it before all goes back into the main branch

Cql/CqlSdkPrototype/Cql/CqlApi.cs
baseTwo: I agree, but I couldn't come up with a good name yet. 

 My thoughts around keeping the name general is that, the instance CqlApi or ElmApi serves as the starting point of what you want to do, not what you want to do. That's because there are many extension methods on it that makes it convenient to go through all the stages of the pipeline and produce output. 
 
 Ideas?

Cql/CqlSdkPrototype/Cql/CqlApi.cs
baseTwo: In functional programming, the data object should only contain the data, and external to it, a function will create a new state object. In this case, those functions are in the API type, and the extension methods to the API type

Cql/CqlSdkPrototype/Cql/CqlApiState.cs
baseTwo: To be cleaned up in  #674

Cql/CqlSdkPrototype/Cql/CqlApiState.cs
baseTwo: I forgot to take it out, I was stepping through the code an needed a breakpoint before leaving the scope

Cql/CqlSdkPrototype/Elm/ElmApiScopedState.cs
baseTwo: Encapsulates the service provider scope and to get access to the scoped services

baseTwo: I want to remove all scoped services during cleanup #674

Cql/CqlSdkPrototype/Elm/Extensions/ElmApiExtensions.Cql.cs
baseTwo: That's why the name CqlApi or ElmApi needs to have a general name that it is the starting point of the pipeline.

Cql/CqlSdkPrototype/GlobalUsings.cs
baseTwo: I'll add all the System and Microsoft namespaces there as a cleanup task #674

Cql/CqlSdkPrototype/Internal/Directories.cs
baseTwo: This is only a temporary class used in the cql prototype, not meant for production.

Cql/CqlSdkPrototype/Internal/InternalExtensions.cs
baseTwo: These extensions have a tendency to explode in size, bloating the shared libraries. Speaking from experience from previous job :)

Cql/CqlSdkPrototype/Internal/Maybe.cs
baseTwo: I use it in the Firely CQL SDK too. I hope Microsoft brings a more formal set of monad types into the framework

Cql/CqlSdkPrototype/Logging/ColorConsoleLogger.cs
baseTwo: This will be moved to the PackagerCLI project after the CqlSdkPrototype is removed. For now its used in both

Cql/CqlSdkPrototype/Runtime/Invokers/LibraryInvoker.2_0_8_0.cs
baseTwo: Actually, there should be no upper boundary. When developing a new library format, and the unit tests break, a new invoker must be written on the new version.

Cql/Cql.Runtime/EmptyDisposable.cs
baseTwo: I'll remove this, I thought it wasn't possible to do a `using var disposable = a null IDisposable instance`, but that is indeed possible

Cql/CqlSdkPrototype/Elm/ElmApi.cs
baseTwo: I split it up into smaller ones

Cql/CoreTests/PrimitiveTests.cs
ewoutkramer: I mean, Assert.IsTrue(true) is obviously always passing, so what is the test testing? Could you add something about the expectation in a comment above this assert.true(true), so later developers understand what this is testing?

Cql/CqlSdkPrototype/Cql/CqlApi.cs
ewoutkramer: Is there any distinction left between CqlApi and its state now? Is it worth keeping two objects?

Cql/CqlSdkPrototype/Elm/ElmApi.cs
ewoutkramer: I'll close this discussion. If we find something more appropriate for CqlApi, we can apply that here too.

Cql/CqlSdkPrototype/Internal/InternalExtensions.cs
ewoutkramer: I'm just jealous I guess. I am afraid you are right based on experiences we had at Furore. I had slight hopes you fared better ;-)

Cql/CqlSdkPrototype/Internal/Maybe.cs
ewoutkramer: Yeah, but until that moment, I don't think we should introduce a variant in our pubic surface and stick to the current Microsoft designs.

Cql/CqlSdkPrototype/Runtime/Extensions/RuntimeApiExtensions.Cql.cs
ewoutkramer: So, you want to remove all scopes in a future PR, closing this discussion.

Cql/CqlSdkPrototype/Runtime/Invokers/LibraryInvoker.2_0_8_0.cs
ewoutkramer: You mean to say that the IsBeween (with upper boundary Version(2,1)) should be removed from this code?

Demo/Test.Measures.Demo/MeasuresTest.cs
ewoutkramer: Yes, so they are both disposable, but what does the wrapper introduce that the AssemblyLoadContext does not have?

Cql/CqlSdkPrototype/Runtime/Invokers/LibraryInvoker.2_0_8_0.cs
baseTwo: Yes, the upper bound is removed now. If we follow semantic versioning on the code generated part, then it would be easy to put an upper limit on the version major part, but the versioning of the code generated libraries is tied to the SDK version itself. That relationship needs to be broken up

Cql/CqlSdkPrototype/Cql/CqlApiState.cs
baseTwo: Have a look at the latest commit. I split up options, entries and services as separate fields on the cqlapi. I'll do elmapi later

Cql/CoreTests/PrimitiveTests.cs
baseTwo: Improve the assertion

Cql/CqlSdkPrototype/Internal/Maybe.cs
baseTwo: Internal, wont go public

Demo/Test.Measures.Demo/MeasuresTest.cs
baseTwo: Will rename to LibrarySetInvoker etc #697

Cql/CqlSdkPrototype/Runtime/Invokers/LibraryInvoker.2_0_8_0.cs
baseTwo: - #698

ewoutkramer (approved):
All open issues discussed, many will be handled in another PR to not make merging harder.

Cql/CoreTests/PrimitiveTests.cs
baseTwo: Fixed in next PR #696

PR #669, opened 2025-01-13
668 C# Generator Changes

### LINKED ISSUE #668: Public API | C# Code Generator Changes
Implement ILibrary as explicit in the C# generated code - so that they don't conflict with potential generated members
Remove the UTF8 Byte Order Marker (UTF-8 uses 1-byte code units, so there is no need for a BOM to indicate a byte order, because there is only 1 byte order possible, and the encoding algorithm determines the ordering of the bytes.)
Fix generated tuple properties so that only those required by the library are generated
Code generating must be able dis/enabled via MSBuild variables

baseTwo:
Work for #668 

C# Generated Code
1. Implement `ILibrary` as explicit in the C# generated code - so that they don't conflict with potential generated members
2. Place regions around code blocks
3. Remove the UTF8 Byte Order Marker (UTF-8 uses 1-byte code units, so there is no need for a BOM to indicate a byte order, because there is only 1 byte order possible, and the encoding algorithm determines the ordering of the bytes.)
4. Fix generated tuple properties (See below `TupleMetadataBuilder`)

**C# Generator**
1. Remove the exception handling callback
2. Remove the callback for converting library name to class name. Use the `VariableNameGenerator` directly
3. Rename `LibrarySetDefinitionsToCSharpCodeProcessor` to `LibrarySetCSharpCodeGenerator`
4. Rename `LibraryDefinitionToCSharpCodeProcessor` to `LibraryDefinitionCSharpCodeGenerator`
5. Rename `IndentedWriter` to `IndentedTextWriter`
6. `IndentedTextWriter` to use the base type `TextWriter` instead of `StreamWriter`
7. Upgrade version to 2.0.10

**TupleMetadataBuilder**
1. The tuple metadata properties must be sorted alphabetically, to reduce changes when comiting to the repo
2. Only generate the necessary tuple metadata properties that are needed for the particular library

**MSBuild**
1. Remove the `Generate-CQL-to-ELM` and `FullDebug` build targets
2. Add MSBuild variables in the props files for enabling building ELM (CqlToElmEnabled) or C# (ElmToCSharpEnabled)
3. Update `.Targets.xml` files conditions for these new variables


Demo/cql-demo.props
```diff
@@ -32,11 +32,14 @@
...
 		<CqlSolutionDir>$(MSBuildThisFileDirectory)/../</CqlSolutionDir>
+		<!-- CqlToElmEnabled and ElmToCSharpEnabled declared in BuildArtifacts.csproj, cql-base.props, cql-demo.props  -->
+		<CqlToElmEnabled>false</CqlToElmEnabled>
+		<ElmToCSharpEnabled>false</ElmToCSharpEnabled>
 	</PropertyGroup>
 
 	<!-- ignore sdk warnings -->
 	<PropertyGroup>
-		<NoWarn>CS1591</NoWarn>
+		<NoWarn>CS1591,CS0618</NoWarn>
```
ewoutkramer:
```suggestion
		<NoWarn>CS1591</NoWarn>
```

cql-sdk.props
```diff
@@ -34,10 +34,9 @@
 	<PropertyGroup>
 		<LangVersion>12</LangVersion>
 		<GenerateDocumentationFile>True</GenerateDocumentationFile>
-		<Configurations>Debug;Release;FullDebug</Configurations>
 		<AnalysisLevelGlobalization>latest-recommended</AnalysisLevelGlobalization>
 		<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-		<NoWarn>RS0026</NoWarn>  <!-- Too many bogus warnings, see https://github.com/dotnet/roslyn-analyzers/issues/6757 -->
+		<NoWarn>RS0026,CS0618</NoWarn>  <!-- Too many bogus warnings, see https://github.com/dotnet/roslyn-analyzers/issues/6757 -->
```
ewoutkramer:
```suggestion
		<NoWarn>RS0026</NoWarn>  <!-- Too many bogus warnings, see https://github.com/dotnet/roslyn-analyzers/issues/6757 -->
```

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #676, opened 2025-01-15
ignore fluent assertions version 8.0 and higher in dependabot

mmsmits:
Decided to ignore moving to 8.0.0 of FluentAssertions for now since they have updated their license to a commercial one:
https://github.com/fluentassertions/fluentassertions/pull/2943


baseTwo (commented):
Could you provide context in the PR description what the purpose is for this change?

baseTwo (approved):

PR #696, opened 2025-01-20
695 25 use cqlelm apis in cqltoelmtests

### LINKED ISSUE #695: Public API | Use CQL/ELM APIs in CqlToElmTests

baseTwo:
Work for 
- #695

## Work
- API
  - Copy more switches from `CqlToElmOptions` to `CqlApiOptions`
  - Remove Empty object pattern from CqlLibraryIdentifier, CqlLibraryVersion and CqlVersionedLibraryIdentifer
  - Comment out FileSystemLibraryProvider for now, it will be removed in the future, but I want to keep the logic there for the future replacement 
- CqlToElmTests
  - Use CQL and ELM API's in CqlToElm tests
  - Remove the need for ClassInitialize in each test
  - Remove static ServiceProvider and Service instances on Base test
  - Use raw strings for cql library texts


Cql/CoreTests/Infrastructure/CqlLibraryStringTests.cs
```diff
@@ -46,8 +58,8 @@ public void FromCql_ValidCqlContent_ReturnsCqlLibraryString()
         }
 
         [TestMethod]
-        [ExpectedException(typeof(FormatException))]
-        public void FromCql_InvalidCqlContent_ThrowsFormatException()
+        [ExpectedException(typeof(ArgumentException))]
+        public void FromCql_InvalidCqlContent_ThrowsArgumentException()
```
ewoutkramer: FormatException would be in line with what other parsers do in the SDK. Why is this an argument exception?

Cql/CqlSdkPrototype/Cql/CqlApiState.cs
```diff
@@ -124,7 +124,7 @@ bool ILibraryProvider.TryResolveLibrary(
 
         var libVer = CqlVersionedLibraryIdentifier.FromNameAndVersion(
             CqlLibraryIdentifier.Parse(libraryName),
-            CqlLibraryVersion.Parse(version ?? throw new ArgumentNullException(nameof(version))));
+            CqlLibraryVersion.ParseOrEmpty(version));
```
ewoutkramer: I cannot see what ParseOrEmpty does, but I am not a fan of using empty string instead of null to mark that something is missing, as it introduces two ways to express that data is missing.

Cql/CqlSdkPrototype/Elm/Extensions/ElmApiExtensions.Adding.cs
```diff
@@ -33,13 +33,13 @@ public static TElmApi AddElmFile<TElmApi>(
...
 
-        if (versionedLibraryIdentifier.Version is null)
+        if (versionedLibraryIdentifier.Version.IsEmpty)
             throw new FileNotFoundException($"Could not find file '{file.FullName}'.");
 
         return elmApi.UseLogger((elmApi, logger) =>
         {
             logger.LogWarning("Could not load library from file with name and version, trying without version: {file}", file.FullName);
-            file = new FileInfo(Path.Combine(directory.FullName, $"{versionedLibraryIdentifier with { Version = null }}.json"));
+            file = new FileInfo(Path.Combine(directory.FullName, $"{versionedLibraryIdentifier with { Version = CqlLibraryVersion.Empty }}.json"));
```
ewoutkramer: So, it looks like a special value to indicate missing. Why did you choose that over a simple null?

Cql/CqlSdkPrototype/Infrastructure/CqlLibraryVersion.cs
```diff
@@ -6,10 +6,12 @@
     IComparable<CqlLibraryVersion>,
     IComparable
 {
-    public static CqlLibraryVersion Empty { get; } = new CqlLibraryVersion("");
+    public static CqlLibraryVersion Empty { get; } = new("");
```
ewoutkramer: Oh no, it's the empty string. I see this as an anti-pattern.

Cql/CqlToElmTests/CoercionTest.cs
```diff
@@ -1,27 +1,22 @@
...
     [TestClass]
     public class CoercionTest : Base
     {
-        internal static CoercionProvider CoercionProvider => ServiceProvider.GetCoercionProvider();
-        internal static ElmFactory ElmFactory =>  ServiceProvider.GetElmFactory();
+        private static CoercionProvider CoercionProvider => CreateCqlApi(
+            EnableListPromotion:true,
+            EnableListDemotion:true,
+            EnableIntervalPromotion:true,
+            EnableIntervalDemotion:true).AsInternal().State.ServiceProvider.GetRequiredService<CoercionProvider>();
```
ewoutkramer: If the CqlApi would also be a service provider, this code could be shortened (when it turns out it is common to get to these services).

ewoutkramer (changes requested):

Cql/CqlSdkPrototype/Cql/CqlApiState.cs
baseTwo: CqlLibraryVersion is a struct, so if it cannot parse it, it returns the default value

baseTwo: Remove CqlLibraryString.Empty and all empty pattern objects on public surface

baseTwo: Fixed

Cql/CqlSdkPrototype/Elm/Extensions/ElmApiExtensions.Adding.cs
baseTwo: Fixed

Cql/CqlSdkPrototype/Infrastructure/CqlLibraryVersion.cs
baseTwo: Fixed

Cql/CqlToElmTests/CoercionTest.cs
baseTwo: Exceptional cases only for internal testing

ewoutkramer (approved):

PR #712, opened 2025-01-28
move measure build out of elm defs

richfirely:


baseTwo:
Could you please add a detailed description to this PR outlining its purpose and the rationale behind it? 

Additionally, please include the relevant issue number for reference.

richfirely:
It's just a minor bug where the measure building was happening on every elm definition statement. I moved it up so it builds once per library. No ticket. I found it testing building measures.

baseTwo (approved):

PR #731, opened 2025-02-17
730 44 Remove Public APi Prototype project

### LINKED ISSUE #730: Public API | Remove prototype project
The public api (toolkits) was developed inside one CqlSdkPrototype project. These  classes should now be moved to the appropriate package projects, and then the temporary prototype project must be removed

baseTwo:
Work for
* #730 

**Work**
* Remove CqlSdkPrototype project
* Moved toolkits to respective projects (a new sub namespace `Toolkit`):
  * CQL toolkit -> Cql.CqlToElm
  * ELM tookit -> Cql.CodeGeneration.NET
  * Packaging toolkit -> Cql.Packaging
  * Invocation toolkit -> Cql.Invocation (NEW PROJECT)

Project Dependencies
image


Cql/Cql.Runtime/Defaults.cs
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * This file is licensed under the BSD 3-Clause license
+ * available at https://raw.githubusercontent.com/FirelyTeam/firely-cql-sdk/main/LICENSE
+ */
+
+namespace Hl7.Cql.Runtime;
+
+internal static class Defaults
+{
+    public static EnumerationOptions EnumerationOptions { get; } = new();
```
ewoutkramer, line 13: I would make an EnumerationOptions.Default instead, easier for the user to discover than a separate Default class.

ewoutkramer (approved):
Nicely documented too.

Just a single remark.

Cql/Cql.Runtime/Defaults.cs
baseTwo: I agree, but this is Microsoft's class not ours

PR #746, opened 2025-02-21
handle large whitespace before library and version

### LINKED ISSUE #745: Public API | CqlLibraryString Regex stuck on trailing spaces
**Describe the bug**
When processing the same QICoreCommon CMS CQL file directly from ecqf repository the files are unix and have some trailing space in the lines. This is causing what seems like an infinite loop in the regex processor.

**To Reproduce**
Compile the attached CQL file which is sourced from ecqf.
OR add a ton of trailing space to the end of any line above/before the library name.

[QICoreCommon.txt](https://github.com/user-attachments/files/18916420/QICoreCommon.txt)

**Expected behavior**

**Screenshots**

Image

**Version used:**
 - Version: 2.0 dev

**Additional context**

richfirely:


Cql/Cql.CqlToElm/CqlLibraryString.cs
```diff
@@ -36,14 +36,15 @@ public static CqlLibraryString FromIdentifierAndString(
     /// </summary>
     /// <returns>A <see cref="Regex"/> to extract the library name and version.</returns>
     [GeneratedRegex("""
-                        (\s+\r?\n?)*        # Skip whitespace and newlines
+                        ^\s*                # Skip whitespace and newlines
```
baseTwo: The `^` needs to be replaced with a `\A` to mean `# The start of the string`. Could you place this on a separate line with that comment?

baseTwo: The behaviour changed such that it no longer includes newline as part of the initial whitespace. Is that intentional?

baseTwo (changes requested):

Cql/Cql.CqlToElm/CqlLibraryString.cs
```diff
@@ -36,14 +36,15 @@ public static CqlLibraryString FromIdentifierAndString(
...
                         library             # until "library" is found
                         \s+
-                        (?<lib>\S+)         # The name of the library
+                        (?<lib>\S+)         # The name of the library
```
baseTwo: We need to be more specific which characters are allowed as part of the library ([qualifiedIdentifier](https://cql.hl7.org/19-l-cqlsyntaxdiagrams.html#qualifiedIdentifier)) and the version ([STRING](https://cql.hl7.org/19-l-cqlsyntaxdiagrams.html#STRING)).

Cql/Cql.CqlToElm/CqlLibraryString.cs
richfirely: \A doesn't appear to work. Carret is beginning of a line or string in multiline

Cql/Cql.CqlToElm/CqlLibraryString.cs
richfirely: Good point. There's a lot to consider here on the variations of [identifier](https://cql.hl7.org/19-l-cqlsyntaxdiagrams.html#IDENTIFIER). I think we should get this working for now to support the known measures we already test and then use a new ticket to revisit all the regex to handle the multitudes of the way library can look. OK?

Cql/Cql.CqlToElm/CqlLibraryString.cs
baseTwo: I tested it with this bit of code. Since we select the multiiline option, is is important to find the library close to the top of the file and not just the start of any line. Are there any other characters above yours? Now that I'm thinking about it, I did not consider the possibility of comments which may appear before the library keyword.

```csharp
	string rex =
"""
\A               # Start of string
\s*              # Skip whitespace including newlines
library          # until "library" is found
\s+
(?<lib>\S+)      # The name of the library                                      
(
\s+              # at least one space
version
\s+              # at least one space
'(?<ver>[^']+)'   # The version of the library between single quotes
)?               # Version is optional
""";

	Regex.Match(
		"""		


library "library" version '1.2.3'
""", rex, RegexOptions.IgnorePatternWhitespace | RegexOptions.Multiline).Dump();

```

Cql/Cql.CqlToElm/CqlLibraryString.cs
baseTwo: Issue created 
- #748

Cql/Cql.CqlToElm/CqlLibraryString.cs
richfirely: Yes that's exactly the case here on qicore file in the ticket, comments above library line.

richfirely: Here's the example

```csharp
string rex =
"""
^               # Start of string
\s*              # Skip whitespace including newlines
library          # until "library" is found
\s+
(?<lib>\S+)      # The name of the library                                      
(
\s+              # at least one space
version
\s+              # at least one space
'(?<ver>[^']+)'   # The version of the library between single quotes
)?               # Version is optional
""";

	Regex.Match(
		"""		


/*
@update: JKR 2024-04-02 ->
Check display code accuracy against source system                                                        
Ensure all fluent functions are not referencing any deprecated functions.   
https://oncprojectracking.healthit.gov/support/browse/CQLIT-414
*/


library QICoreCommon version '2.1.000'

using QICore version '4.1.1'

include FHIRHelpers version '4.4.000'

codesystem "LOINC": 'http://loinc.org'
""", rex, RegexOptions.IgnorePatternWhitespace | RegexOptions.Multiline).Dump();
```

baseTwo: Ok, then we need to find a way to skip any comments above the library keyword. I'll have a look

Cql/Cql.CqlToElm/CqlLibraryString.cs
```diff
@@ -36,14 +36,15 @@ public static CqlLibraryString FromIdentifierAndString(
...
                         library             # until "library" is found
                         \s+
-                        (?<lib>\S+)         # The name of the library
+                        (?<lib>\S+)         # The name of the library                                      
                         (
-                          \s+
+                          \s+               # at least one space
                           version
-                          \s'(?<ver>[^']+)' # The version of the library between single quotes
+                          \s+               # at least one space
```
richfirely: FYI. This line appears to be pretty important. The NCQA LSC_Bulk, LSC_Details, and LSC_Reporting are not building without considering multiple spaces between library and version.

Cql/Cql.CqlToElm/CqlLibraryString.cs
richfirely: Best I could come up with after reading about this is removing the comments. Also this will help when the comments have the same keywords we are looking for  see below.

```csharp
void Main()
{
	string input = """		
	/*
	@update: JKR 2024-04-02 -> see
	library Ok version  '5'
	Check display code accuracy against source system                                                         
	Ensure all fluent functions are not referencing any deprecated functions.   
	https://oncprojectracking.healthit.gov/support/browse/CQLIT-414
	/* /
	*/
	
	//test comment
	
	/* test comment */

	library QICoreCommon version '2.1.000'

	using QICore version '4.1.1'

	include FHIRHelpers version '4.4.000'

	codesystem "LOINC": 'http://loinc.org'
	""";

	string rex =
	"""
    (\s+\r?\n?)*               # Start of string
    \s*            # Skip whitespace including newlines
    library        # until "library" is found
    \s+
    (?<lib>\S+)    # The name of the library                                      
    (
    \s+            # at least one space
    version
    \s+            # at least one space
    '(?<ver>[^']+)' # The version of the library between single quotes
    )?             # Version is optional
    """;
	// Remove comments (block and line)
	string noComments = Regex.Replace(input, @"/\*.*?\*/|//.*?$", string.Empty, RegexOptions.Singleline | RegexOptions.Multiline);
	

	noComments.Dump();


	// Match against cleaned input
	var match = Regex.Match(noComments, rex, RegexOptions.IgnorePatternWhitespace | RegexOptions.Multiline);

	if (match.Success)
	{
		Console.WriteLine($"\n");
		Console.WriteLine($"Library: {match.Groups["lib"].Value}");
		Console.WriteLine($"Version: {match.Groups["ver"].Value}");
	}
	else
	{
		Console.WriteLine("No match found.");
	}
}
```

baseTwo (approved):

PR #749, opened 2025-02-25
fix identifiers, refactor, and fix known issues.

### LINKED ISSUE #714: Identifier format (max length 64 chars, and underscore handling)
The format of the id in a FHIR Library Resource was changed from {libraryName}-{libraryVersion}  to {libraryName} in this PR
- #384 

Please look into why this was needed, since there is a need to have resources with duplicate names but diferent versions

Reopened on 2025-03-04. ID cannot be longer than 64 characters.

richfirely:
- Fixed identifiers as described in ticket
- Had to reintroduce LibrarySet to a few methods to support walking down the libraries to get Related Artifacts and ValueSets.
- Retired Resource.CanonicalUri - will do test removal in cleanup
- Moved ValueSets from parameters to related artifacts

Should ResourceFileNames be left as is or change id to name everywhere. I could see using this generically?


richfirely:
@baseTwo can you give this a once over for general direction approval while I do cleanup and fix build? I'm stopping here so the PR doesn't get too big.

baseTwo (dismissed):

richfirely:
I am going to leave ResourceFileNames as is for now. I think when refine on packaging more items such as valuesets (which would use ID), we can consider another modification later.

baseTwo (approved):

PR #768, opened 2025-03-18
Feature/767 improve vs handling

### LINKED ISSUE #767: Improve code lookups in valuesets
There are a few things that make integrating the CQL SDK and the .NET SDK slow or tedious:

* HashValueSetDictionary requires you to load all valuesets up front, and it is not possible to pass in the SDK's ResourceResolvers to do it just-in-time.
* HashValueSetDictionary and some of the operations like Union are not doing their utmost to make sure that list of codes that ARE valuesets (and thus IValueSetFacade) will remain facades, which means the `In` operator does a linair scan instead of a hashtable lookup.
* Different parts of the terminology-related operations use different comparison logic.

Fixes for this have been in the `introduce-terminology-services` branch for about two years, so let us try to integrate these changes into 2.0.

ewoutkramer:
This PR is a refresh of the terminology operations to make sure we retain IValueSetFacades as much as possible, and makes comparison logic (through the ICqlComparers) consistent (Union, ValueSetFacade and HashValueSetDictionary used different hashers from the CqlCodeCqlComparer!).  

We also add a new `ValueSetSource` that improves on the existing `ValueSetLoader` plus `HashValueSetDictionary` by
* Removing the need for a loader, valuesets CAN be loaded up front, but can also be loaded as needed
* Leveraging the more elaborate expansion logic from the FHIR SDK.

I have changed all the tests over to using the new ValueSetSource, which means I was able to remove the `ValueSetLoader` entirely, as its expansion logic is too flawed to keep around.

:warn: Also review https://github.com/FirelyTeam/Ncqa.DQIC/pull/19 and https://github.com/FirelyTeam/Firely.Cql.Sdk.Integration.Runner/pull/36


Cql/CoreTests/Input/ValueSets/intensional-value-set.json
```diff
@@ -7,10 +7,14 @@
     "include": [
       {
         "valueSet": [
-          "https://www.ncqa.org/fhir/valueset/2.16.840.1.113883.3.464.1004.1009",
```
ewoutkramer, line 10: This is wrong, it means select all codes that occur in BOTH valuesets, which is an empty collection. Instead, it should have been two separate `valueSet` statements, which is then handled as an OR.

Cql/Cql.Runtime/ValueSets/CqlCodeExtensions.cs
```diff
@@ -9,44 +9,59 @@
...
+        new CqlCodeEqualityComparer<T>(comparer ?? throw new ArgumentNullException(nameof(comparer)), precision, useEquivalence);
 
-            public bool Equals(CqlCode? x, CqlCode? y) => Comparer.Compare(x, y, null) == 0;
+    private class CqlCodeEqualityComparer<T>
+        (ICqlComparer<T> comparer, string? precision = null, bool useEquivalence = false)
+        : IEqualityComparer<T>
+    {
+        public ICqlComparer<T> Comparer { get; } = comparer ?? throw new ArgumentNullException(nameof(comparer));
 
-            public int GetHashCode(CqlCode obj) =>
```
ewoutkramer, line 44: So, we have a Comparer for equality, but for hashing the original code uses something else. This is bound to give problems in HashSets, since its lookup assumes that when x.Equals(y) then x.GetHashCode() == y.GetHashCode().

Cql/Cql.Runtime/ValueSets/CqlValueSetFacade.cs
```diff
@@ -46,39 +45,14 @@ public CqlValueSetFacade(string id, IValueSetDictionary valueSets)
         private readonly string _canonical;
         private readonly IValueSetDictionary _valuesets;
 
-        /// <summary>
-        /// Returns <see langword="true"/> if <paramref name="code"/> is in this value set.
-        /// </summary>
-        /// <param name="code">The code to check.</param>
-        /// <returns><see langword="true"/> if <paramref name="code"/> is in this value set.</returns>
-        public bool? IsCodeInValueSet(CqlCode? code)
```
ewoutkramer, line 54: Although it makes sense that the return of this function is null when the parameter is null, this actually never happens if you look at the actual invocations.  Also, CqlCode.code being null should not be a reason to return a null here. I think that kind of null  propagation should be in the binding, not here in these basic operators.

Cql/Cql.Runtime/ValueSets/CqlValueSetFacade.cs
```diff
@@ -46,39 +45,14 @@ public CqlValueSetFacade(string id, IValueSetDictionary valueSets)
...
-        {
-            if (code == null || system == null)
-                return null;
-            else return _valuesets.IsCodeInValueSet(_canonical, code, system);
-        }
+        /// <inheritdoc/>
+        public bool IsCodeInValueSet(string code, string? system) => _valuesets.IsCodeInValueSet(_canonical, code, system);
+
+        /// <inheritdoc/>
+        public bool IsCodeInValueSet(string code) => _valuesets.IsCodeInValueSet(_canonical, code);
```
ewoutkramer, line 40: There was divergent behaviour for how to handle a system being null. In some implementations this meant: find a code with no system, in other implementations of this interface (and IValueSetDictionary) it meant: find any code that matches. I have split these up into two separate methods.

Cql/Cql.Runtime/ValueSets/HashValueSetDictionary.cs
```diff
@@ -108,28 +92,7 @@ public bool TryGetCodesInValueSet(string valueSetUri, out IReadOnlyCollection<Cq
...
-        private readonly Dictionary<string, HashSet<CqlCode>> _codesInValueSet =
-            new(StringComparer.OrdinalIgnoreCase);
-
-        private class CqlCodeHasher : IEqualityComparer<CqlCode>
-        {
-            public bool Equals(CqlCode? x, CqlCode? y) => true;
-
-            // we're using low ASCII values that are invalid in real codes
-            public int GetHashCode(CqlCode obj) =>
-                ($"{obj.code ?? "\x1"}\x1").GetHashCode()
```
ewoutkramer, line 129: Yet another way to hash.

Cql/Cql.Runtime/ValueSets/HashValueSetDictionary.cs
```diff
@@ -3,101 +3,85 @@
...
-        /// <returns><see langword="true"/> if the given code is present in the given value set.</returns>
+        /// <param name="codes">The codes to add.</param>
+        /// <exception cref="ArgumentException">If the valueset already exists in the dictionary.</exception>
+        public void Add(string valueSetUri, IEnumerable<CqlCode> codes)
+        {
+            if (_codesInValueSet.ContainsKey(valueSetUri))
+                throw new ArgumentException($"Valueset {valueSetUri} already exists in dictionary.");
+
+            var internedCodes = codes.Select(Intern);
+            _codesInValueSet.Add(valueSetUri, new InMemoryValueSet(internedCodes));
```
ewoutkramer, line 63: I have reused the `InMemoryValueSet` instead of re-doing comparison and lookup logic in the HashValueSetDictionary. The implementation becomes a simple hash of valusets + lookup in the valueset. This makes it possible to return ValueSetFacades when necessary, not lists of codes from a shared hashtable.

Cql/Cql.Runtime/ValueSets/HashValueSetDictionary.cs
```diff
@@ -3,101 +3,85 @@
...
-        /// Tries to ge the codes in the value set as an <see cref="IReadOnlyCollection{CqlCode}"/>.
+        /// Tries to get the codes in the value set as an <see cref="IReadOnlyCollection{CqlCode}"/>.
         /// </summary>
         /// <param name="valueSetUri">The value set's canonical URI.</param>
         /// <param name="codes">The <see langword="out"/> parameter for the value set's codes, or <see langword="null"/>.</param>
         /// <returns><see langword="true"/> if the given value set is defined; otherwise, <see langword="false"/>.</returns>
-        public bool TryGetCodesInValueSet(string valueSetUri, out IReadOnlyCollection<CqlCode>? codes)
+        public bool TryGetCodesInValueSet(string valueSetUri, out IEnumerable<CqlCode>? codes)
         {
             if (_codesInValueSet.TryGetValue(valueSetUri, out var codeSet))
```
ewoutkramer, line 86: This is *not* returning a `IValueSetFacade`, which means everything coming out of here requires a full lineair scan.

Cql/Cql.Firely/ValueSetSource.cs
```diff
@@ -0,0 +1,207 @@
...
+        return new ValueSetExpander(expansionOptions);
+    }
+
+
+    /// <summary>
+    /// Adds a list of <see cref="ValueSet"/>s to the cache, so they will not be retrieved using the resolver.
+    /// </summary>
+    public async Task<IReadOnlyCollection<IValueSetFacade>> Add(IEnumerable<ValueSet> vslist)
+    {
+        var tasks = vslist.Select(Add);
```
baseTwo: This can potentially saturate the thread pool. Perhaps better to use the Parallel type to set a maximum degree of parallelization and reduce the amount of concurrent tasks

Cql/Cql.Firely/ValueSetSource.cs
```diff
@@ -0,0 +1,207 @@
...
+
+/// <summary>
+/// Helper methods for constructing an <see cref="IValueSetDictionary"/> from a collection of <see cref="ValueSet"/>s.
+/// </summary>
+public static class ValueSetSourceExtensions
+{
+    /// <summary>
+    /// Construct a new <see cref="IValueSetDictionary"/> from the given <paramref name="values"/>.
+    /// </summary>
+    public static IValueSetDictionary ToValueSetDictionary(this IEnumerable<ValueSet> values)
```
baseTwo: Would be nice to have an async version of this as well

Cql/Cql.Firely/ValueSetSource.cs
```diff
@@ -0,0 +1,207 @@
...
+        {
+            return false;
+        }
+    }
+}
+
+/// <summary>
+/// Helper methods for constructing an <see cref="IValueSetDictionary"/> from a collection of <see cref="ValueSet"/>s.
+/// </summary>
+public static class ValueSetSourceExtensions
```
baseTwo: Should be ValueSetExtensions

Cql/Cql.Runtime/Comparers/CqlCodeCqlComparer.cs
```diff
@@ -86,16 +112,17 @@ public bool Equivalent(CqlCode x, CqlCode y, string? precision)
...
 
+        /// <inheritdoc/>
         public bool Equivalent(object? x, object? y, string? precision) => Equivalent((x as CqlCode)!, (y as CqlCode)!, precision);
 
+        /// <inheritdoc/>
         public int GetHashCode(CqlCode? x) =>
             x == null
             ? typeof(CqlCode).GetHashCode()
-            : $"{x.code ?? "null"}\0{x.system}\0".GetHashCode();
+            : $"{x.code?.ToLowerInvariant() ?? "null"}\0{x.system?.ToLowerInvariant()}\0".GetHashCode();
```
baseTwo: Since GetHashCode might get used millions of times, I think it could boost performance to calculate the hashcode on x.code by using ReadOnlySpans, instead of creating strings on the heap

baseTwo (changes requested):

Cql/Cql.Firely/ValueSetSource.cs
ewoutkramer: I have no knowledge of parrallelism APIs I must admit, this was a suggestion from copilot. I will turn it into something sync.

ewoutkramer: Mmmm....but this is an async method. Maybe I can use your help here, I don't know how to do that.

Cql/Cql.Firely/ValueSetSource.cs
ewoutkramer: In fact, it should just be async I think, I'll take a look at the call sites to see if you are working async already anyway.

Cql/Cql.Runtime/Comparers/CqlCodeCqlComparer.cs
ewoutkramer: I don't know anything about Spans, can you show me how?

ewoutkramer: We should do it in the IdentifierComparer too then.

Cql/Cql.Firely/ValueSetSource.cs
ewoutkramer: Ok, got curious. I replaced it with Parallel.ForEachAsync, that seems to work. I am reading in the docs that I should not bother setting the max number of threads myself, so I skipped doing that. Let me know if that needs to change.

Cql/Cql.Runtime/ValueSets/InMemoryValueSet.cs
```diff
@@ -15,32 +15,69 @@ namespace Hl7.Cql.ValueSets
     /// <summary>
     /// An <see cref="IValueSetFacade"/> that has all the codes in memory.
     /// </summary>
-    internal class InMemoryValueSet : IValueSetFacade
+    public class InMemoryValueSet : IValueSetFacade
     {
-        private static readonly ICqlComparer _defaultComparer = new StringCqlComparer(StringComparer.OrdinalIgnoreCase);
+        private const string NullCodeSystem = "\0";
```
baseTwo, line 20: Can we get an comment explanation why we need a null char string instead of an empty or null string?

Cql/Cql.Firely/ValueSetSource.cs
```diff
@@ -0,0 +1,213 @@
...
+            var codes = ToCodes(vs.Expansion.Contains);
+            return new InMemoryValueSet(codes);
+        }
+    }
+
+    /// <summary>
+    /// Given a canonical, returns the <see cref="IValueSetFacade"/> for that canonical from the
+    /// cache, or uses the resolver to load it.
+    /// </summary>
+    public async Task<IValueSetFacade?> Load(string canonical)
```
baseTwo, line 104: Async methods need a CancellationToken parameter

baseTwo (approved):

Cql/Cql.Runtime/ValueSets/InMemoryValueSet.cs
ewoutkramer: Because Evan thought it looked cool?  I see he uses it to separate code and system too, so probably an inheritance from the old C days where a string got terminated by a `\0` ;-)

Cql/Cql.Firely/ValueSetSource.cs
ewoutkramer: https://github.com/FirelyTeam/firely-net-sdk/issues/3081

PR #790, opened 2025-04-02
776 public api cleanup cql comparers 5

### LINKED ISSUE #776: Public API | Clean up CQL Comparers
**Additional context**
Read this on the Java repo:  https://github.com/cqframework/cql-engine/issues/70

baseTwo:
Also review related submodules!
* [Integration Runner](https://github.com/FirelyTeam/Firely.Cql.Sdk.Integration.Runner/pull/39)

More information in CoPilot's PR review in the comment section below.

**Findings and Solution**

This PR aims to get the null-handling logic consistent with CQL comparison, equality and equivalence. Logic was scattered and inconsistent for comparison, half of comparers returned 1 or -1 if on or the other input was null, and others returned null if any input was null. In many cases equality and equivalence derives from the output of comparison, making the outcomes of those inconsistent too.

It was discovered that based on context, this logic can change, where the context is from the CQL operators. e.g. Greater Than would return null if any input is null. 

So the solution is to keep the comparer logic consistent (returning 1 or -1 if either but not both inputs are null), and only where an operator has different rules, handle them there before calling the comparer.

Since the operator already handles nullability, it also made sense to have methods on the comparer interfaces where null is no longer a concern and only values are accepted.

**CQL Comparer and Equivalence Interface Changes**

The non-generic `ICqlComparer` and `IEquivalenceComparer` adds additional methods paired with the existing methods, where the original methods accepts null parameters, and the additional methods don't. This skips unnecessary null checking when comparers are chained.

`IEquivalentable<T>` and `ICqlComparable<T>` were changed so that it only accepts non-null arguments. Null checks are handled in the comparers, not the instances that implement these interfaces. (See implementation of this in `CqlDate`, `CqlDateTime`, `CqlTime`)

**CqlComparer<T> Base Type**
Refactored all comparers to derive from common base type `CqlComparer<T>`. This ensures the behaviour for CQL comparisons is consistent.

`CqlComparer<T>` Type Hierarchy
```
CqlComparer<T>(in Hl7.Cql.Comparers)
    CqlCodeCqlComparer(in Hl7.Cql.Comparers)
    CqlCodeCqlEquivalentComparer(in Hl7.Cql.Comparers)
    CqlComparers(in Hl7.Cql.Comparers)
    CqlConceptCqlComparer(in Hl7.Cql.Comparers.CqlComparers)
    CqlQuantityCqlComparer(in Hl7.Cql.Comparers.CqlComparers)
    CqlTupleTypeComparer(in Hl7.Cql.Comparers.CqlComparers)
    DecimalCqlComparer(in Hl7.Cql.Comparers.CqlComparers)
    InterfaceCqlComparer<T>(in Hl7.Cql.Comparers.CqlComparers)
    KeyValuePairCqlComparer<TKey, TValue>(in Hl7.Cql.Comparers.CqlComparers)
    ListEqualComparer(in Hl7.Cql.Comparers.CqlComparers)
    CqlComparerWrapper<T, TInner>(in Hl7.Cql.Comparers)
        CodeOfEnumCqlComparer<TEnum>(in Hl7.Cql.Fhir.Comparers)
        ResourceIdCqlComparer(in Hl7.Cql.Fhir.Comparers)
        PlainCqlComparerWrapper<T>
    DefaultCqlComparer<T>(in Hl7.Cql.Comparers)
    EnumComparer(in Hl7.Cql.Comparers)
    IntervalComparer<T>(in Hl7.Cql.Comparers)
    StringCqlComparer(in Hl7.Cql.Comparers)
    LegacyTupleBaseTypeComparer(in Hl7.Cql.CqlToElm.Test)
    FhirEnumComparer(in Hl7.Cql.Fhir.Comparers)
    IdentifierComparer(in Hl7.Cql.Fhir.Comparers)
    IValueComparer<T>(in Hl7.Cql.Fhir.Comparers)
    PrimitiveTypeAgainstStringComparer(in Hl7.Cql.Fhir.Comparers)
```

`CqlComparer<T>` implements the generic `ICqlComparer<T>` and `IEquivalenceComparer<T>` interfaces _implicitly_, while the non-generic types ICqlComparer and IEquivalenceComparer are implemented _explicitly_. 

Since Equality and Equivalence can be derived from each other or through Comparison, `CqlComparer<T>` allows derived types to divert implementation.

**CqlOperators**
Comparer behaviour may be overridden in `CqlOperators` type by checking for nullability first, such that if any input is null, the result is null, before passing on to the registered comparers.

Bugfix:  When calling equivalence on strings from `CqlOperators`, the `IEnumerable<T>` overload was picked over object, meaning the strings were compared character by character. This has been fixed.

**Other Refactoring**
* Various types moved to other projects or namespaces
* `TupleBaseTypeComparer` moved to a file class `LegacyTupleBaseTypeComparer` in `XmlTests`. The legacy tuples are used when generating definitions, but are replaced in C# with value tuples. `XmlTests `still invokes definitions, instead of the generated C# assembles
* `KeyValuePairComparer` moved to `CqlComparer.KeyValuePairCqlComparer`
* `CharCqlComparer` removed. A bug was fixed where this was used to compare `string` by enumerating its characters
* Where possible, use singleton instances of comparers, instead of creating new instances.


Cql/Cql.Firely/Extensions/CqlComparersExtensions.cs
```diff
@@ -45,44 +45,23 @@ public static CqlComparers AddFhirComparers(this CqlComparers comparers)
...
-                return base.Compare(x, y, precision);
-            }
-
-            public override bool? Equals(
-                object? x,
-                object? y,
-                string? precision) =>
-                throw new UnreachableException("CqlComparers always goes through Compare, so we never reach here");
-        }
+        private enum DummyEnum {}
```
ewoutkramer, line 69: I think it's worth a few lines of documentation to describe what you use this for.

Cql/Cql.Runtime/Operators/MessageEventArgs.cs
```diff
@@ -6,6 +6,15 @@
  * available at https://raw.githubusercontent.com/FirelyTeam/firely-cql-sdk/main/LICENSE
  */
 
+
```
ewoutkramer, line 9: Double copyright?

ewoutkramer (approved):

PR #794, opened 2025-04-04
Support string based unit definition in convert-statement

### LINKED ISSUE #793: Conversion expressions with target unit as string are not implemented
Expressions like 

```cql
convert 1.75 'm' to 'cm'
```

where the target unit is given as a string can't be converted to ELM.

The conversion throws an exception and logs a message like:

```
warn: Hl7.Cql.CqlToElm.Toolkit.CqlToolkit[0]
      Could not translate CQL to ELM: Convert-1.0.0
      System.InvalidOperationException: Syntax error; expected unit.
```

The spec for [conversionExpressionTerm](https://cql.hl7.org/19-l-cqlsyntaxdiagrams.html#conversionexpressionterm) and [unit](https://cql.hl7.org/19-l-cqlsyntaxdiagrams.html#unit-1) states that a unit can be given as a `dateTimePrecision`, `pluralDateTimePrecision` or a `string`, but the `string` case is not implemented in `VisitConversionExpressionTerm()` ([link](https://github.com/FirelyTeam/firely-cql-sdk/blob/8ef8dde6709bc9aa032b8ca6127a8ca1f09f1ab0/Cql/Cql.CqlToElm/Visitors/UnaryExpressionVisitor.cs#L255)).

**To Reproduce**

Run the following code to reproduce:

```csharp
CqlToolkitConfig cqlConfig = new([CqlModel.ElmR1, CqlModel.Fhir401]);
var libraryString = CqlLibraryString.Parse("""
   library Convert version '1.0.0'

   define "Convert m to cm": convert 1.7 'm' to 'cm'
   """);

var cqlToolkit = new CqlToolkit(LoggerFactory, config: cqlConfig)
    .AddCqlLibraries(libraryString)
    .ConvertCqlToElm();
```

**Expected behavior**

The expression should be properly converted into valid ELM.

**Version used:** 2.0.13-build-20250404-1

and-hus:
Fixes #793 

Handle case for a unit given as a string in `VisitConversionExpressionTerm()` via additional `else if`-statement

Adding unit test for meter to centimeter conversion.

### DIFF
--- a/Cql/CqlToElmTests/(tests)/ConvertTest.cs
+++ b/Cql/CqlToElmTests/(tests)/ConvertTest.cs
@@ -47,5 +47,28 @@ namespace Hl7.Cql.CqlToElm.Test
 
         }
 
+        [TestMethod]
+        public void Convert_Meters_to_Centimeters()
+        {
+            var library = CreateCqlToolkit().MakeLibraryFromExpression("convert 5 'm' to 'cm'");
+            var quantity = library.Should().BeACorrectlyInitializedLibraryWithStatementOfType<ConvertQuantity>();
+            var result = Run(quantity, library);
+            var q = result.Should().BeOfType<CqlQuantity>().Subject;
+            q.value.Should().Be(500m);
+            q.unit.Should().Be("cm");
+        }
+
+        /* Mass units are not yet defined and respective conversions not yet implemented, see UcumUnits.cs and UnitConverter.cs
+        [TestMethod]
+        public void Convert_Grams_to_Kilograms()
+        {
+            var library = CreateCqlToolkit().MakeLibraryFromExpression("convert 500 'g' to 'kg'");
+            var quantity = library.Should().BeACorrectlyInitializedLibraryWithStatementOfType<ConvertQuantity>();
+            var result = Run(quantity, library);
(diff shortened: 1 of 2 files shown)


Cql/CqlToElmTests/(tests)/ConvertTest.cs
```diff
@@ -47,5 +47,28 @@ convert q to weeks
...
+        {
+            var library = CreateCqlToolkit().MakeLibraryFromExpression("convert 5 'm' to 'cm'");
+            var quantity = library.Should().BeACorrectlyInitializedLibraryWithStatementOfType<ConvertQuantity>();
+            var result = Run(quantity, library);
+            var q = result.Should().BeOfType<CqlQuantity>().Subject;
+            q.value.Should().Be(500m);
+            q.unit.Should().Be("cm");
+        }
+
+        /* Mass units are not yet defined and respective conversions not yet implemented, see UcumUnits.cs and UnitConverter.cs
```
baseTwo: Could you uncomment that test, and just attribute the test to be skipped with that message?

baseTwo (changes requested):

baseTwo (approved):

PR #837, opened 2025-05-12
Add subjectType in FHIR library based on ELM context

### LINKED ISSUE #833: subjectCodeableConcept should be set to "Patient" for CQL libraries with Patient contexts
If the CQL library contains "context Patient", make sure that this information is accessible in the FHIR Library resource. This is done by setting subjectCodeableConcept to "Patient". See https://www.hl7.org/fhir/R4/valueset-subject-type.html

alexzautke:
Fixes #833

### DIFF
--- a/Cql/Cql.Packaging/ResourcePackager.cs
+++ b/Cql/Cql.Packaging/ResourcePackager.cs
@@ -345,10 +345,10 @@ internal static class LibraryPackager
         fhirLibrary.Status = PublicationStatus.Active;
         fhirLibrary.Date = new DateTimeIso8601(date.ToLocalTime(), Iso8601DateTimePrecision.Millisecond).ToString();
 
-        //if (elmLibrary.contexts.Any(context => nameof(ResourceType.Patient).Equals(context.name)))
-        //{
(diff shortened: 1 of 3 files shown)


Cql/Cql.Packaging/ResourcePackager.cs
```diff
@@ -345,6 +345,11 @@ private static FhirLibrary CreateFhirLibrary(
         fhirLibrary.Status = PublicationStatus.Active;
         fhirLibrary.Date = new DateTimeIso8601(date.ToLocalTime(), Iso8601DateTimePrecision.Millisecond).ToString();
 
+        if (elmLibrary.contexts.Any(context => nameof(ResourceType.Patient).Equals(context.name)))
```
baseTwo: Is this only for Patient resource types, or could there be others as well?

alexzautke: In the standard there could be others, but I don't think that we support anything besides "context Patient" at the moment in the engine

baseTwo (approved):

PR #885, opened 2025-07-04
886 embed debug symbols pdf files to fhir libraries

### LINKED ISSUE #886: Package fhir libraries with debug symbols if available
See parent issue for details.

This issue is to
* The PackagingToolkit must be able to accept pdb file
* The PackagingToolkit must embed the pdf files into FHIR Libraries, when they are provided
* Some refactoring required to clean up the input/result artifacts for each toolkit

baseTwo:
Part of a 2 sub issue task.

The main goal is to add debug symbols to the FHIR Library resource, if they are provided. For that it needs to be passed along the results of the ElmToolkit, to the PackagingToolit.

The `ResourcePackager.InputArtifacts` has a new `DebugSymbols` property.

Since the PackagingToolkit had a complex input/result artifact data structure, it was simplified into a struct for inputs `PackagingToolkitInputArtifacts` and results `PackagingToolkitResultArtifacts`, and a pair that contains both `PackagingToolkitArtifacts`. This theme is duplicated in the CQL and ELM toolkits as well, so some name refactoring was done:
* The toolkit's `Conversions` property is renamed to `ArtifactsById`
* The ELM toolkit takes and ELM Library as input, and `ElmToolkitResultArtifacts` as result (containing C#, DLL and PDB). They are paired into `ELMToolkitArtifacts`
* The CQL toolkit takes a CQL Library as input and ELM Library as output. They are paired into `CqlToolkitArtifacts`

The toolkit extensions to get results have been updated to return a tuple of individual artifacts, instead of a C# record to contain them, which just adds unnecessary bloat.

More unit tests added to `LibraryPackagerTest`, for testing the properties and attachments on the FHIR Library after it was packaged

The next step is to update the Packager CLI to accept a commandline argument to enable building of debug symbols. (An appsetting for it is already present)


Examples/CqlSdkExamples/251 Package incl debug symbols via CQL Toolkit, Save FHIR Resources to dir/PackagerCliExample.ps1
```diff
@@ -0,0 +1,14 @@
+$cqlInDir = "input/cql"
+$fhirOutDir = "output/fhir"
+
+cql-package cql `
+    --cql "$cqlInDir" `
+    --fhir "$fhirOutDir"
+# REVIEW- TODO: We need an attribute here allowing us to specify that debug symbols be generated. Perhaps with a pdb folder (or sharing the dll folder)
```
baseTwo, line 7: Will be added in #887

Examples/CqlSdkExamples/220 Load ELM from dir, Save C#,DLL,PDB to dir/PackagerCliExample.ps1
```diff
@@ -6,6 +6,7 @@ cql-package cql `
     --cql "$cqlInDir" `
     --cs "$csOutDir" `
     --dll "$dllOutDir"
+# REVIEW- TODO: We need an attribute here allowing us to specify that debug symbols be generated. Perhaps with a pdb folder (or sharing the dll folder)
```
baseTwo, line 9: Will be added in #887

Cql/CodeGeneration.NET/Toolkit/ElmToolkit.cs
```diff
@@ -70,35 +70,39 @@ public ElmToolkit SetBatchProcessExceptionContinuation(BatchProcessExceptionCont
...
-    private void ReplaceConversions(
-        ElmToolkitConversionDictionary conversions)
-    {
-        _conversions = conversions;
-    }
+    /// <remarks>This method updates the internal state to use the provided collection of artifacts. Ensure
+    /// that <paramref name="artifactsById"/> contains valid data before calling this method.</remarks>
+    /// <param name="artifactsById">The collection of artifacts to replace the current collection.  This parameter cannot be null.</param>
+    private void ReplaceArtifactsById(
+        ElmToolkitArtifactsById artifactsById) =>
```
ewoutkramer, line 88: Are there `ElmToolkitArtifactsBy....` any other sorting? Why are you exposing that in the name?

baseTwo: I  use that convention for dictionaries. Just seeing a Dictionary<string, Artifact> it is not obvious what the key is

Cql/Cql.CqlToElm/Toolkit/CqlToolkit.cs
```diff
@@ -107,7 +114,7 @@ public CqlToolkit AddCqlLibraries(IEnumerable<CqlLibraryString> cqlLibraries)
                                                  logMessage("Could not add CQL library to CqlToolkit: {lib}.", conversionRecord.LibraryIdentifier)));
 
         if (count > 0)
-            ReplaceConversions(conversions.ToImmutable());
+            ReplaceArtifactsById(builder.ToImmutable());
```
ewoutkramer, line 117: The name suggests you replace the "artifacts" by id, but rather you are replacing the "artifactsbyid" ;-)

baseTwo: Agree it could be ambiguous. Happy to name this method ReplaceArtifacts, but the type and fields name, I'd like to keep "artifacts"By"Id" to my "value"By"Key" naming convention

ewoutkramer: Looking at it again, I think it is indeed the name of the function that confuses me, not the "byid" names of fields or properties.

ewoutkramer (approved):

ewoutkramer:
Oh, this repo merges with unresolved comments ;-)

PR #932, opened 2025-08-11
827 Port to SDK6

### LINKED ISSUE #827: Update the CQL SDK to use the new FHIR SDK.
As we are preparing for SDK6, we need to have a SDK6 version of the CQL SDK, so FirelyServer can test these new versions.

ewoutkramer:
These are the changes to the main CQL SDK necessary to upgrade to SDK6.

We should not update the main branch (develop-2.0) yet I think - but we will need to be able to create a version of the CQL SDK that we can use in the preliminary SDK6 version of Firely Server.

This PR also includes PRs from the subrepos:
* https://github.com/FirelyTeam/Ncqa.DQIC/pull/33
* https://github.com/FirelyTeam/Firely.Cql.Sdk.Integration.Runner/pull/53

⚠️ DO NOT MERGE THIS INTO DEVELOP-2.0 YET⚠️


alexzautke:
Suggest to merge into branch "develop2.0-sdk6" and adjust the build pipeline to produce packages for that branch as well.

ewoutkramer:
In summary, what needed to be done (only major things listed):

* Element.id has changed from a simple string to a full FhirString type in SDK6, this causes changes in property access and generated code
* CQL allows times/datetimes without a timezone, FHIR does not. We were (accidentally) lenient about that in SDK5, but I have tightened the rules. I needed to change code that assumed FHIR could work without timezones as well. BTW,
* Base no longer implementes `IEnumerable<KeyValuePair<>>`, had to add comparison capabilities to make that work again via the new `EnumerateElements()` call.
* `ISystemAndCode` is gone, official workaround is via ICoded.ToCodings().SIngle. Do wonder whether we should bring that interface back as the new way is not necessarily performant.
* Fixed errors (actually, warnings) coming from the new nullability annotations.
* Parsing error codes `<type>_LITERAL_INVALID_CODE` is now just `LITERAL_INVALID_CODE`, needed changes when filtering errors.
* `FhirJsonPocoDeserializerSettings` were expected for the `ForFhir()` call, but it now `FhirJsonConverterOptions`. This is very common, so maybe we should re-introduce FhirJsonPocoDeserializerSettings as an obsolete subclass of FhirJsonConverterOptions, so we can point the user in the right direction.

baseTwo:
⚠️ DO NOT MERGE THIS INTO DEVELOP-2.0 YET ⚠️

Cql/Cql.Firely/Comparers/BaseComparer.cs
```diff
@@ -0,0 +1,32 @@
+﻿/*
+ * Copyright (c) 2024, Firely, NCQA and contributors
```
baseTwo: 2025

Cql/Cql.Firely/ValueSetSource.cs
```diff
@@ -76,6 +76,8 @@ public async Task<IReadOnlyCollection<IValueSetFacade>> Add(IEnumerable<ValueSet
     /// </summary>
     public async Task<IValueSetFacade> Add(ValueSet vs)
     {
+        if(vs.Url is null) throw new ArgumentNullException(nameof(vs), "ValueSet must have a URL to be added to the cache.");
```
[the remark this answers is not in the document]
baseTwo: I agree, but nameof(vs) should be the parameter name for ArgumentException

Cql/Cql.Firely/Comparers/PrimitiveTypeAgainstStringComparer.cs
```diff
@@ -30,7 +30,7 @@ internal class PrimitiveTypeAgainstStringComparer(ICqlComparer inner) :
                     throw new InvalidOperationException(
                         $"Precision '{precision}' is not supported for comparing Code<T> to string.");
 
-                return StringComparer.Ordinal.Compare(xCode.ObjectValue, yString);
+                return StringComparer.Ordinal.Compare(xCode.JsonValue, yString);
```
baseTwo: This is going to conflict with PR #931 where the type `PrimitivieTypeAgainstStringComparer` was renamed, as well as changes to this method

Cql/Cql.Runtime/Comparers/CqlComparers.CqlComparerImplementation.cs
```diff
@@ -32,12 +32,25 @@ partial class CqlComparers : CqlComparer<object>
             }
         }
 
-        ICqlComparer? comparer = null;
-        if (Comparers.TryGetValue(xType, out var c))
+        var comparer = SelectComparer(x, xType);
```
baseTwo, line 48: This will be a merge conflict with PR #931 . Suggest merging from develop-2.0 and fix

baseTwo (changes requested):
I've made some comments. I also suggest merging PR #931  into develop-2.0, then develop-2.0 into these branches, and then to fix the merge conflicts in the comparers

Cql/Cql.Firely/Comparers/PrimitiveTypeAgainstStringComparer.cs
ewoutkramer: Yes, I noticed that too while doing a review, but that's a minor thing, no problem.

Cql/Cql.Firely/ValueSetSource.cs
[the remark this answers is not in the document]
ewoutkramer: Agree.

Cql/Cql.Firely/Comparers/PrimitiveTypeAgainstStringComparer.cs
ewoutkramer: Fixed.

baseTwo (approved):

PR #960, opened 2025-08-20
checkin latest measures

richfirely:
@baseTwo I am checking in all the latest copies of measures from CMS.

Updated CMS folder with qi-core-2024 (QI-Core 4.0)
Added new folder qi-core-2025  (QI-Core 6.0)
Added new folder cms-2025   (QI-Core 6.0)

I cleaned up all the CQL I knew of (date fix for QICoreCommon) and there were a few bad version references which I will report to the authors.

I also updated launchsettings and did my best to add the synonymous exclusions to appsettings.

With that said, there are compile issues in all folders. For WGM in mid September my priority is to get the CMS-2025 measures working. Would you mind looking at those as the priority and see what fixes are needed. We will eventually need to resolve all issues, but the CMS subset is "refined" for connectathon testing in Sept.


alexzautke:
This was only closed as "develop-2.0" now longer exists, need to be pointed again to the new develop

PR #1019, opened 2025-09-09
Fix the method to add given quantity to given date of max value

### LINKED ISSUE #1018: Fix - method to add a quantity to a given date of max value 9999-12-31 throws exception and not catched
Description:
The functionality to add a given quantity to a given date results in ArgumentOutOfRangeException. This is because current implementation does not catch the exception when the given date is of max value, see below

```
public static readonly CqlDate MaxValue = new(9999, 12, 31);

public CqlDate? Add(CqlQuantity? quantity)
{
    if (quantity == null || quantity.value == null || quantity.unit == null)
        return null;
    quantity = quantity.NormalizeTo(Precision);
    var value = quantity.value!.Value;
    var dto = Value.DateTimeOffset;
    switch (quantity.unit![0])
    {
       ...........

```
Reproduce Issue and Expected Behavior:

To reproduce the issue, please add below tests under Primitive Tests and note that for specifically the Add method, when it tries to add 1 day to max value date 9999-12-31, it encounters an exception

```
[TestMethod]
public void Add_Date_Quantity()
{
    var rc = GetNewContext();
    var fcq = rc.Operators;

    var inputDate = new CqlDate(9999, 12, 30);
    var quantity = new CqlQuantity(1, "day");
    CqlDate expectedDate = new CqlDate(9999, 12, 31);

    var newDate = fcq.Add(inputDate, quantity);
    Assert.IsNotNull(newDate);
    Assert.AreEqual(expectedDate, newDate);

    var inputDateMaxValue = CqlDate.MaxValue;

    Assert.ThrowsException<ArgumentOutOfRangeException>(() =>
    {
        var newDateAddMax = fcq.Add(inputDateMaxValue, quantity);
    });
}
```

For reference, below test passes the expected output as the implementation for similar Subtract method has a catch block which returns null. We are expecting Add function to behave in similar fashion.

```
[TestMethod]
public void Subtract_Date_Quantity()
{
    var rc = GetNewContext();
    var fcq = rc.Operators;

    var inputDate = new CqlDate(1, 1, 2);
    var quantity = new CqlQuantity(1, "day");
    CqlDate expectedDate = new CqlDate(1, 1, 1);

    var newDate = fcq.Subtract(inputDate, quantity);
    Assert.IsNotNull(newDate);
    Assert.AreEqual(expectedDate, newDate);

    var inputDateMinValue = CqlDate.MinValue;
    var newDateSubtractedMin = fcq.Subtract(inputDateMinValue, quantity);
    Assert.IsNull(newDateSubtractedMin);
}
```

Version used:
-sdk 2 develop branch

igajurelNCQA:
Fixes https://github.com/FirelyTeam/firely-cql-sdk/issues/1018


Cql/CoreTests/PrimitiveTests.cs
```diff
@@ -3539,5 +3539,44 @@ public void QuantityToString()
             var s = ops.ConvertQuantityToString(new CqlQuantity(125, "cm"));
             s.Should().Be("125 'cm'");
         }
+
+        [TestMethod]
+        public void Add_Date_Quantity()
```
baseTwo, line 3545: Split these up into regular addition and outofrange-returns-null tests

Cql/CoreTests/PrimitiveTests.cs
```diff
@@ -3539,5 +3539,44 @@ public void QuantityToString()
...
+            Assert.AreEqual(expectedDate, newDate);
+
+            var inputDateMaxValue = CqlDate.MaxValue;
+
+            var newDateAddMax = fcq.Add(inputDateMaxValue, quantity);
+            Assert.IsNull(newDateAddMax);
+        }
+
+        [TestMethod]
+        public void Subtract_Date_Quantity()
```
baseTwo, line 3571: Split these up into regular addition and outofrange-returns-null tests

Cql/Cql.Abstractions/Primitives/CqlDate.cs
```diff
@@ -97,42 +97,50 @@ public static bool TryParse(string s, out CqlDate? cqlDate)
             quantity = quantity.NormalizeTo(Precision);
             var value = quantity.value!.Value;
             var dto = Value.DateTimeOffset;
-            switch (quantity.unit![0])
+            try
```
baseTwo: CqlTime and CqlDateTime should then also follow the same logic (returning null when addition/subtraction goes out of range)?

baseTwo (changes requested):

baseTwo:
Since the CQL specification does not explicitly define what should happen in under/overflow cases (e.g. dates) it is the expectation the behavior is implementation-dependent: so either resulting in null or an exception.

It looks like the behavior chosen is to result in a null.

Some questions:
- Should this then not also be implemented on time, datetime, integers, etc... ?
- Perhaps when an out of range exception is caught, ICqlOperators.Message could be called with details

igajurelNCQA:
> Since the CQL specification does not explicitly define what should happen in under/overflow cases (e.g. dates) it is the expectation the behavior is implementation-dependent: so either resulting in null or an exception.
> 
> It looks like the behavior chosen is to result in a null.
> 
> Some questions:
> 
> * Should this then not also be implemented on time, datetime, integers, etc... ?
> * Perhaps when an out of range exception is caught, ICqlOperators.Message could be called with details

Yes, similar fixes need to added for Add/Subtract methods under Time, DateTime as well. The bug is scoped to identify ArgumentOutOfRangeException when quantity is being added/subtracted to a Cql date values which can be Max/Min default ones. So, I can propagate this fix to Time and DateTime stuffs where Add/Subtract a quantity is being done.

Regarding adding the ICqlOperators.Message, I can see that the Message method could be invoked on a Cql operator that implements ICqlOperators interface. However, in this context, I am not finding a straight-forward way to add it here. Could you elaborate with a possible example.

baseTwo:
> > Since the CQL specification does not explicitly define what should happen in under/overflow cases (e.g. dates) it is the expectation the behavior is implementation-dependent: so either resulting in null or an exception.
> > It looks like the behavior chosen is to result in a null.
> > Some questions:
> > 
> > * Should this then not also be implemented on time, datetime, integers, etc... ?
> > * Perhaps when an out of range exception is caught, ICqlOperators.Message could be called with details
> 
> Yes, similar fixes need to added for Add/Subtract methods under Time, DateTime as well. The bug is scoped to identify ArgumentOutOfRangeException when quantity is being added/subtracted to a Cql date values which can be Max/Min default ones. So, I can propagate this fix to Time and DateTime stuffs where Add/Subtract a quantity is being done.
> 
> Regarding adding the ICqlOperators.Message, I can see that the Message method could be invoked on a Cql operator that implements ICqlOperators interface. However, in this context, I am not finding a straight-forward way to add it here. Could you elaborate with a possible example.

Since the implementation is in `CqlOperators.Add`, I would handle exceptions there instead of in CqlDate.Add. An implementation could look like this, perhaps the catch block could use logging instead of sending a message, we would have to determine that.

```csharp
public CqlDate? Add(CqlDate? left, CqlQuantity? right)
{
    if (left == null || right == null)
        return null;
    try
    {
        return left.Add(right);
    }
    catch (ArgumentOutOfRangeException e)
    {
        Message(new { left, right, e }, "some code", "Warning", "Date addition overflow ignored, returned null.");
        return null;
    }
}
```

baseTwo:
I noticed from this page on [datetime arithmetic](https://cql.hl7.org/02-authorsguide.html#datetime-arithmetic), that "[If the resulting year is out of range, an error is thrown](https://cql.hl7.org/02-authorsguide.html#datetime-arithmetic:~:text=If%20the%20resulting%20year%20is%20out%20of%20range%2C%20an%20error%20is%20thrown)"

Then on another page on the addition of primitive types (int, decimal, quantity) "[If the result of the addition cannot be represented (i.e. arithmetic overflow), the result is null.](https://cql.hl7.org/09-b-cqlreference.html#add:~:text=If%20the%20result%20of%20the%20addition%20cannot%20be%20represented%20(i.e.%20arithmetic%20overflow)%2C%20the%20result%20is%20null.)". It does not mention temporal types here specifically, but I assume this rule would apply as a whole to addition.

The spec can be really contradictory at times

Cql/Cql.Abstractions/Primitives/CqlDate.cs
```diff
@@ -97,42 +97,50 @@ public static bool TryParse(string s, out CqlDate? cqlDate)
...
+                        break;
+                    case 's':
+                        dto = dto.AddSeconds(Math.Truncate((double)value));
+                        break;
+                    default: throw new ArgumentException($"Unknown date unit {quantity.unit} supplied");
+                }
+            }
+            catch (ArgumentOutOfRangeException)
+            {
+                // In cases where e.g. Predecessor is called on minimum Date.
```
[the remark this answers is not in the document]
igajurelNCQA: Not applicable, with new approach to handle exception from cql operators

Cql/CoreTests/PrimitiveTests.cs
igajurelNCQA: done

Cql/CoreTests/PrimitiveTests.cs
igajurelNCQA: done

Cql/Cql.Abstractions/Primitives/CqlDate.cs
igajurelNCQA: Yes that is required and has been added now. Thanks!

igajurelNCQA:
> I noticed from this page on [datetime arithmetic](https://cql.hl7.org/02-authorsguide.html#datetime-arithmetic), that "[If the resulting year is out of range, an error is thrown](https://cql.hl7.org/02-authorsguide.html#datetime-arithmetic:~:text=If%20the%20resulting%20year%20is%20out%20of%20range%2C%20an%20error%20is%20thrown)"
> 
> Then on another page on the addition of primitive types (int, decimal, quantity) "[If the result of the addition cannot be represented (i.e. arithmetic overflow), the result is null.](https://cql.hl7.org/09-b-cqlreference.html#add:~:text=If%20the%20result%20of%20the%20addition%20cannot%20be%20represented%20(i.e.%20arithmetic%20overflow)%2C%20the%20result%20is%20null.)". It does not mention temporal types here specifically, but I assume this rule would apply as a whole to addition.
> 
> The spec can be really contradictory at times

Yes, that seems to be correct on arithmetic operators. I have added similar fixes for those as well and some unit tests.

baseTwo (approved):

PR #1026, opened 2025-09-11
Refactor library preprocessing with new classes

### LINKED ISSUE #1008: Fix qicore-2025 Bugs: Revisit AliasRef.resultTypeSpecifier resolver logic
Double check the functionality in : ExpressionBuilderContext.FixMissingAliasRefTypeSpecifiers
- Maybe find a generic way to extract this into a separate class, injected as a service
- Should this be added to CQL to ELM too, to fix ELM?
- Compare against Java repo

baseTwo:
Related submodule PR: https://github.com/FirelyTeam/Firely.Cql.Sdk.Integration.Runner/pull/73

This PR refactors the library preprocessing infrastructure by introducing new `LibraryPreprocessor` and `LibraryPreprocessorBuilder` classes to replace the existing `ElmPreprocessor`. The preprocessing logic is reorganized into a dedicated namespace with separate corrector classes. Additionally, the changes update generated code files and configuration settings.

- Introduces `LibraryPreprocessor` and `LibraryPreprocessorBuilder` as replacements for `ElmPreprocessor`
- Reorganizes preprocessing logic into `Hl7.Cql.Compiler.Preprocessing` namespace with specialized corrector classes
- Updates dependency injection and builder patterns to use the new preprocessor infrastructure
- Introduced `MissingResultTypeSpecifierCorrector` which set missing `resultTypeSpecifier` for elements with only `resultTypeName`


Cql/Cql.Compiler/Preprocessing/AmbiguousOverloadCorrector.cs
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2024, NCQA and contributors
```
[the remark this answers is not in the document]
baseTwo: This is just a renamed file

Cql/Cql.Compiler/Preprocessing/LibraryPreprocessor.cs
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (c) 2024, NCQA and contributors
```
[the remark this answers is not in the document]
baseTwo: This is a renamed file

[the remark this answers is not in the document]
baseTwo: fix

alexzautke:
@ewoutkramer Would you mind taking over this PR review?

Cql/Cql.Compiler/ExpressionBuilderContext.LibraryDefs.cs
```diff
@@ -137,7 +131,16 @@ public void ProcessExpressionDef(
                     }
                 }
 
-                CreateAndAddDefinition(expressionDefName, parameters, parameterTypes, originalParameterNames);
+                if (expressionDef.expression is { } expressionDefExpression)
```
ewoutkramer, line 134: Why is inlining the function a good idea?

Cql/Elm/ElmTreeWalker.cs
```diff
@@ -14,9 +14,26 @@ namespace Hl7.Cql.Elm
     /// <param name="visitor">A function that is passed the nodes in the tree. It should return <c>false</c> if this function
     /// handles the children of the nodes by itself, or <c>true</c> if the walker should visit the children automatically.</param>
     /// <param name="filter">A function that, given the type of a property, controls whether the walker visits that child property or not.</param>
-    internal class ElmTreeWalker(Func<object, bool> visitor,
-                               Func<Type, bool>? filter = null) : BaseElmTreeWalker
+    internal class ElmTreeWalker(
+        ElmTreeWalker.Visitor visitor,
+        ElmTreeWalker.Filter? filter = null) : BaseElmTreeWalker
     {
+        public delegate bool Visitor(object node);
```
ewoutkramer, line 21: I think turning Action<T>/Func<T> into delegates is an anti-pattern, is there added value to doing this?

ewoutkramer (commented):
Nothing major, just some double checks.

Cql/Elm/ElmTreeWalker.cs
baseTwo: I added additional delegates (e.g. `VisitorWithSelf` with parameters `self` and `node`. ). I prefer named parameters with these, but for consistency I also changed the func/action into named delegates. I would bother me to have a mix between the two, so which one do you prefer here? I find it useful to have names for parameters.

Cql/Cql.Compiler/ExpressionBuilderContext.LibraryDefs.cs
baseTwo: The local function was small but had many parameters, so I found it less readable since we have another 5 additional lines just to take the parameters. It also uses scoped variables, which is easier to understand when they're used in the nested context here. 

I think local functions only make sense if it really simplifies the main content of a complex method a lot, such that it becomes a simple recipe of 10 to 15 lines or so, but this method doesn't really gain anything by putting these few lines in a local function when all the other code isn't

baseTwo:
@ewoutkramer , would you have a look at reviewing my replies?

baseTwo:
> Nothing major, just some double checks.

There is one additional corrector to set missing `resultTypeSpecifier` when `resultTypeName` is known on each element. This has an effect on some generated code such that a few methods use the correct return type instead of `object`

ewoutkramer:
> > Nothing major, just some double checks.
> 
> There is one additional corrector to set missing `resultTypeSpecifier` when `resultTypeName` is known on each element. This has an effect on some generated code such that a few methods use the correct return type instead of `object`

Yes, I saw it, I liked it!

ewoutkramer (approved):

PR #1027, opened 2025-09-14
Change CQL Long mapping to return String for R4

### LINKED ISSUE #1028: Change CQL Long mapping to return String for R4
Update the CQL Long type mapping for FHIR R4 to return String instead of throwing a NotSupportedException, to align with the latest CQL-FHIR conformance specifications.

See latest version of https://build.fhir.org/ig/HL7/cql-ig/conformance.html#fhir-type-mapping

alexzautke:
Updated mapping for CQL Long type to return String in R4 instead of throwing an exception.

See latest version of https://build.fhir.org/ig/HL7/cql-ig/conformance.html#fhir-type-mapping


baseTwo (changes requested):
Instead of removing the unit test, please update it to match the change in the outcome; assert that type entry for Long is a string

baseTwo (approved):

PR #1041, opened 2025-09-25
1012 fix qicore 2025 bugs tuple element value does not have a resulttypespecifier

### LINKED ISSUE #1012: Fix qicore-2025 Bugs: Tuple element value does not have a resultTypeSpecifier
There are a number of elm files that are skipped in the packager appsettings files like so:
```
      // Tuple element value does not have a resultTypeSpecifier
      "CMS2FHIRPCSDepressionScreenAndFollowUp.json",
      "CMS145FHIRCADBetaBlockerTherapyPriorMIorLVSD.json",
      "CMS832HHAKIFHIR.json",
```

Remove these skipped files and discover the issue at runtime when running the PackagerCLI for those appsetting specific configurations (e.g. `Hl7.Cql.Packager.ecqm-content-qicore-2025.appsettings.json` maps to launch configuration `ecqm-content-qicore-2025`)

Notice that the error originates from `ExpressionBuilderContext.cs` line 229 : `throw new InvalidOperationException($"Tuple element value does not have a resultTypeSpecifier").WithContext(this)))`

Find a way to try and resolve these during preprocessing

baseTwo:
This PR fixes QICore 2025 ELM compilation bugs related to missing `resultTypeSpecifier` properties on tuple element values. The fix enhances the `MissingResultTypeSpecifierCorrector` preprocessor to handle additional cases where type specifiers are missing in tuple operations, particularly in CQL expressions that return complex structured data.

Key changes:
- Enhanced the MissingResultTypeSpecifierCorrector with additional logic to handle tuple element type specifications
- Removed previously skipped CQL files from packager configurations
- Updated submodule references and added generated C# files for newly supported measures

### DIFF
--- a/Cql/Cql.Compiler/Preprocessing/MissingResultTypeSpecifierCorrector.cs
+++ b/Cql/Cql.Compiler/Preprocessing/MissingResultTypeSpecifierCorrector.cs
@@ -39,8 +39,13 @@ internal class MissingResultTypeSpecifierCorrector(ILogger<MissingResultTypeSpec
                 {
                     resultTypeSpecifier: ListTypeSpecifier { elementType: { } elementType },
                     @return: { resultTypeSpecifier: null } returnClause,
-                })
+                } query)
             {
+                logger.LogDebug(
+                    "Setting missing ReturnClause.resultTypeSpecifier to ListTypeSpecifier.elementType to '{resultType}' on Query @ {locator}.\n{expressionKey}",
+                    elementType,
+                    query.locator,
+                    self.ContextStackString);
                 returnClause.resultTypeSpecifier = elementType;
             }
         }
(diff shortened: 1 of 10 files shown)


Cql/Cql.Compiler/Preprocessing/MissingResultTypeSpecifierCorrector.cs
```diff
@@ -18,17 +19,68 @@ internal class MissingResultTypeSpecifierCorrector(ILogger<MissingResultTypeSpec
...
+            {
+                logger.LogDebug(
+                    "Setting missing Tuple.element.value.resultTypeSpecifier from TupleTypeSpecifier on Tuple @ {locator}.\n{expressionKey}",
+                    tuple.locator,
+                    self.ContextStackString);
+                foreach (var (index, tupleElement) in elements.Select((v, i) => (i, v)))
+                {
+                    tupleElement.value.resultTypeSpecifier ??= tupleTypeSpecifier.element[index].elementType;
+                }
+            }
```
ewoutkramer, line 88: Why would copilot have turned the switch into consecutive anonymous code blocks with nested ifs? At least, it looks like that on my diff?

I think these could all be neat cases on a switch, expressing concisely the different types of nodes that are corrected here.

ewoutkramer (changes requested):

Cql/Cql.Compiler/Preprocessing/MissingResultTypeSpecifierCorrector.cs
baseTwo: They should not be in a switch, because they could go into more than one clause, eg. an Expression or a Query

baseTwo: I added a code comment to make it more clear what the idea is and not to refactor it to a switch

ewoutkramer (approved):

PR #1051, opened 2025-10-23
Improve performance with lazy and global caching

baseTwo:


Cql/CodeGeneration.NET/LibrarySetCSharpCodeGenerator.cs
```diff
@@ -145,17 +145,29 @@ public LibraryWriter(
...
+                return null;
+            }
+        }
         public LibraryWriter AddIndent(int addIndent = 1)
         {
             return this with { IndentedTextWriter = IndentedTextWriter.AddIndent(addIndent) };
         }
 
-        public void WriteLibraryFile()
+        public void WriteLibraryFile(Library library)
```
baseTwo, line 166: You already have access to the library via the `Library` property, no need for this parameter

Cql/CodeGeneration.NET/LibrarySetCSharpCodeGenerator.cs
```diff
@@ -208,8 +220,9 @@ private void WriteLibraryInterfaceImplementation()
                                 .GetLibraryDependencies(LibraryName, throwError: true)
                                 .Select(dep => IdentifierNormalizer.Normalize(dep.VersionedLibraryIdentifier))
                                 .Select(typeName => $"{typeName}.Instance");
+
             IndentedTextWriter.WriteLine($"""
-                                          public ILibrary[] Dependencies => [{string.Join(", ", dependencies)}];
+                                          public ILibrary[] Dependencies => [];
```
baseTwo, line 225: Please do not remove the dependencies

Cql/CodeGeneration.NET/LibrarySetCSharpCodeGenerator.cs
```diff
@@ -256,7 +269,7 @@ private void WriteNamespaceFileScope()
                       .OfType<CqlCodeDefinition>()
                       .ToArray());
 
-        private void WriteMethods()
+        private void WriteMethods(Library library)
```
baseTwo, line 272: You already have access to the library via the `Library` property, no need for this parameter

Cql/CodeGeneration.NET/LibrarySetCSharpCodeGenerator.cs
```diff
@@ -308,7 +321,7 @@ private void WriteMethods()
             }
         }
 
-        private void WriteClass()
+        private void WriteClass(Library library)
```
baseTwo, line 324: You already have access to the library via the `Library` property, no need for this parameter

baseTwo (commented):
Just some early comments. There is no need to pass the library on each method of the `LibraryWriter` class, since it already has a property for it

DamienMosier:
@baseTwo Curious where this is as Ishan stated that after a call you were wanting to try a different approach. We cannot integrate the 2.0 engine until performance is restored.

baseTwo:
> @baseTwo Curious where this is as Ishan stated that after a call you were wanting to try a different approach. We cannot integrate the 2.0 engine until performance is restored.

This PR is stale and should be closed. The plan agreed was that small incremental steps towards performance gains will be implemented. 

The first step would be to disable the distinct operation on query results, (or just use the `all` keyword in the CQL. Ishan probably has more information on the performance gains here. 

We agreed not to start with my proposal until that step was completed first.

igajurelNCQA:
@baseTwo 

**Regarding Distinct:**
@DamienMosier had mentioned during the meeting, that if "return all" indeed does not bind distinct operator automatically, then applying the distinct fixes and testing won't make any change to performance. So, I verified the current implementation as-is, and it seems true that "return all" basically ensures that distinct operator won't be applied automatically. So, our CQLs have been mostly updated already to "return all", and we can update the rest (if any) as well to "return all". Therefore, I am not seeing any potential performance gains on this!

```

define "Test Define 1":
    ("All intervals") interval
    **return**
        end of interval
		
[CqlExpressionDefinition("Test Define 1")]
public IEnumerable<CqlDate> Test_Define_1(CqlContext context)
{
	IEnumerable<CqlInterval<CqlDate>> a_ = this.All_intervals(context);
	CqlDate b_(CqlInterval<CqlDate> interval)
	{
		CqlDate e_ = context.Operators.End(interval);

		return e_;
	};
	**IEnumerable<CqlDate> c_ = context.Operators.Select<CqlInterval<CqlDate>, CqlDate>(a_, b_);**
	**IEnumerable<CqlDate> d_ = context.Operators.Distinct<CqlDate>(c_);**

	return d_;
}

define "Test Define 2":
    flatten(
        ("All intervals") interval
        **return all** LibraryABC."Function ABC"( 
            "Claims", 
            "EOBs", 
            "Conditions", 
            "Observations", 
            Patient.birthDate, 
            interval
        ) 
    )


    [CqlExpressionDefinition("Test Define 2":)]
    public IEnumerable<IEnumerable<CqlQuantity>> Test_Define_2:(CqlContext context)
    {
        IEnumerable<CqlInterval<CqlDate>> a_ = this.All_intervals(context);
        IEnumerable<CqlQuantity> b_(CqlInterval<CqlDate> interval)
        {
            IEnumerable<Claim> d_ = this.Claims(context);
            IEnumerable<ExplanationOfBenefit> e_ = this.EOBs(context);
            IEnumerable<Condition> f_ = this.Conditions(context);
            IEnumerable<Observation> g_ = this.Observations(context);
            Patient h_ = this.Patient(context);
            Date i_ = h_?.BirthDateElement;
            CqlDate j_ = FHIRHelpers_4_0_1.Instance.ToDate(context, i_);
            IEnumerable<CqlQuantity> k_ = LibraryABC.Instance.Function_ABC(context, d_, e_, f_, g_, j_, interval);

            return k_;
        };
        **IEnumerable<IEnumerable<CqlQuantity>> c_ = context.Operators.Select<CqlInterval<CqlDate>, IEnumerable<CqlQuantity>>(a_, b_);**

        return c_;
    }
```


I think we should move forward with your proposals on tweaking the Caching changes. I was in an impression earlier that the testing on "distinct" and the POC proposals on tweaking these Caching changes are going in parallel. My bad if I missed anything.

baseTwo:
Fixed in PR
* #1078

PR #1089, opened 2026-01-07
Spike/remove id example working

### LINKED ISSUE #1044: need to edit FHIRHelpers with FHIR.Id function
**Describe the bug**
To compile to C# both CMS and NCQA measures require this edit to FHIRHelpers.cql

`define function ToString(value FHIR.id): value. Value`

**Reproduction Steps**
1. remove this line from FhirHelpers.cql. (It only exists because I added it to get the repo to compile)
2. See complaints from compile in output window in VS.

**Expected behavior**
We shouldn't have to edit the CQL pre-compile - was this interpretation removed from SDK at some point or Elm/spec issue?

**Version used:**
Latest 2.0

**Additional context**

richfirely:
Showing @baseTwo [regarding ](https://github.com/FirelyTeam/firely-cql-sdk/issues/1044) https://github.com/FirelyTeam/firely-cql-sdk/issues/1044

FHIR.id basetype is FHIR.string in the [modelinfo](https://github.com/FirelyTeam/firely-cql-sdk/blob/f90faf49183fd21b2d18b6405c5fd1ed30e443e5/Cql/Cql.Model/Models/fhir-modelinfo-4.0.1.xml#L11778).

I think we just need to ensure that ID is always converted to FHIR.string then the code here will work properly.

```
Id e_ = Locations?.IdElement;
FhirString f_ = context.Operators.Convert<FhirString>(e_);   <-------- missing conversion
string g_ = FHIRHelpers_4_4_000.Instance.ToString(context, f_);
```

### DIFF
--- a/Cql/Cql.Firely/FhirTypeResolver.cs
+++ b/Cql/Cql.Firely/FhirTypeResolver.cs
@@ -123,6 +123,8 @@ namespace Hl7.Cql.Fhir
 
             Types["{http://hl7.org/fhir}SimpleQuantity"] = Types["{http://hl7.org/fhir}Quantity"];
             Types["{http://hl7.org/fhir}MoneyQuantity"] = Types["{http://hl7.org/fhir}Quantity"];
+
+            Types["{http://hl7.org/fhir}Id"] = typeof(Hl7.Fhir.Model.FhirString);
         }
 
         private void AddTypesFromInspector()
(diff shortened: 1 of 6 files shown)


baseTwo:
Replaced by:
* https://github.com/FirelyTeam/firely-cql-sdk/pull/1091

PR #1169, opened 2026-02-17
replace library urls to use dash when underscore exists.

### LINKED ISSUE #1168: Fix library url with underscores
**Describe the bug**
HEDIS libraries use underscore in the CQL naming. This has downstream impact on file names etc, but the simplest working "fix" is to follow the specification in library generation output.

https://build.fhir.org/ig/HL7/cql-ig/en/conformance.html#library-name-and-url

`The prohibition against underscores in CQL library names is required to ensure compliance with the canonical URL pattern (because URLs by convention should not use underscores). In addition, many publishing environments will use the canonical tail (i.e. the name of the library) as the logical id of the Library resource, which does not allow underscores per the FHIR specification.`

I think fixing the URL for now is sufficient and a good safeguard. I'll get a PR ready.

richfirely:
1. Replace library url underscores with dash
2. Minor fix to use path for RelatedArtifact libraries (deduplicates when see multiple times under alias)


baseTwo (changes requested):
I disagree that it is the responsibility of the SDK to fix the identifier. I can provide you with a script to fix filenames and cql load statements instead. 

If we allow this, it has a big impact on how we track input and output files, since they will be named differently. There might be more consequences than this.

We should disallow underscores in the root url PackagingToolkitConfig.CanonicalRootUrl

Cql/Cql.Packaging/ResourceCanonicalBuilderFactory.cs
```diff
@@ -23,6 +23,7 @@ public static ResourceCanonicalBuilder Create(
             string identifier,
             string? version = null) =>
         {
+            identifier = identifier.Replace('_', '-');
```
baseTwo, line 27: Best place to put this is rather in FhirLibraryExtensions.Create, just after calling resourceCanonicalBuilder and before setting it on the Url

baseTwo (changes requested):
Just one comment to place the identifier fix in the Create method of a FhirLibrary

alexzautke:
Shouldn’t we introduce a setting to this? This has downstream impact. Shouldn’t we tell the author by default there’s an issue so that it can be corrected in the source?

Cql/Cql.Packaging/ResourceCanonicalBuilderFactory.cs
richfirely: This would make it so we "fix" it in multiple places - measures, related artifacts, libraries. Why not in 1 place?

richfirely:
> Shouldn’t we introduce a setting to this? This has downstream impact. Shouldn’t we tell the author by default there’s an issue so that it can be corrected in the source?

It's not an authors issue as much as a workflow/spec issue given by the CQL/ELM spec. Underscore appears to be perfectly valid up until FHIR Library/Measure generation. While we address it with NCQA and/or the CQL Specs we need to do something to keep things working.

alexzautke:
> While we address it with NCQA and/or the CQL Specs we need to do something to keep things working.

And that could be based on an option that needs to be enabled to we do this as suggested. The other option is to let the author remove the underscore and use a dash throughout the process.

richfirely:
Ticket https://github.com/FirelyTeam/firely-cql-sdk/issues/1188 was entered to address further refinement while I merge this stop gap.

baseTwo (approved):

PR #1282, opened 2026-05-27
Inject IMetricService into CQL runtime for UCUM unit arithmetic

### LINKED ISSUE #1177: Replace the old Fhir.Metrics library with the faster Hl7.
The UCUM type from the Fhir.Metrics library is considered slow/legacy since it is using a lot of string based operations.

It needs to be replaced by [FirelyMetricService](https://github.com/FirelyTeam/Firely.Ucum/blob/develop/Firely.Ucum/FirelyMetricService.cs) (implementing IMetricsService) in Firely.Ucum, but before picking this up, reach out to the Firely .NET SDK, because it does not implement division yet!

ewoutkramer:
## Summary

- Adds an `IMetricService` injection point to `FhirCqlContext`/`FhirCqlContextOptions` so callers can supply a custom UCUM implementation instead of the built-in default.
- Introduces `DefaultUcumMetricService` (internal) — wraps `FhirMetricService` directly (the temporary `TryConvertTo` workaround has been removed now that `Fhir.Metrics` 1.4.0 implements it natively).
- Improves several `CqlOperators` methods that operate on `CqlQuantity` to route through the injected `IMetricService` where it was previously ignored:
  - `Add` and `Subtract`: cross-unit calendar aliases (day/days) and commensurable UCUM units (kg+g via `TryAdd`/`TrySubtract`), returning results in the most granular of the two input units per CQL §9.B
  - `Divide`: cross-unit UCUM division via `TryDivide` (e.g. `2 kg / 1 g → 2000`)
  - `Multiply`: scalar multiplication now preserves the non-`'1'` operand's unit instead of collapsing it to `'1'`
  - `Between`: uses `Comparer` which routes through the service for cross-unit comparisons, returning `null` (not an incorrect boolean) for dimension-incompatible comparisons
  - `CanConvertQuantity`: fixed to actually check convertibility via `UnitConverter.ChangeUnits` instead of always returning `false`
- Distinguishes two failure modes in cross-unit arithmetic: `false` from `Try*` means genuinely incompatible units; `NotImplementedException` from `Try*` means the configured service doesn't implement the operation — each produces a distinct `NotSupportedException` message.
- Fixes pre-existing copy/paste bugs in `Subtract` and `Modulo` (both called `Add` internally in their mixed-unit branch).
- Bumps `Fhir.Metrics` from `1.3.0` to `1.4.0` (adds `TryAdd`/`TrySubtract`/`TryConvertTo` support natively) and adds the reference to `Cql.Firely` for the first time.
- Adds `CqlQuantityTests` covering Add, Subtract, Modulo, Divide, Multiply, Between, and CanConvertQuantity with same-unit, calendar-alias, cross-unit, granular-unit, incompatible-dimension, null, and error cases.

## Injecting a custom IMetricService

The default service (`DefaultUcumMetricService`) wraps `FhirMetricService` from the open-source `Fhir.Metrics` package (1.4.0), which now implements `TryAdd`, `TrySubtract`, and `TryConvertTo` natively alongside its existing `TryDivide`/`TryCompare`/`TryCanonicalize` — so cross-unit arithmetic (e.g. `1 kg + 500 g`) works against the default service without needing a custom implementation.

To supply an alternative `IMetricService` implementation via `FhirCqlContextOptions`:

```csharp
// Default: uses the built-in FhirMetricService wrapper
var context = FhirCqlContext.ForBundle(bundle);

// Custom: inject an improved custom IMetricService 
var context = FhirCqlContext.ForBundle(bundle, options: new FhirCqlContextOptions
{
    MetricService = new MyFullUcumMetricService()
});

// Same pattern for WithDataSource:
var context = FhirCqlContext.WithDataSource(source, options: new FhirCqlContextOptions
{
    MetricService = new MyFullUcumMetricService()
});
```

`IMetricService` is defined in the `Fhir.Metrics` NuGet package. 

## Test plan

- [x] `CqlQuantityTests` — all passing, 0 skipped (cross-unit Add/Subtract now exercised against the default `Fhir.Metrics` 1.4.0 service)
- [x] Existing `UcumTests` still pass
- [x] No regressions in `CqlContextOperatorTests` or `ExpressionBuilderTests` (verified via the full `CoreTests` suite: 969/977 passed on net8.0, the 8 skips are pre-existing and unrelated to this PR)


Cql/Cql.Firely/FhirCqlContextOptions.cs
```diff
@@ -65,5 +66,11 @@ public enum CodeInOperatorSemantics
         /// The default <see cref="TypeConverter"/> to use is <see cref="FhirTypeConverter.Create(ModelInspector,Nullable{int})"/>, unless otherwise overriden here.
         /// </summary>
         public TypeConverter? OverrideTypeConverter { get; init; }
+
+        /// <summary>
+        /// The <see cref="IMetricService"/> to use for UCUM unit conversions, or <see langword="null"/> to use the default service.
+        /// Inject a custom implementation here to substitute the commercial Firely UCUM service.
```
ewoutkramer:
```suggestion
        /// Inject a custom implementation here.
```

Cql/Cql.Runtime/Operators/CqlOperators.ArithmeticOperators.cs
```diff
@@ -100,12 +102,28 @@ internal partial class CqlOperators
...
+                {
+                    if (MetricServiceExtensions.TryAdd(MetricService,
+                            (left.value.Value, leftUnit, "http://unitsofmeasure.org"),
+                            (right.value.Value, rightUnit, "http://unitsofmeasure.org"),
+                            out var result))
+                        return new CqlQuantity(result!.Value.Item1, result.Value.Item2);
+                }
+                catch (NotImplementedException)
+                {
+                    throw new NotSupportedException($"The configured IMetricService does not implement Add for units {left.unit} and {right.unit}. Inject a full IMetricService implementation to enable cross-unit arithmetic.");
```
ewoutkramer: I see this code for Add, Subtract and Divide - is there no comparable case for Multiply?  In any case these bits of code look very similar - is there a way to avoid code duplication?

Cql/Cql.Firely/FhirCqlContextOptions.cs
ewoutkramer: I don't want a mention of "the commerical UCUM service", since we have no such product offering - this is just an extension point to supply another implementation.

Cql/CoreTests/CqlQuantityTests.cs
```diff
@@ -78,4 +82,381 @@ public void OperatorNegate_NullQuantity_ReturnsNull()
...
+        var ops = GetOperators();
+        Assert.IsNull(ops.Divide((CqlQuantity?)null, new CqlQuantity(2m, "mg")));
+    }
+
+    #endregion
+
+    #region Multiply
+
+    [TestMethod]
+    public void Multiply_ByScalar_ReturnsProductWithLeftUnit()
```
baseTwo: Fix test name - it says result keeps left unit, but assertion checks dimensionless unit on the right

baseTwo: Fixed

baseTwo: Actually this is still not correct, see https://github.com/FirelyTeam/firely-cql-sdk/pull/1282#pullrequestreview-4448406392

baseTwo (changes requested):
@alexzautke / @ewoutkramer the assertions in the unit tests do not seem correct. I had a chat with Ewout yesterday, and we determined that more work is needed on this PR.

Specific focus is needed on 
- Multiply_ByScalar_ReturnsScalarProduct (previously Multiply_ByScalar_ReturnsProductWithLeftUnit)
  - A scalar product with unit-quantity should also be a unit-quantity
- Divide_ByScalar_ReturnsQuotientWithLeftUnit.
  - When dividing a scalar with a unit-quantity, is it the idea that the si prefix is shifted from mg to g?
  
There is a possible logic error in `CqlOperators.ArithmeticOperators.cs` method `Multiply(CqlQuantity? left, CqlQuantity? right)`
- There is a lot of pre-optimization on cql quantities done here, before it even goes to the ucum binary operator logic. I believe this is the correct logic
```
        public CqlQuantity? Multiply(CqlQuantity? left, CqlQuantity? right)
        {
            if (left == null || right == null)
                return null;
            else if (left.value == null || right.value == null)
                return null;
            else
                return TryUcumBinaryOp(left.value.Value, left.unit!, right.value.Value, right.unit!, MetricServiceExtensions.TryMultiply, "Multiply");
        }
```

baseTwo:
@copilot `Fhir.Metrics` 1.4.0 is now published (https://www.nuget.org/packages/Fhir.Metrics/1.4.0) — please pick this branch up now. This is a sizeable punch list; work through it in order and push as you go, don't try to land it all in one silent commit.

## 1. Merge `develop` in

This branch is 7+ weeks stale and currently shows `CONFLICTING`. Merge `origin/develop` in (a normal merge commit, not a rebase — this isn't a stacked PR). Resolve conflicts by preserving both sides' intent; don't take either side wholesale without checking what the other side actually changed.

## 2. Bump the dependency

Update the `Fhir.Metrics` package reference to `1.4.0` in every `.csproj` that references it (`Cql/Cql.Runtime/Cql.Runtime.csproj`, `Cql/Cql.Firely/Cql.Firely.csproj`, `Cql/CoreTests/CoreTests.csproj`).

## 3. Three real correctness bugs to fix (independent of the version bump — these are bugs in this PR's own code)

**3a. `Multiply(CqlQuantity?, CqlQuantity?)` collapses the unit to `"1"` instead of preserving it.** Current code (`Cql/Cql.Runtime/Operators/CqlOperators.ArithmeticOperators.cs`):
```csharp
else if (left.unit != "1" && right.unit != "1")
    return TryUcumBinaryOp(left.value.Value, left.unit!, right.value.Value, right.unit!, MetricServiceExtensions.TryMultiply, "Multiply");
else
    return new CqlQuantity(Multiply(left.value, right.value), "1");
```
Whenever *either* operand's unit is `"1"`, the result is hardcoded to unit `"1"` — discarding whichever operand carried the real unit. Per the CQL spec (`spec/cql/condensed/09-b-cqlreference.md`, Multiply section): "the resulting quantity will have the appropriate unit... When a quantity has no units specified, it is treated as a quantity with the default unit ('1')" — multiplying by the identity-dimension `'1'` must preserve the *other* operand's unit (`3 'mg' * 4` → `12 'mg'`, not `12 '1'`). Fix so the result keeps the non-`"1"` operand's unit when exactly one side is `"1"` (only collapse to `"1"` when *both* are `"1"`) — or better, just always route through `TryUcumBinaryOp`/`MetricServiceExtensions.TryMultiply` rather than special-casing `"1"` at all, for consistency with how Add/Subtract/Divide are structured. Also fix `Multiply_ByScalar_ReturnsScalarProduct` in `Cql/CoreTests/CqlQuantityTests.cs` — it currently asserts `UCUMUnits.Default` ("1") as the expected result unit, i.e. it asserts the bug as correct. It should assert the preserved unit instead.

**3b. `Between` is dimension-blind — and this is a regression from the pre-PR code.** It delegates to `Comparer.Compare`, whose `CqlQuantityCqlComparer.CompareValues` canonicalizes each operand *independently* via `TryCanonicalize` and then compares only the raw numeric values — it never checks the two canonicalized units are actually the same dimension. Concrete failure: `ops.Between(new CqlQuantity(5m, "kg"), new CqlQuantity(1m, "s"), new CqlQuantity(10m, "kg"))` returns `true`, even though the low bound (1 second) is dimensionally meaningless against a mass argument. The pre-PR code had an explicit guard for exactly this (`if (argument.unit != low.unit && argument.unit != high.unit) throw new NotSupportedException("Only like quantity units are comparable.")`) that was removed in favor of this broken delegation. Fix `CompareValues` so that after both `TryCanonicalize` calls succeed, it verifies the resulting canonical units actually match (or route the cross-unit comparison through `IMetricService.TryCompare` instead, which should do a proper dimension-aware comparison in one step) — throwing/returning null on a dimension mismatch rather than silently comparing incompatible numbers. Add a regression test for `Between` with an argument/bound pair in incompatible dimensions (e.g. kg vs s) asserting it does *not* return a spurious `true`.

**3c. Add/Subtract's cross-unit result unit won't be CQL-spec-compliant once routed through `Fhir.Metrics` 1.4's native `TryAdd`/`TrySubtract`.** The CQL spec is explicit: "When adding quantities... **the unit of the result will be the most granular unit of either input**" (e.g. `1 'km' + 1 'm'` → result in `'m'`, since `'m'` is more granular than `'km'`). `Fhir.Metrics`' `TryAdd`/`TrySubtract` (implemented in 1.4.0) canonicalize both operands to the UCUM *canonical/base* unit and return the result in that unit — which happens to coincide with "most granular" for a km/m pair (both examples in Fhir.Metrics' own test suite only cover that case), but does **not** in general: `1 'mg' + 2 'g'` canonicalizes to grams (discarding the milli- prefix), giving `2.001 'g'`, whereas the spec requires `2001 'mg'` — same real quantity, wrong unit. Our own `TryUcumBinaryOp` currently uses whatever unit `TryAdd`/`TrySubtract` returns verbatim (`new CqlQuantity(result.Item1, result.Item2)`), with no re-expression step. Fix: after getting the canonical-unit result from `TryAdd`/`TrySubtract`, determine which of the two input units is more granular (compare their magnitude — e.g. canonicalize `1 <leftUnit>` and `1 <rightUnit>` and see which is numerically smaller), then use `IMetricService.TryConvertTo` (also implemented in 1.4.0) to re-express the raw result in that more-granular unit before constructing the final `CqlQuantity`. Add tests covering a prefix-mismatched pair in both operand orders (e.g. `1 'mg' + 2 'g'` and `2 'g' + 1 'mg'`, both expecting a result in `'mg'`).

## 4. Fhir.Metrics-1.4-specific cleanup

- **Remove `DefaultUcumMetricService`'s manual `TryConvertTo` workaround** (`Cql/Cql.Runtime/Conversion/UcumConversionExtensions.cs`) — it exists only because `FhirMetricService.TryConvertTo` threw `NotImplementedException` pre-1.4.0. Delegate straight to `_fhirService.TryConvertTo(quantity, targetUnit, out result)`, delete the now-unused `_system` `Lazy<M.SystemOfUnits>` field and the private `ToQuantity` helper, remove the two `TODO(#1296)` comments and the stale XML doc wording, and close issue **#1296** once this lands (its acceptance criteria are exactly this change).
- **Un-ignore the 2 cross-unit Add/Subtract tests** in `Cql/CoreTests/CqlQuantityTests.cs` (currently `[Ignore("Requires a full IMetricService implementation...")]`) — remove the `Ignore(...)` and run them against the real default service (no fake needed now that 1.4.0 implements `TryAdd`/`TrySubtract` natively). Adjust asserted values/units per the 3c fix above (the raw canonical result needs the granularity re-expression, not the bare canonical value). This resolves the two existing unresolved review threads on those lines.
- Update the PR body's "Injecting a custom IMetricService" section — it currently states `TryAdd`/`TrySubtract` "are not implemented" on the default service, which becomes false. Update the Test Plan checklist counts too.

## 5. Two independent loose ends spotted in the same code region (not required for this PR, but flag/fix if convenient — otherwise file a quick follow-up issue)

- `TruncatedDivide(CqlQuantity?, CqlQuantity?)` was never updated to route through `IMetricService`/`TryUcumBinaryOp` like Add/Subtract/Divide/Multiply were — it still hard-throws `NotSupportedException("Unit arithmetic is not supported.")` for any non-`"1"` unit.
- `CompareNormalizedUnits` (bottom of `CqlOperators.ArithmeticOperators.cs`) is now dead code, unreferenced since Add/Subtract/Modulo were rewritten to use `UcumConversionExtensions.AreSameCqlCalendarUnit`.

## 6. Before pushing

- Run the full `UcumTests` + `CqlQuantityTests` suites on both `net8.0` and `net10.0` — confirm no regressions from the delegate swap, the TFM change (Fhir.Metrics 1.4.0 dropped its net8.0-specific build in favor of `netstandard2.0`, low risk but worth confirming), and the three bug fixes.
- Add a `docs/releases/vnext/<slug>.md` release-note fragment (per this repo's fragment-file convention, named after this PR's own number once it's opened/updated — see `docs/releases/vnext/README.md`) covering: cross-unit UCUM Add/Subtract now succeed against the default injected metric service, the three correctness fixes, and the `Fhir.Metrics` 1.4.0 bump.
- No `GeneratorToolVersion` bump needed — no `ICqlOperators` signature or generated-C# shape changes, only internal delegate targets change.

Cql/Cql.Runtime/Operators/CqlOperators.ComparisonOperators.cs
```diff
@@ -46,11 +46,10 @@ internal partial class CqlOperators
...
             if (argument?.value == null || low?.value == null || high?.value == null)
                 return null;
-            else if (argument.unit != low.unit && argument.unit != high.unit)
-                throw new NotSupportedException("Only like quantity units are comparable.");
+            else if (Comparer.Compare(argument, low, null) >= 0 && Comparer.Compare(argument, high, null) <= 0)
+                return true;
             else
-                return Between(argument.value, low.value, high.value);
-
+                return false;
```
baseTwo: **Required fix (blocker):** This trades one wrong answer for another. The spec is explicit that a dimension-incompatible quantity comparison must yield `null`, not `false` — its own worked example is literally named `QuantityBetweenIsNull` (`3.5 'cm2' between 3.0 'cm' and 4.8 'cm'`). Right now `Comparer.Compare(argument, low, null) >= 0 && Comparer.Compare(argument, high, null) <= 0` silently collapses a null (dimension-incompatible) sub-comparison into `false`, because C#'s lifted `>=`/`<=` operators return `false` — not `null` — when an operand is null. That's inconsistent with the sibling `Greater`/`Less`/`GreaterOrEqual`/`LessOrEqual` operators in this same file, which correctly `switch` on `null => null`.

Hand-traced against the original bug report: `Between(5 'kg', 1 's', 10 'kg')` — `Compare(5kg, 1s)` is null (incompatible dimensions, correctly detected), `Compare(5kg, 10kg)` is `-1`. `(null >= 0) && (-1 <= 0)` = `false && true` = `false`. Per three-valued logic, `null AND true` should be `null`, not `false`.

Please rewrite to explicitly check for null before combining, matching the Greater/Less pattern:
```csharp
else
{
    var lowCompare = Comparer.Compare(argument, low, null);
    var highCompare = Comparer.Compare(argument, high, null);
    if (lowCompare == null || highCompare == null)
        return null;
    else
        return lowCompare >= 0 && highCompare <= 0;
}
```

Cql/CoreTests/CqlQuantityTests.cs
```diff
@@ -78,4 +82,428 @@ public void OperatorNegate_NullQuantity_ReturnsNull()
...
+    {
+        var ops = GetOperators();
+        Assert.IsTrue(ops.Between(new CqlQuantity(500m, "g"), new CqlQuantity(0.4m, "kg"), new CqlQuantity(1m, "kg")));
+    }
+
+    [TestMethod]
+    public void Between_CqlQuantity_IncompatibleDimensions_ReturnsFalse()
+    {
+        var ops = GetOperators();
+        Assert.IsFalse(ops.Between(new CqlQuantity(5m, "kg"), new CqlQuantity(1m, "s"), new CqlQuantity(10m, "kg")));
```
baseTwo: **Required fix (part of the same blocker above):** Per the spec's `QuantityBetweenIsNull` example, this should assert `Assert.IsNull(...)`, not `Assert.IsFalse(...)` — right now it locks in the incorrect `false` behavior instead of catching it as a regression. Please rename the test too (e.g. `Between_CqlQuantity_IncompatibleDimensions_ReturnsNull`) once the code fix lands.

docs/releases/vnext/1282.md
```diff
@@ -0,0 +1,6 @@
+# PR #1282
```
baseTwo: **Required fix:** This fragment doesn't follow the conventions documented in `docs/releases/vnext/README.md` (adopted repo-wide just before this PR started):

1. **Naming**: should be `<PR-number>-<short-slug>.md` (e.g. `1282-ucum-metric-service-injection.md`), not a bare `1282.md` — every sibling fragment in this directory follows the slug pattern.
2. **Format**: should use `##` category headings matching `vnext-release-notes.md`'s own structure (`## Fixes`, `## Features`, etc.) with each bullet ending in `(#1282)` — not an `# PR #1282` H1 with a flat, unattributed bullet list.
3. **Completeness**: it's missing the `DefaultUcumMetricService.TryConvertTo` workaround removal (a real, substantive part of this PR, tied to closing #1296) — please add a bullet for it.

Something like:
```md
## Features

- `Hl7.Cql.Fhir.FhirCqlContextOptions` now accepts a custom `IMetricService` implementation for UCUM unit arithmetic (`MetricService`), and cross-unit `Add`/`Subtract` work against the default injected service, returning results in the most granular of the two input units per §9.B (#1282).

## Fixes

- Quantity `Multiply` by a scalar/unitless quantity now preserves the non-`'1'` unit instead of collapsing the result to `'1'` (#1282).
- Quantity `Between` now returns `null` (not an incorrect `true`) when the argument and bounds have incompatible dimensions (#1282).
- `DefaultUcumMetricService.TryConvertTo`'s temporary workaround is removed now that `Fhir.Metrics` 1.4.0 implements it natively; the `Fhir.Metrics` dependency is bumped from `1.3.1` to `1.4.0` (#1282, closes #1296).
```
(adjust wording/section split as you see fit — the point is matching the repo's established fragment shape, not this exact text).

Cql/Cql.Runtime/Operators/CqlOperators.ArithmeticOperators.cs
```diff
@@ -517,10 +523,16 @@ internal partial class CqlOperators
                 return null;
             else if (left.value == null || right.value == null)
                 return null;
-            else if (left.unit != "1" && right.unit != "1")
-                throw new NotSupportedException("Unit arithmetic is not supported.");
+            else if (left.unit == null || right.unit == null)
+                return null;
+            else if (left.unit == UCUMUnits.Default && right.unit == UCUMUnits.Default)
```
baseTwo, line 537: **Required test-coverage fix (minor, per this repo's standing convention that coverage gaps are required, not nice-to-have):** The new both-operands-unit-`"1"` branch (added by this commit) has no test — I confirmed it's actually correct (`3 '1' * 4 '1'` → `12 '1'`) by adding and running a throwaway test, but nothing in the committed suite exercises it. Please add a test for this case alongside the existing scalar-multiplication tests.

baseTwo (dismissed):
@copilot Great progress — I independently verified everything by building and actually running the tests, not just reading the diff. Confirmed fully correct and complete: the Add/Subtract most-granular-unit fix (hand-traced 1 'mg' + 2 'g' and the reverse order, both give 2001 'mg' as the spec requires; full test suite ran and passed), the Fhir.Metrics 1.4.0 bump, and the `DefaultUcumMetricService.TryConvertTo` workaround removal (issue #1296's acceptance criteria are now fully met — I'll close it once this merges). No regressions in the full `CoreTests` suite (969/977 passed on net8.0, the 8 skips are pre-existing and unrelated, confirmed identical on develop).

Two things still need work before this is mergeable:

1. **`Between`'s dimension-check fix is incomplete** — it now returns `false` for incompatible dimensions instead of the CQL-spec-mandated `null`.
2. **The release-note fragment doesn't follow this repo's own conventions** (naming, format, and completeness).

Details and exact fixes below. Also one required test-coverage addition.

Cql/Cql.Runtime/Operators/CqlOperators.ComparisonOperators.cs
```diff
@@ -46,11 +46,10 @@ internal partial class CqlOperators
         {
             if (argument?.value == null || low?.value == null || high?.value == null)
                 return null;
-            else if (argument.unit != low.unit && argument.unit != high.unit)
-                throw new NotSupportedException("Only like quantity units are comparable.");
+            else if (Comparer.Compare(argument, low, null) >= 0 && Comparer.Compare(argument, high, null) <= 0)
+                return true;
```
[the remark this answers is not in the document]
baseTwo: Partially addressed in the latest commits (the spurious `true` for incompatible dimensions is gone), but not fully — it now returns `false` where the spec requires `null`. See my new inline comment on the current `Between` implementation for the exact fix needed. Leaving this open until that lands.

docs/releases/vnext/1282.md
baseTwo: Naming, format, and the TryConvertTo-removal bullet are all correct now — thanks. Two small things left in the new `1282-ucum-metric-service-injection.md` before this is done:

1. **Wrong version number**: the bullet says `Fhir.Metrics` went from `1.3.1` to `1.4.0`, but `Cql.Runtime.csproj` at this PR's base actually pinned `1.3.0` (confirmed against the base commit and the diff). Please correct to `1.3.0 -> 1.4.0`, and it's also worth noting this PR adds the `Fhir.Metrics` reference to `Cql.Firely.csproj` for the first time (it wasn't there before).
2. **Missing a real fix**: `CqlOperators.CanConvertQuantity` (in `CqlOperators.TypeOperators.cs`) previously always returned `false` unconditionally for any non-null argument/unit — a genuine bug, now fixed to actually check convertibility via `UnitConverter.ChangeUnits` (and covered by this PR's own new `CanConvertQuantity_*` tests). This is a real, user-visible fix bundled in this PR and isn't mentioned anywhere in the fragment — please add a `## Fixes` bullet for it.

baseTwo:
@copilot One thing left before this is mergeable — the release-note fragment `docs/releases/vnext/1282-ucum-metric-service-injection.md` still has two issues (see the unresolved review thread on that file for full context):

1. **Wrong version number**: it says `Fhir.Metrics` went from `1.3.1` to `1.4.0`. The actual base version in `Cql.Runtime.csproj` before this PR was `1.3.0`. Please correct the bullet to `1.3.0 -> 1.4.0`, and mention that this PR also adds the `Fhir.Metrics` reference to `Cql.Firely.csproj` for the first time (it wasn't referenced there before).
2. **Missing a real fix**: `CqlOperators.CanConvertQuantity` (in `CqlOperators.TypeOperators.cs`) previously always returned `false` unconditionally for any non-null argument/unit — a genuine bug. This PR fixes it to actually check convertibility via `UnitConverter.ChangeUnits`, and it's covered by this PR's own new `CanConvertQuantity_*` tests. Please add a `## Fixes` bullet for this — it's real, user-visible behavior bundled in this PR that isn't mentioned anywhere in the fragment right now.

Everything else (the `Between` null-semantics fix, the Multiply both-scalar test, the Fhir.Metrics 1.4.0 bump, the `TryConvertTo` workaround removal) has been independently verified and is good. This is the last item.

baseTwo:
Do not approve yet

Cql/Cql.Runtime/Operators/CqlOperators.ArithmeticOperators.cs
```diff
@@ -815,16 +827,92 @@ internal partial class CqlOperators
...
+                }
+            }
+            catch (NotImplementedException)
+            {
+                throw new NotSupportedException(
+                    $"The configured IMetricService does not implement {opName} for units {leftUnit} and {rightUnit}. Inject a full IMetricService implementation to enable cross-unit arithmetic.");
+            }
+
+            throw new NotSupportedException(
+                $"Arithmetic on quantities with incompatible units {leftUnit} and {rightUnit} is not supported.");
```
baseTwo: **Blocker:** `TryUcumBinaryOp` — which `Add`, `Subtract`, `Multiply`, and `Divide` all route through for cross-unit operands — throws `NotSupportedException` here whenever the units are genuinely incompatible (e.g. `1 'kg' + 1 's'`). Per CQL spec §9.B, this must return `null`, not throw: the spec's Add/Subtract descriptions say verbatim "units of 'cm2' and 'cm' cannot [be added]... Attempting to operate on quantities with invalid units will result in a null."

This PR itself already fixes the *identical* scenario correctly for `Between`/comparisons (`CqlQuantityCqlComparer.CompareValues` now returns `null` for incompatible dimensions, with a comment quoting this same spec passage) — but left `Add`/`Subtract`/`Multiply`/`Divide` throwing, which is inconsistent within this same PR and violates the spec for four operators at once.

Please change this to return `null` instead of throwing when the units are incompatible (keep the existing `catch (NotImplementedException)` branch as-is — that one's a different, correctly-thrown case: "the configured service doesn't implement this op at all", not "these units can't be combined"). Update `Add_IncompatibleUnits_Throws` and `Subtract_IncompatibleUnits_Throws` (currently asserting `[ExpectedException(typeof(NotSupportedException))]`) to assert `Assert.IsNull(...)` instead — they currently lock in the wrong behavior. Please also add the equivalent tests for `Multiply` and `Divide` with incompatible dimensions, since none exist for those two.

Cql/Cql.Runtime/Operators/CqlOperators.ArithmeticOperators.cs
```diff
@@ -481,12 +488,11 @@ internal partial class CqlOperators
...
-                string? leftUnit = left.unit;
-                string? rightUnit = right.unit;
-                CompareNormalizedUnits(leftUnit, rightUnit);
-
-                return new CqlQuantity(Add(left.value, right.value), leftUnit);
+                string leftUnit = left.unit ?? string.Empty;
+                string rightUnit = right.unit ?? string.Empty;
+                if (UcumConversionExtensions.AreSameCqlCalendarUnit(leftUnit, rightUnit))
+                    return new CqlQuantity(Modulo(left.value, right.value), left.unit);
+                throw new NotSupportedException($"Modulo of quantities with incompatible units {left.unit} and {right.unit} is not supported.");
```
baseTwo: **Major:** `Modulo`'s cross-unit branch was never migrated to the `TryUcumBinaryOp`/`IMetricService` pattern its siblings (`Add`, `Subtract`, `Multiply`, `Divide`) now use — it only special-cases calendar aliases (day/days) and otherwise unconditionally throws, even for a straightforwardly commensurable pair like `7000 'g' % 3 'kg'`. Please migrate it the same way (delegate to `TryUcumBinaryOp`/`MetricServiceExtensions.TryModulo` if `Fhir.Metrics` 1.4.0 has one, or the closest equivalent operation available — check what's actually exposed on `IMetricService`), including the null-for-incompatible-dimensions fix above. Add tests mirroring the sibling operators' coverage: a commensurable-cross-unit case (e.g. kg/g) and an incompatible-dimension case (e.g. kg/s, expecting `null`).

Cql/Cql.Runtime/Operators/CqlOperators.ArithmeticOperators.cs
```diff
@@ -915,33 +1003,6 @@ internal partial class CqlOperators
             else
                 return new CqlQuantity(TruncatedDivide(left.value, right.value), "1");
```
baseTwo: **Major:** `TruncatedDivide(CqlQuantity?, CqlQuantity?)` still has the old pre-PR guard (`left.unit != "1" && right.unit != "1"` → throw) and has real bugs, independent of whether it's migrated:

1. It throws for **same-unit** operands (e.g. `6 'mg' / 2 'mg'`), where sibling `Divide` correctly returns `3` with unit `'1'`.
2. For a **scalar divisor** (`right.unit == '1'`), it hardcodes the result unit to `'1'` regardless of the left unit, contradicting `Divide`'s behavior of preserving the left unit in that case (see `Divide_ByScalar_ReturnsQuotientWithLeftUnit`).
3. It has **zero test coverage** anywhere in the repo — the only other `TruncatedDivide` hits are ELM-parsing tests in `CqlToElmTests`, which never execute this runtime method at all.

This was flagged as an optional follow-up in an earlier review round, but given the scope of the bugs found now (not just "never migrated", but actively wrong for the same-unit and scalar-divisor cases), please fix it as part of this PR: mirror `Divide`'s branching structure (same-unit → `'1'`, scalar-divisor → preserve left unit, cross-unit → delegate to `TryUcumBinaryOp`/`TryDivide`) and apply `decimal.Truncate`/`TruncatedDivide(decimal, decimal)` to the final numeric result. Add a full test region matching `Divide`'s (same-unit, scalar-divisor, commensurable-cross-unit, incompatible-dimension, null-argument).

Cql/Cql.Runtime/Operators/CqlOperators.TypeOperators.cs
```diff
@@ -217,8 +217,18 @@ internal partial class CqlOperators
 
         #region CanConvertQuantity
 
-        public bool? CanConvertQuantity(CqlQuantity? argument, string? unit) =>
-            argument == null || unit == null ? null : false;
+        public bool? CanConvertQuantity(CqlQuantity? argument, string? unit)
+        {
+            if (argument == null || unit == null) return null;
```
baseTwo: **Minor (real regression introduced by this PR):** This only guards `argument == null || unit == null` before calling `UnitConverter.ChangeUnits(argument, unit)`. But a non-null `CqlQuantity` can still have a null `.value` (this repo's established representation of a "null quantity", e.g. from `Negate`/`Add`/`Subtract` on overflow) — `ChangeUnits` returns `null` for that case, and `null is not null` evaluates to `false`, so `CanConvertQuantity(new CqlQuantity(null, "kg"), "g")` returns `false` instead of `null`. Its sibling `ConvertQuantity` (a few lines below, unchanged by this PR) already guards this exact case (`argument == null || argument.value == null || unit == null`). Please add the same `argument.value == null` check here, and add a test for it (a non-null `CqlQuantity` with a null value).

Cql/Cql.Runtime/Conversion/UcumConversionExtensions.cs
```diff
@@ -1,11 +1,12 @@
-﻿/*
+/*
  * Copyright (c) 2023, Firely (info@fire.ly) and contributors
  * See the file CONTRIBUTORS for details.
  *
  * This file is licensed under the BSD 3-Clause license
- * available at https://raw.githubusercontent.com/FirelyTeam/firely-net-sdk/master/LICENSE
+ * available at https://raw.githubusercontent.com/FirelyTeam/firely-cql-sdk/master/LICENSE
```
baseTwo: **Nit:** This changes the LICENSE URL in this pre-existing file's copyright header from `firely-net-sdk` to `firely-cql-sdk`. Per this repo's convention, an existing file's header should never be modified (even a factually-incomplete/incorrect one) — please revert this specific line back to `firely-net-sdk` as an unrelated, incidental edit (the rest of this file's changes are all in scope and fine).

baseTwo (changes requested):
@copilot I did a final, holistic pass over the entire PR (not just the specific fixes from the last two rounds) before approving, since this touches core arithmetic correctness. It surfaced one genuine blocker and a few gaps that are within scope of this PR's own rework — all verified against the actual code, several confirmed by actually running/tracing the code. Details below, ordered by severity. Also two small non-blocking items to note.

baseTwo (dismissed):
All 5 items from the final round independently verified (read the actual code + built and ran the full test suite in an isolated worktree, not just the diff):

- `TryUcumBinaryOp` now returns `null` for incompatible dimensions (spec-compliant), with `Add`/`Subtract` tests correctly renamed/updated and new `Multiply`/`Divide` incompatible-unit tests added.
- `Modulo` now handles commensurable cross-unit operands via conversion and returns `null` for genuinely incompatible ones, with both cases tested.
- `TruncatedDivide` now correctly mirrors `Divide`'s branching (same-unit, scalar-divisor, cross-unit) with truncation applied, and has a full test region where there was previously none.
- `CanConvertQuantity` now correctly returns `null` for a non-null quantity with a null value.
- The incidental header URL change is reverted.

Full regression run: `CqlQuantityTests`+`UcumTests` 91/91 passed (net8.0 and net10.0), full `CoreTests` 980 passed/8 pre-existing skips, full `CqlToElmTests` 2497 passed/33 pre-existing skips — zero failures anywhere.

One minor, pre-existing (not introduced by this PR) inconsistency noted for a possible follow-up: `TruncatedDivide`'s cross-unit branch represents a dimensionless result as `string.Empty` while the same-unit branch uses `UCUMUnits.Default` (`"1"") — inherited from `Divide`'s existing behavior, doesn't affect the correctness of what's under review here.

This is thorough, well-tested work across several rounds of increasingly deep review. Approving.

baseTwo:
@copilot Correction — please do NOT add the dependency-diagram work (issue #1436) to this PR/branch. Create a **new branch off `feature/ucum-metric-service-injection`** (so you're working from this PR's not-yet-merged code, which is what introduces the dependency the diagram needs to show) and open a **separate PR** for issue #1436 targeting `develop` as normal. This PR (#1282) should stay scoped to just the UCUM/IMetricService fixes already reviewed and approved here — don't add any further commits to it for the diagram work.

alexzautke (approved):

baseTwo:
@copilot This PR's branch just got a new commit (`6c799c248`, "Apply remaining changes") that adds `docs/dependency-diagrams.md`/`dependency-diagrams.runtime-services.mmd`/`.svg` — an earlier draft of the diagram from #1437's work, apparently committed here by accident (likely a side effect of creating the stacked PR). This PR was already reviewed and approved scoped to just the UCUM/`IMetricService` fixes — please **revert commit `6c799c248`** from this branch entirely. The diagram work should live only in #1437 (which already has the complete, final version in its own commit `463fd495`) — this also appears to be causing #1437's current merge conflict, since both branches now have files at the same paths.

PR #1323, opened 2026-07-06
Optimize RunLibrary/SelectResults hot path: value set fast path, direct-delegate invoker, cache tuning (-40%)

### LINKED ISSUE #1322: Perf: RunLibrary/SelectResults hot path — sync-over-async value set checks dominate evaluation CPU (~48%)
# Summary

CPU profiling of the CQL evaluation hot path (`CqlLibraryRunner.RunLibrary` → `DefinitionInvokerExtensions.SelectResults`) shows that roughly half of all evaluation CPU time is spent in avoidable overhead in the value set lookup path, with additional fixed costs in reflection-based definition invocation and the definition memoization cache. The findings below are ordered by expected performance gain, largest first.

# Profiling Setup

- Harness: IntegrationRunner benchmarks (`--direct-runlibrary` mode) running only `FhirCqlContext.ForBundle` + `RunLibrary` in a loop — bundle loading, test discovery, and DI setup excluded from the measured loop.
- Case: `CMS156FHIRHighRiskMedsElderly` test case `8b33d091` (slowest warm case: ~2 ms median / ~523 KB allocated per iteration), 30,000 iterations.
- Tooling: `dotnet-trace` with `Microsoft-DotNETCore-SampleProfiler`, speedscope conversion, custom subtree analysis rooted at `CqlLibraryRunner.RunLibrary`.
- Percentages below are the share of all sampled CPU time with `RunLibrary` on the stack (5.0 s of samples total).

# Findings and Recommended Fixes (Max Gain First)

## 1. `ValueSetSource.CheckInternalAndExternalTs` — sync-over-async on every code check (~48%)

**Observed Behavior**

`Cql/Cql.Firely/ValueSetSource.cs` (`CheckInternalAndExternalTs`) calls `TaskHelper.Await(() => Load(valueSetUri))` for **every single code membership check**. `TaskHelper.Await` (Firely SDK) wraps the call in `Task.Run(...)` and blocks, so even a 100% warm cache hit pays for: lambda allocation → thread pool dispatch → `Task.InternalWait` (44% of samples) → spin/block (`Task.SpinWait` 26%, `Monitor.Enter_Slowpath` 22%). All of that to perform what is ultimately a `ConcurrentDictionary.TryGetValue`.

In the integration runner scenario the value sets are always preloaded via `Add(...)`, so the async resolver path is never actually needed at check time — the overhead is pure waste.

**Root Cause**

No synchronous fast path for the cache-hit case; the async `Load` is invoked unconditionally through `TaskHelper.Await`.

**Recommended Fix**

Add a synchronous fast path in `CheckInternalAndExternalTs` (and `TryGetCodesInValueSet`) that checks `_valueSets.TryGetValue(valueSetUri, out var cached)` before falling back to `TaskHelper.Await(() => Load(...))`:

```csharp
private bool CheckInternalAndExternalTs(
    string valueSetUri,
    Predicate<IValueSetFacade> @internal,
    Action<ValidateCodeParameters> external)
{
    // Fast path: warm cache hit needs no async machinery.
    if (_valueSets.TryGetValue(valueSetUri, out var cached))
    {
        if (@internal(cached)) return true;
    }
    else if (TaskHelper.Await(() => Load(valueSetUri)) is { } vs && @internal(vs))
    {
        return true;
    }
    // ... existing external terminology service fallback ...
}
```

**Expected Gain**: Removes ~40%+ of evaluation CPU for warm value sets (the normal case). Also reduces allocation pressure (lambda + Task per check).

## 2. `IndexedBundle.FilterByType` retrieve path — repeated `ToCodings()` and per-coding value set round trips (~54% inclusive; mostly contains item 1)

**Observed Behavior**

`Cql/Cql.Firely/IndexedBundle.cs` `FilterByType<T>(Predicate<Coding> filter)` re-enumerates all resources of the requested type for every retrieve, calls `ICoded.ToCodings()` (allocating) per candidate, and invokes the value set predicate per coding — each predicate invocation being the expensive `IsCodeInValueSet` path from item 1.

**Recommended Fix**

After fixing item 1 (which removes most of this cost), consider as a follow-up:

- Cache the `ToCodings()` result per bundle entry inside `IndexedBundle` (built once per bundle, reused across retrieves and definitions).
- In `BundleDataSource.RetrieveByValueSet`, resolve the value set to its `InMemoryValueSet` facade once per retrieve and test codings against it directly, instead of going through `IsCodeInValueSet(string, string, string)` per coding.

**Expected Gain**: Secondary; reduces allocations (523 KB/iteration measured) and removes redundant coding extraction across repeated retrieves.

## 3. Reflection-based definition invocation (~30%)

**Observed Behavior**

`DefinitionInvoker.Invoke` dispatches through `RuntimeMethodInfo.Invoke` / `MethodBaseInvoker.InvokeWithOneArg`, with `CustomAttribute.GetCustomAttributes` frames appearing at runtime inside the hot loop. This is a fixed per-definition-invocation cost paid for every definition of every library in `SelectResults`.

**Options Considered**

- **Option A — generated `InvokeDefinition` switch per library**: generator emits a method on each library class that switches on the definition name and calls the definition directly. Zero reflection and AOT-friendly, but **rejected for now**: it requires a `GeneratorToolVersion` bump (new invoker version + regenerating all checked-in `*.g.cs`), and a switch on name alone cannot dispatch overloaded CQL function definitions (`DefinitionSignature` = name + parameter types).
- **Option B — compiled delegate cached per `DefinitionInvoker` (chosen, do now)**: replace `MethodInfo.Invoke` in `DefinitionInvoker_5_0` (`Cql/Cql.Invocation/Toolkit/Internal/LibraryInvoker.5.0.cs`) with an expression-tree-compiled `Func<CqlContext, object?[], object?>` built once per definition at invoker construction. No codegen change, no version bump, benefits all existing generated assemblies immediately; delegate invocation is near direct-call speed.
- **Option C — generated static dispatch table (planned follow-up)**: generator emits a signature-keyed table (e.g. `FrozenDictionary<DefinitionSignature, Func<CqlContext, object?[], object?>>`) on each library class; the invoker consumes the table. Same runtime characteristics as A but handles overloads naturally. Requires the `GeneratorToolVersion` bump and regeneration, so deferred to a follow-up after Option B lands.

**Decision**: Implement **Option B** in this ticket; pursue **Option C** as a follow-up when a generator version bump is warranted. Option A was considered but rejected in favor of C for the generated-code evolution.

**Expected Gain**: Removes reflection dispatch overhead (~30% inclusive share, partially overlapping with definition execution itself).

## 4. Definition memoization cache — `ConcurrentDictionary<long, object?>` write overhead (~22%)

**Observed Behavior**

The per-context definition cache (`Cql/Cql.Runtime/Runtime/CqlContext.Internals.cs`, `_cache`) shows `ConcurrentDictionary.TryAddInternal` with nested `Monitor.Enter` at ~22% inclusive. Evaluation of a single context is effectively single-threaded, so the locking in the concurrent dictionary is pure overhead.

**Recommended Fix**

Use a plain `Dictionary<long, object?>` (or pre-sized dictionary) for the per-run cache when evaluation is single-threaded, or make the cache implementation pluggable so single-threaded scenarios avoid lock costs.

**Expected Gain**: Cuts locking/CAS overhead of memoization writes; smaller but measurable.

## 5. `CqlOperators.Union` (~36%, largely inherent)

**Observed Behavior**

`UnionIterator.ToList` (36%), `CqlOperators.Union` (26%), and `HashSet.UnionWith`/`FillSet` (10%) reflect CQL union semantics using `DelegatedEqualityComparer`. This is real semantic work, not overhead.

**Recommended Fix**

No action for now. Re-profile after items 1–4; only revisit if it remains a top cost (possible avenues: comparer hash quality, pre-sizing hash sets).

# GC / Allocation Pressure

`PollGCWorker` frames are visible throughout the trace, consistent with the measured 523 KB allocated per iteration. Items 1 and 2 (lambda + `Task` per code check, `ToCodings()` per candidate per retrieve) are the main allocation sources and their fixes will reduce GC pressure as a side effect.

# Acceptance Criteria

- [ ] `ValueSetSource` has a synchronous fast path for cached value sets; no `Task.Run`/blocking wait on warm cache hits.
- [ ] Warm per-iteration median for the CMS156 benchmark case improves measurably (re-run `--direct-runlibrary` profile to confirm the `CheckInternalAndExternalTs` share drops).
- [ ] Follow-up items (coding caching in `IndexedBundle`, compiled definition invoker delegates, single-threaded memoization cache) evaluated and either implemented or ticketed separately based on the re-profile.
- [ ] Integration runner tests still pass (no behavioral change to value set resolution semantics, including the external terminology service fallback).

# Related

- Benchmark harness: FirelyTeam/Firely.Cql.Sdk.Integration.Runner#98 (BenchmarkDotNet + `--direct-runlibrary` profiling mode)
- SDK PR for benchmark support work: #1321

baseTwo:
Addresses #1322 — implements the top-ranked fixes from the RunLibrary/SelectResults hot-path profiling investigation.
Related IntegrationRunner PR https://github.com/FirelyTeam/Firely.Cql.Sdk.Integration.Runner/pull/99

### Issue item 1 — `ValueSetSource` synchronous fast path (~48% of evaluation CPU removed)

`CheckInternalAndExternalTs` called `TaskHelper.Await(() => Load(valueSetUri))` for **every code-membership check**. `TaskHelper.Await` wraps the call in `Task.Run(...)` and blocks, so even a 100% warm cache hit paid for a lambda allocation, a thread-pool dispatch, `Task.InternalWait`, and spin/lock waits — all to perform what is ultimately a `ConcurrentDictionary.TryGetValue`.

`CheckInternalAndExternalTs` and `TryGetCodesInValueSet` now check `_valueSets.TryGetValue` synchronously before falling back to the async `Load` path. The external terminology service fallback and resolver semantics are unchanged.

### Issue item 3 — Direct-delegate definition invocation (Option B2, ~30% reflection overhead removed)

`DefinitionInvoker_5_0.Invoke` dispatched through `MethodInfo.Invoke` (reflection) for every definition invocation, including an args-array copy per call.

It now binds a strongly-typed delegate once per definition at invoker construction:

- For the common shapes (`CqlContext` + 0–4 typed arguments, non-void return) it uses `MethodInfo.CreateDelegate` through small generic `WrapArityN` adapters — a direct call, no IL generation, AOT/trimming friendly.
- Other shapes fall back to a compiled expression tree.
- Exceptions still propagate unwrapped, matching the previous `BindingFlags.DoNotWrapExceptions` behavior — including during adapter construction itself, where the `WrapArityN` binder is invoked via the `MethodBase.Invoke` overload with `BindingFlags.DoNotWrapExceptions` so a `CreateDelegate` failure for an unexpected method shape isn't masked by a `TargetInvocationException`.

Design decision (documented in #1322, item 3): a generated `InvokeDefinition` switch per library (Option A) was considered but rejected — it requires a `GeneratorToolVersion` bump and cannot dispatch overloaded CQL functions by name alone. A generated signature-keyed dispatch table (Option C) is the planned follow-up when a generator version bump is warranted.

### Issue item 4 — Memoization cache tuning (~2%)

`CqlContext.UseNewCache()` created a `ConcurrentDictionary<long, object?>` with default settings (lock per CPU, capacity 31). A fresh cache is created per evaluation, so every definition is a cache write, and growth resizes acquire all internal locks. The cache is now created with `concurrencyLevel: 1` and pre-sized to 1024 entries by default. The thread-safety contract is unchanged.

`UseNewCache` now also accepts an optional `initialCapacity` parameter so callers with library sets that are much larger or smaller than the default assumption can tune the pre-sizing themselves, with a public `MinimumCacheInitialCapacity` (16) enforced to guard against values too small to benefit from pre-sizing at all.

## Performance Results

Benchmark: `IntegrationRunner.Benchmarks --direct-runlibrary 8b33d091 <N>` (CMS156FHIRHighRiskMedsElderly, slowest warm case), Release, net10.0, same machine/session. Each iteration = `FhirCqlContext.ForBundle` + `CqlLibraryRunner.RunLibrary` with the bundle preloaded outside the loop. Comparisons use equal iteration counts.

### 30,000 iterations (steady state)

| Step | Change | Total | ms/iter | vs baseline |
|---|---|---|---|---|
| 0 | Baseline (develop) | 8.054 s | 0.268 | — |
| 1 | ValueSetSource fast path + CreateDelegate invoker | 4.898 s | 0.163 | **-39.2% (1.64×)** |
| 2 | + cache tuning | 4.790 s | 0.160 | **-40.3% (1.68×)** |

### 10,000 iterations (includes proportionally more JIT warmup)

| Step | Change | Total | ms/iter | vs baseline |
|---|---|---|---|---|
| 0 | Baseline | 4.367 s | 0.437 | — |
| 1 | ValueSetSource fast path + CreateDelegate invoker | 2.515 s | 0.251 | **-42.6% (1.74×)** |
| 2 | + cache tuning | 2.560 s | 0.256 | -41.4% (within noise of step 1) |

### Profile before/after (CPU samples under `CqlLibraryRunner.RunLibrary`, 30k-iteration traces)

| Cost | Baseline share | After |
|---|---|---|
| `ValueSetSource.CheckInternalAndExternalTs` (sync-over-async) | ~48% | not visible |
| Reflection `MethodBase.Invoke` dispatch | ~30% | near-zero (adapter frames only) |
| `IndexedBundle.FilterByType` + `ToCodings` | ~54% incl. (mostly the value set checks) | ~12% |
| `ConcurrentDictionary` cache writes + `Monitor` | ~22% | reduced (residual overlaps GC poll) |

After these changes the profile is dominated by genuine CQL semantic work (`SelectResults` enumeration/`ToList`, `CqlOperators.Where`/`Union`, generated definition bodies) rather than infrastructure overhead.

## Validation

- IntegrationRunner: **2369 passed, 0 failed** (1488 skipped = not in passes file) — identical to baseline.
- CoreTests: **715 passed, 0 failed, 9 skipped** — identical to baseline; added coverage in `CacheTest.cs` for the new `UseNewCache(initialCapacity)` parameter (default/custom/minimum-boundary/too-small cases).
- Profiling methodology: `dotnet-trace` (`Microsoft-DotNETCore-SampleProfiler`) + speedscope subtree analysis rooted at `CqlLibraryRunner.RunLibrary`.

## Not implemented in this PR (tracked in #1322)

- **Issue item 2** — `IndexedBundle` coding caching / per-retrieve value set facade resolution: not implemented here; residual ~12% after the item 1 fast path landed (down from ~54% inclusive, most of which was item 1's overhead flowing through this path). Confirmed with a fresh, independent re-profile on the current branch state (30k-iteration `--direct-runlibrary` trace, same CMS156 case): total sampled CPU under `RunLibrary` dropped from 5.0 s (baseline) to 3.0 s, of which `IndexedBundle.FilterByType`'s enumeration (inclusive of `ICoded.ToCodings()`) accounts for ~13% (0.4 s) — consistent with the original ~12% estimate and not affected by the item 3/4 changes. `ValueSetSource.CheckInternalAndExternalTs` no longer registers in the profile at all, confirming the item 1 fast path holds up. Deferred pending a dedicated implementation pass.
- **Issue item 3, Option C** — generated signature-keyed dispatch table on library classes: not implemented here (Option B2 was implemented instead, see above). Deferred until a `GeneratorToolVersion` bump is warranted.
- **Issue item 5** — `CqlOperators.Union` cost (~36% of baseline samples): not implemented; the issue marks this as largely inherent CQL semantic work, not overhead, and recommends no action unless it remains a top cost after items 1–4 are re-profiled.

---

## Auxiliary Work

- Updated the `Firely.Cql.Sdk.Integration.Runner` submodule pointer to include the `--direct-runlibrary` profiling harness used to produce the measurements above (FirelyTeam/Firely.Cql.Sdk.Integration.Runner#99).

## Review Feedback Addressed

- **Copilot**: adapter construction (`CreateInvoker`'s `WrapArityN` binder call) could mask exceptions in a `TargetInvocationException`. Fixed by invoking the binder via the `MethodBase.Invoke` overload with `BindingFlags.DoNotWrapExceptions`.
- **@alexzautke**: requested the memoization cache's initial capacity be configurable rather than a fixed constant. Added an optional `initialCapacity` parameter to `UseNewCache` (default unchanged) plus a validated `MinimumCacheInitialCapacity` floor, with unit test coverage.

### DIFF
--- a/Cql/Cql.Runtime/Runtime/CqlContext.cs
+++ b/Cql/Cql.Runtime/Runtime/CqlContext.cs
@@ -99,25 +99,44 @@ namespace Hl7.Cql.Runtime
         /// <summary>
         /// Invalidates the current cache, forcing subsequent operations to use fresh data.
         /// </summary>
+        /// <param name="initialCapacity">The initial capacity of the definition/expression memoization cache.
+        /// Sizing this to (an upper bound of) the number of definitions/expressions expected to be cached during
+        /// the evaluation avoids internal resizing of the cache. Must be at least <see cref="MinimumCacheInitialCapacity"/>.
+        /// Defaults to <see cref="CacheInitialCapacity"/>.</param>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="initialCapacity"/> is less than
+        /// <see cref="MinimumCacheInitialCapacity"/>.</exception>
         /// <remarks>Call this method to clear any cached data and ensure that future operations do not use stale
         /// information. This is useful when the underlying data source has changed and the cache needs to be
         /// refreshed.</remarks>
-        public void UseNewCache()
+        public void UseNewCache(int initialCapacity = CacheInitialCapacity)
         {
+            if (initialCapacity < MinimumCacheInitialCapacity)
+                throw new ArgumentOutOfRangeException(
+                    nameof(initialCapacity),
+                    initialCapacity,
+                    $"The initial cache capacity must be at least {MinimumCacheInitialCapacity}.");
+
             // A fresh cache is populated by mostly-sequential definition evaluation, so use a single
             // lock (concurrencyLevel: 1) and pre-size it to avoid repeated resizes, which acquire
             // all internal locks and dominated the write cost with the default settings.
-            _cache = new ConcurrentDictionary<long, object?>(concurrencyLevel: 1, capacity: CacheInitialCapacity);
+            _cache = new ConcurrentDictionary<long, object?>(concurrencyLevel: 1, capacity: initialCapacity);
             _cacheCallCount = 0;
             _cacheFactoryInvocations = 0;
         }
 
         /// <summary>
-        /// The initial capacity of the definition/expression memoization cache. Sized to hold the
+        /// The default initial capacity of the definition/expression memoization cache. Sized to hold the
         /// cached results of a typical measure evaluation (all definitions across all libraries in
         /// a library set) without internal resizing.
         /// </summary>
-        private const int CacheInitialCapacity = 1024;
+        public const int CacheInitialCapacity = 1024;
+
+        /// <summary>
+        /// The smallest initial capacity accepted by <see cref="UseNewCache(int)"/>. Values below this are
+        /// rejected because they defeat the purpose of pre-sizing the cache: it would immediately need to be
+        /// resized as definitions/expressions are evaluated.
+        /// </summary>
+        public const int MinimumCacheInitialCapacity = 16;
 
         /// <summary>
         /// Disables caching for subsequent operations by resetting the cache state.
--- a/Cql/Cql.Runtime/Runtime/CqlContext.cs
+++ b/Cql/Cql.Runtime/Runtime/CqlContext.cs
@@ -104,10 +104,20 @@ namespace Hl7.Cql.Runtime
         /// refreshed.</remarks>
         public void UseNewCache()
         {
-            _cache = new ConcurrentDictionary<long, object?>();
+            // A fresh cache is populated by mostly-sequential definition evaluation, so use a single
+            // lock (concurrencyLevel: 1) and pre-size it to avoid repeated resizes, which acquire
+            // all internal locks and dominated the write cost with the default settings.
+            _cache = new ConcurrentDictionary<long, object?>(concurrencyLevel: 1, capacity: CacheInitialCapacity);
             _cacheCallCount = 0;
             _cacheFactoryInvocations = 0;
         }
+
+        /// <summary>
+        /// The initial capacity of the definition/expression memoization cache. Sized to hold the
+        /// cached results of a typical measure evaluation (all definitions across all libraries in
+        /// a library set) without internal resizing.
+        /// </summary>
+        private const int CacheInitialCapacity = 1024;
 
         /// <summary>
         /// Disables caching for subsequent operations by resetting the cache state.
--- a/Cql/CoreTests/CacheTest.cs
+++ b/Cql/CoreTests/CacheTest.cs
@@ -294,4 +294,53 @@ public class CacheTest
         Assert.AreEqual(3, ((ICqlContextInternals)ctx).CacheMisses, "All calls should be misses without cache");
         Assert.AreEqual(0, ((ICqlContextInternals)ctx).CacheHits, "Should have no hits without cache");
     }
+
+    [TestMethod]
+    public void Cache_UseNewCacheWithCustomCapacity_ShouldCacheResults()
+    {
+        // Arrange
+        var ctx = FhirCqlContext.ForBundle();
+        ctx.UseNewCache(initialCapacity: 64);
+        var lib = CqlNestedTupleTest_1_0_0.Instance;
+
+        // Act - Call the same expression twice
+        var result1 = lib.Result(ctx);
+        var result2 = lib.Result(ctx);
+
+        // Assert - Caching still works with a non-default capacity
+        Assert.IsNotNull(result1);
+        Assert.IsNotNull(result2);
+        Assert.AreEqual(result1, result2);
+    }
+
+    [TestMethod]
+    public void Cache_UseNewCacheWithMinimumCapacity_ShouldSucceed()
+    {
+        // Arrange
+        var ctx = FhirCqlContext.ForBundle();
+
+        // Act - The minimum accepted capacity should not throw
+        ctx.UseNewCache(initialCapacity: CqlContext.MinimumCacheInitialCapacity);
+        var lib = CqlNestedTupleTest_1_0_0.Instance;
+        var result1 = lib.Result(ctx);
+        var result2 = lib.Result(ctx);
+
+        // Assert
+        Assert.AreEqual(result1, result2);
+    }
+
+    [DataTestMethod]
+    [DataRow(0)]
+    [DataRow(1)]
+    [DataRow(-1)]
+    [DataRow(int.MinValue)]
+    public void Cache_UseNewCacheWithTooSmallCapacity_ShouldThrow(int initialCapacity)
+    {
+        // Arrange
+        var ctx = FhirCqlContext.ForBundle();
+
+        // Act & Assert
+        var ex = Assert.ThrowsException<ArgumentOutOfRangeException>(() => ctx.UseNewCache(initialCapacity));
+        Assert.AreEqual("initialCapacity", ex.ParamName);
+    }
 }
--- a/Cql/Cql.Firely/ValueSetSource.cs
+++ b/Cql/Cql.Firely/ValueSetSource.cs
@@ -151,7 +151,15 @@ public class ValueSetSource : IValueSetDictionary
         Predicate<IValueSetFacade> @internal,
         Action<ValidateCodeParameters> external)
     {
-        if (TaskHelper.Await(() => Load(valueSetUri)) is { } vs && @internal(vs)) return true;
+        // Fast path: a warm cache hit does not need the sync-over-async machinery in Load.
+        if (_valueSets.TryGetValue(valueSetUri, out var cached))
+        {
+            if (@internal(cached)) return true;
+        }
+        else if (TaskHelper.Await(() => Load(valueSetUri)) is { } vs && @internal(vs))
+        {
+            return true;
+        }
 
         if (_termService is null) return false;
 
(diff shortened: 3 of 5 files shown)


Cql/Cql.Runtime/Runtime/CqlContext.cs
```diff
@@ -104,11 +104,21 @@ public CqlContext RaiseContextEvent(ContextEventArgs eventData)
...
             _cacheCallCount = 0;
             _cacheFactoryInvocations = 0;
         }
 
+        /// <summary>
+        /// The initial capacity of the definition/expression memoization cache. Sized to hold the
+        /// cached results of a typical measure evaluation (all definitions across all libraries in
+        /// a library set) without internal resizing.
+        /// </summary>
+        private const int CacheInitialCapacity = 1024;
```
alexzautke: I think it would be good to make this configurable

alexzautke (changes requested):

Cql/Cql.Runtime/Runtime/CqlContext.cs
baseTwo: Good idea — added an optional `initialCapacity` parameter to `UseNewCache` (default unchanged: 1024), plus a public `MinimumCacheInitialCapacity` (16) that's validated to guard against pathologically small values that would defeat the pre-sizing optimization. Public API surface updated in `PublicAPI.Unshipped.txt`, and unit tests added in `CacheTest.cs`.

alexzautke (approved):

PR #1331, opened 2026-07-07
Typed IR and C# emitter for the Linq.Expressions removal (phase 2)

ewoutkramer:
## What

Phase 2 of the Linq.Expressions removal: the new typed IR and its C# printer, plus the design document. Nothing is wired into the production pipeline yet — these components are internal, currently only reachable from their unit tests. The existing Expression-based pipeline is untouched.

See `docs/linq-expression-removal-plan.md` (added here) for the full rationale, target architecture, and phase plan. Short version: the builder''s reliance on per-node `System.Type` (operator overload resolution, generic inference, cast decisions) means any replacement must stay typed, so the target is a small typed IR that prints C# directly — not Roslyn SyntaxNodes.

## Components

**`Cql.Compiler\Ir\` — the typed IR** (~18 sealed node kinds)
- Every node carries the .NET `Type` of the C# expression it prints as (the role `Expression.Type` played).
- Constructors validate what `Expression.Call`/`Expression.Convert` used to check silently (argument assignability, cast legality, receiver/static agreement), so type errors surface at tree-construction time with ELM context available instead of as Roslyn diagnostics against generated code.
- `IrLocal` has reference-identity semantics (same instance = same variable), replacing `ParameterExpression` identity for query-alias scoping; final C# names are allocated by the emitter, so the builder never needs unique names.
- The old custom Expression subclasses map to plain nodes: `?.` → `IrProperty(nullConditional)`, case/when → `IrIfChain`, definition/function calls → `IrDefinitionCall`, `ElmAs` → `IrCast`. No `Reduce()` machinery.

**`CodeGeneration.NET\CSharpIrEmitter` — the statement-aware printer**

Replaces four rewrite passes plus `LambdaDefinitionWriter` with one component:
- linearizes nested expressions into sequential `T a_ = …;` declarations at print time (was `SimplifyExpressionsVisitor`);
- allocates variable names from a scope-shared sequence, name hints winning unless they collide (was `RenameVariablesVisitor`);
- dedups identical right-hand sides keyed on printed code + declared type (was `LocalVariableDeduper`''s expensive internal-DebugView comparison);
- prints lambdas as local functions and if-chains as native `if`/`else` (no lambda-wrap + `Invoke` tricks);
- ports the #1311 peepholes: `is`-patterns unwrap nullables, print tuple types in `ValueTuple<>` form and box value-typed operands; casts route through `object` when C# defines no direct conversion from the printed operand''s type.

**`ICSharpNamingConventions`** — the one deliberate seam worth reviewing: expression bodies must reference names invented by the class scaffolding writer (the `CqlTupleMetadata_…` static fields, and definition-call targets like `this.Foo` / `FHIRHelpers_4_0_001.Instance.ToCode`). The emitter takes this small interface, implemented by the scaffolding writer when phase 5 wires the pipelines together, and faked in tests. A convention mismatch produces non-compiling generated C#, caught by compilation and the golden-file tests.

**Tests**: 19 exact-output unit tests (`CoreTests\CSharpIrEmitterTests.cs`) building small IR trees and asserting the emitted C# text — output shape (naming, hoisting order, dedup, statement-vs-expression selection) is the component''s contract, so assertions compare full block text.

## Review pointers

- The IR node set was derived from a survey of the `Expression.*` factory methods the current builder actually calls, plus the five custom Expression classes.
- `IrConditional`/`IrIfChain` tests require `bool` (not `bool?`), matching what `Expression.Condition` enforced — the builder coerces CQL''s three-valued logic before testing, and the ported builder will keep doing so.
- Writing the tests surfaced two emitter warts that are fixed in this PR rather than shipped: trial linearization burned variable names (leaving gaps in the `a_, b_, …` sequence, which would break golden-parity later), and tail-position if-chains printed as disconnected blocks.

## Verification

- 19/19 emitter tests pass (net10.0 locally; net8.0 leg relies on CI — no x64 .NET 8 runtime on the dev machine).
- Golden-file tests and both full suites are unaffected (nothing reachable from the existing pipeline changed; the only touched existing file is `VariableNameGenerator`, which gained a string-based `ForNewScope` overload).


Cql/Cql.Compiler/Ir/IrConstant.cs
```diff
@@ -0,0 +1,61 @@
...
+internal sealed class IrDefault(Type type) : IrExpression
+{
+    public override Type Type { get; } = type;
+}
+
+/// <summary>
+/// Throws an exception. Used only for the generator's error-recovery stubs; carries a result
+/// type so it can stand in for an expression of any type.
+/// </summary>
+internal sealed class IrThrow : IrExpression
```
ewoutkramer: I wouldn't expect an IrThrow class in a file called IrConstant, maybe give this its own cs file.

Cql/Cql.Compiler/Ir/IrProperty.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Hl7.Cql.Compiler.Ir;
+
+/// <summary>
+/// A property or field access, optionally null-conditional (<c>x?.Member</c>). Replaces both
+/// <c>MemberExpression</c> and the old <c>NullConditionalMemberExpression</c> custom node.
+/// </summary>
+internal sealed class IrProperty : IrExpression
+{
+    public IrProperty(IrExpression? receiver, MemberInfo member, bool nullConditional = false)
```
ewoutkramer, line 17: Isn't it too limiting to allow the null conditional monad only on property access? Even if we don't need it for method calls now, would it be more future proof to have it anyway?

Cql/Cql.Compiler/Ir/IrConditional.cs
```diff
@@ -0,0 +1,72 @@
...
+
+    public IrExpression IfTrue { get; }
+
+    public IrExpression IfFalse { get; }
+
+    public override Type Type { get; }
+}
+
+/// <summary>
+/// A multi-way conditional producing a value, printed as an <c>if / else if / else</c>
```
ewoutkramer, line 43: Wouldn't we want a case expression rather than repeated if/else if/else?  Or maybe that's not what we do now (even if we might want to do so in the future?)

Cql/CodeGeneration.NET/CSharpIrEmitter.cs
```diff
@@ -0,0 +1,330 @@
...
+/// print time (was <c>SimplifyExpressionsVisitor</c>);</item>
+/// <item>variable names are allocated per method scope from a shared sequence
+/// (was <c>RenameVariablesVisitor</c>);</item>
+/// <item>identical right-hand sides reuse the earlier local, keyed on the printed code and
+/// declared type (was <c>LocalVariableDeduper</c>'s DebugView comparison);</item>
+/// <item>lambdas print as local functions, if-chains as native <c>if</c>/<c>else</c>
+/// statements (no lambda-wrap tricks).</item>
+/// </list>
+///
+/// <para>The printed output deliberately follows the existing generated-code style so that
```
ewoutkramer, line 29: So, in the end, we could switch to Roslyn trees to have formatted output?

Cql/CodeGeneration.NET/CSharpIrEmitter.cs
```diff
@@ -0,0 +1,330 @@
...
+    {
+        public Type Type => Node.Type;
+    }
+
+    /// <summary>
+    /// One statement scope (a method body or local-function body). Holds the statements
+    /// hoisted so far, the dedup table for this scope, and the name reservations. The name
+    /// sequence itself is shared across nested scopes so generated names never collide.
+    /// </summary>
+    private sealed class Scope
```
ewoutkramer: This has enough body to warrant being in a separate C# file.

Cql/CodeGeneration.NET/CSharpIrEmitter.cs
```diff
@@ -0,0 +1,330 @@
...
+        {
+            foreach (var statement in _statements)
+                isb.AppendLine(statement);
+        }
+
+        /// <summary>
+        /// Linearizes <paramref name="node"/>: hoists compound subexpressions into
+        /// <c>T a_ = …;</c> statements on this scope and returns the resulting simple
+        /// expression — or null when the node was a tail-position if-chain emitted entirely
+        /// as <c>return</c>-ing statements.
```
ewoutkramer: In the end we can replace these returning statements as a case expression and remove this special case?

Cql/Cql.Compiler/Ir/IrCall.cs
```diff
@@ -0,0 +1,100 @@
...
+/// (<see cref="Receiver"/> null). This covers calls to <c>ICqlOperators</c> methods
+/// (receiver is a property access on the context), <c>CqlContext</c> members, and any other
+/// .NET method the builder binds to.
+///
+/// <para>The constructor performs the validation <c>Expression.Call</c> used to do silently:
+/// the method must be resolved (no open generic definitions), the argument count must match,
+/// and each argument must be assignable to its parameter. Overload <i>resolution</i> stays in
+/// the binder; the IR only checks that the chosen method is callable with these arguments.</para>
+/// </summary>
+internal sealed class IrCall : IrExpression
```
ewoutkramer: This is really separate from a DefinitionCall (below), which calls definitions that we have generated from the CQL definitions. It would be more readable if this was called Invoke, since Call and DefinitionCall are too much alike.

Cql/Cql.Compiler/Ir/IrConstant.cs
ewoutkramer: Done in 0ba116f0e — IrThrow has its own file now (and the IrCall/IrDefinitionCall split below also gave each of those its own file).

Cql/Cql.Compiler/Ir/IrProperty.cs
ewoutkramer: Agreed — implemented in 0ba116f0e rather than deferred: IrInvoke (see rename below) now takes a nullConditional flag, printing `x?.Method(...)` and lifting value return types to Nullable<T>, mirroring IrProperty. So the IR will not need extending when the builder port hits constructs that guard calls rather than member accesses.

Cql/Cql.Compiler/Ir/IrCall.cs
ewoutkramer: Done in 0ba116f0e — renamed to IrInvoke (invokes pre-existing .NET methods), keeping IrDefinitionCall for calls to the C# methods this compiler generates from CQL definitions. Both now live in their own files, and the doc comments cross-reference each other to make the distinction explicit.

Cql/CodeGeneration.NET/CSharpIrEmitter.cs
ewoutkramer: Done in 0ba116f0e — Scope moved to CSharpIrEmitter.Scope.cs (same nested-partial pattern LambdaDefinitionWriter uses).

Cql/Cql.Compiler/Ir/IrConditional.cs
ewoutkramer: Good instinct, and it is possible: CQL case/when has independent boolean conditions per branch (no single scrutinee), but C# can express that as `_ switch { _ when c1 => v1, ..., _ => vElse }` or nested ternaries. Two reasons I kept if/else for now: (1) parity — the golden corpus was printed in if/else style by the old writer, and phase 5 diffs the two pipelines byte-wise; (2) switch-expression arms must be *expressions*, so any branch that needs hoisted statements (locals, local functions) cannot live in an arm — the statement form has to exist as the general case regardless. Post-migration, when the generator version bumps and output is free to change, emitting a switch expression for the all-simple-branches case is a nice readability upgrade. I'll note it in the plan doc as a post-parity cleanup.

Cql/CodeGeneration.NET/CSharpIrEmitter.cs
ewoutkramer: Partially — same trade-off as the IrIfChain comment above. Once parity stops constraining output, the all-simple-branches case can collapse to `return cond1 switch { ... }` and this tail special case shrinks. But it cannot disappear entirely: branches that hoist statements (locals, local functions) cannot live inside a switch-expression arm, so the statement form remains the general fallback. Noted in the plan doc as post-parity cleanup.

Cql/CodeGeneration.NET/CSharpIrEmitter.cs
ewoutkramer: Exactly right. The argument against Roslyn trees was about replacing the *typed layer* (Roslyn nodes carry no System.Type, and overload resolution needs types on every node) — not about the final formatting step. Once phase 6 deletes the old pipeline and byte-parity no longer pins the output, the printing backend is swappable in isolation: the IR stays, and the emitter could build Roslyn SyntaxNodes + NormalizeWhitespace instead of strings, or simply reformat. I've made the emitter's class doc say this explicitly (see 0ba116f0e).

Cql/CodeGeneration.NET/CSharpIrEmitter.cs
```diff
@@ -0,0 +1,330 @@
...
+        }
+
+        private void NameParameters(IReadOnlyList<IrLocal> parameters)
+        {
+            foreach (var p in parameters)
+            {
+                if (!_emitter._assignedNames.ContainsKey(p))
+                    _emitter._assignedNames[p] = p.NameHint ?? _names.Next();
+            }
+        }
```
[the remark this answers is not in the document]
ewoutkramer: Fixed in 0ba116f0e — Scope.AllocateName now honors a hint only when it is not a C# keyword and not already used in the emission (tracked in a per-emission used-name set); otherwise it falls back to generated names. Covered by NameHints_KeywordAndDuplicateHintsFallBackToGeneratedNames.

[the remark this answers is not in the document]
ewoutkramer: Fixed in 0ba116f0e (same fix as the thread above: per-emission used-name set + keyword check in Scope.AllocateName, with a regression test).

Cql/CodeGeneration.NET/CSharpIrEmitter.cs
```diff
@@ -0,0 +1,330 @@
...
+
+    /// <summary>
+    /// Emits the body of a definition as a C# block statement, including the final
+    /// <c>return</c>. Parameter names are taken from the lambda's <see cref="IrLocal"/>
+    /// name hints (allocated names on collision).
+    /// </summary>
+    public string EmitBodyBlock(IrLambda lambda)
+    {
+        var scope = Scope.CreateRoot(this, lambda.Parameters);
+        var result = scope.Linearize(lambda.Body, tailPosition: true);
```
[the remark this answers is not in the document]
ewoutkramer: Fixed in 0ba116f0e — EmitBodyBlock clears the naming state at the start of each emission, and the class doc now states instances are not thread-safe with per-emission naming state.

Cql/CodeGeneration.NET/CSharpIrEmitter.cs
```diff
@@ -0,0 +1,330 @@
...
+            }
+            isb.AppendLine("}");
+        }
+
+        /// <summary>
+        /// A condition of type <c>bool?</c> must compare to <c>true</c> to be used in an
+        /// <c>if</c>/ternary, mirroring CQL's three-valued logic (null is not true).
+        /// </summary>
+        private static string NullableTestCondition(Atom test) =>
+            test.Type == typeof(bool?) ? $"{test.Code} ?? false" : test.Code;
```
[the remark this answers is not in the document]
ewoutkramer: Removed in 0ba116f0e (it was a leftover from an earlier bool?-condition design that the IR constructors now rule out).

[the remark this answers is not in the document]
ewoutkramer: Removed in 0ba116f0e.

Cql/Cql.Compiler/Ir/IrConstant.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <summary>The constant's value; null prints as the <c>null</c> literal.</summary>
+    public object? Value { get; }
+
+    public override Type Type { get; }
+}
+
+/// <summary>
+/// The C# <c>default</c> value for a type, printed as <c>null</c> for nullable types and
+/// <c>default(T)</c> otherwise.
+/// </summary>
```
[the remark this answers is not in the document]
ewoutkramer: Fixed in 0ba116f0e — the doc now states the actual contract: null for any type that can hold null (reference types and Nullable<T>), default(T) for other value types.

Cql/CodeGeneration.NET/CSharpIrEmitter.Print.cs
```diff
@@ -0,0 +1,219 @@
...
+            {
+                Enum e when Enum.IsDefined(e.GetType(), e) => $"{e.GetType().FullName}.{e}",
+                Enum e => $"({e.GetType().FullName}){e}",
+                bool b => b ? "true" : "false",
+                decimal d => FormattableString.Invariant($"{d}m"),
+                long l => FormattableString.Invariant($"{l}L"),
+                int i => FormattableString.Invariant($"{i}"),
+                null => "default",
+                var v => FormattableString.Invariant($"{v}"),
+            };
```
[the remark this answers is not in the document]
ewoutkramer: Confirmed and fixed in 0ba116f0e — boxed default struct values print as "default" via IsObjectNullOrDefault(), restoring the previous pipeline's behavior, with a regression test (default(DateTime)). Writing the tests also exposed a second gap in the same switch: char constants printed bare (x) instead of as literals; now formatted via SymbolDisplay.FormatLiteral.

PR #1348, opened 2026-07-11
Filter retrieves by profile discriminators (QICore negation)

alexzautke:
## Description
Profile-informed ELM carries the retrieve's profile in `templateId`, but `BundleDataSource` ignored it: QICore negation-profile retrieves (`ServiceNotRequested`, `MedicationNotRequested`, `ProcedureNotDone`, `ObservationCancelled`, ...) returned positive instances, and positive-profile retrieves returned negated ones. This was previously masked by the late-bound property bug fixed in #1347: the wrongly retrieved resources always failed their `authoredOn`/timing checks because the property read returned null.

This PR introduces `IRetrieveProfileFilter`, mapping a retrieve's `templateId` to a predicate that instances must satisfy. The default `QICoreRetrieveProfileFilter` hardcodes the discriminating elements the QICore negation profiles fix and their inverse for the paired positive profiles:

- `servicenotrequested` / `medicationnotrequested`: `doNotPerform = true`
- `procedurenotdone`, `medicationadministrationnotdone`, `communicationnotdone`: `status = not-done`
- `observationcancelled`: `status = cancelled`
- `taskrejected`: `status = rejected`
- positive counterparts (`qicore-servicerequest`, `qicore-medicationrequest`, `qicore-procedure`, the QICore observation profiles, etc.) exclude instances in the negated state

Only instance data is inspected — declared `meta.profile` is never consulted, so real-world resources without profile declarations behave exactly as before. Unknown template ids and retrieves without one remain unfiltered. The filter can be replaced via `FhirCqlContextOptions.OverrideRetrieveProfileFilter`.

The integration-runner submodule is bumped to restore the CMS22 (`c3f55d79`) and CMS108 (`41f2785f`) test cases to the known-passing list: with this change they match their official MADiE expectations again, this time for the right reason.

Two integration cases that data-driven filtering intentionally cannot pass stay out of the known-passing list, for symmetric reasons:

- CMS190 (`a82cd0c1`) — its test resource carries `doNotPerform=false` while declaring the `servicenotrequested` profile, so it is a **negation by declaration only**.
- CMS22 (`86618b52`) — **removed** from the known-passing list. Its two follow-up `ServiceRequest`s declare the positive `qicore-servicerequest` profile but carry `doNotPerform=true`, so they are excluded from the Numerator interventions (actual `0` vs MADiE-expected `1`): a **positive by declaration only** case.

In both, the instance data contradicts the declared profile, which data-driven filtering deliberately does not reconcile.

The paired integration-runner commits live on the `claude/cms2-measure-logic-dm2r61` branch of that repository (FirelyTeam/Firely.Cql.Sdk.Integration.Runner#104), rebased onto the runner's current `main`. It carries the CMS22/CMS108 restoration, the full-precision measurement-period changes, and the `86618b52` removal; the submodule points at its tip.

## Related issues
Follow-up to #1347.

## Testing
- New `RetrieveProfileFilterTests` (6 tests) exercising negation-profile, positive-profile, unknown-profile and no-template-id retrieves plus the override option, through the public `FhirCqlContext`/`ICqlOperators.Retrieve` path.
- Full CoreTests suite green on .NET 8 and .NET 10 (743 passed each).
- Full integration suite green with the bumped submodule: 2700 passed, 0 failed (verified locally against this branch on .NET 10).

## FirelyTeam Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qp7JD2pmAV3KTXWmFkRgq1)_


Cql/Cql.Firely/RetrieveProfileFilter.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <summary>
+        /// Returns a predicate that an instance must satisfy to be returned by a retrieve
+        /// declared against the given profile, or <see langword="null"/> when no
+        /// constraints are known for the profile (the retrieve remains unfiltered).
+        /// </summary>
+        Predicate<Resource>? GetFilter(string templateId);
+    }
+
+    /// <summary>
+    /// A <see cref="IRetrieveProfileFilter"/> for the QICore negation profiles and their
```
ewoutkramer, line 35: This should make really clear this stuff needs to be kept up to date if a new version of QiCore is published + state the version this was derived from.

ewoutkramer (changes requested):

Cql/Cql.Firely/RetrieveProfileFilter.cs
alexzautke: Good point. Documented in b8a9cf7: the class `<remarks>` now states the table is derived from **QICore 6.0.0 (STU6)** and must be revisited/kept in sync whenever the SDK targets a new QICore version (profiles and their discriminating elements can change between releases), and there's a matching inline comment on the `FiltersByProfile` table.

---

ewoutkramer (dismissed):

ewoutkramer (dismissed):

alexzautke:
Also merge https://github.com/FirelyTeam/Firely.Cql.Sdk.Integration.Runner/pull/104

ewoutkramer (approved):

Issue #1350, opened 2026-07-11
Choice-typed interval operands lowered as Interval<CqlDate> with hard casts, crashing CMS1173


alexzautke:
**Describe the bug**
When an interval is constructed over operands whose static type the expression builder cannot pin down (choice-typed values surfaced as `object`), the builder emits `context.Operators.Interval(...)` typed as `CqlInterval<CqlDate>` with **hard CLR casts** on the operands. At runtime the operands are `CqlDateTime`s, so every evaluation throws:

```
InvalidCastException: Unable to cast object of type 'Hl7.Cql.Primitives.CqlDateTime' to type 'Hl7.Cql.Primitives.CqlDate'.
```

This is likely the same root cause as #543: with nothing to go on, overload resolution picks the first `Interval` overload, which happens to be `CqlDate` — but here it produces a runtime crash in published measure content rather than a mistyped null interval.

**To Reproduce**
CMS1173FHIRDiagnosticDelayVTE (dqm-content-qicore-2025), Numerator:

```cql
and ( AntiCoagulantOrdered.authoredOn 12 hours or less before VTEStudy.effective
    or AntiCoagulantOrdered.authoredOn 12 hours or less after VTEStudy.effective )
```

`authoredOn` is a FHIR `dateTime` and `VTEStudy.effective` is a choice type, so the operands flow through `FHIRHelpers.ToValue` as `object`. The generated C# (`CMS1173FHIRDiagnosticDelayVTE-1.0.000.g.cs`, around the Numerator's timing check) contains:

```csharp
object al_ = FHIRHelpers_4_4_000.Instance.ToValue(context, n_);      // boxes a CqlDateTime
CqlDateTime ap_ = context.Operators.Add(an_ as CqlDateTime, y_);     // statically CqlDateTime!
CqlInterval<CqlDate> aq_ = context.Operators.Interval((CqlDate)al_, (CqlDate)(ap_ as object), false, true);
```

`(CqlDate)al_` on a boxed `CqlDateTime` and `(CqlDate)(ap_ as object)` on a statically-typed `CqlDateTime` can never succeed. The builder then converts the interval to `CqlInterval<object>`, picks it apart into `.low`/`.high`/`.lowClosed`/`.highClosed`, and reassembles a `CqlInterval<CqlDateTime>` — so the intended element type was evidently `CqlDateTime` all along.

[...]

PR #1351, opened 2026-07-11
Fix interval construction with choice-typed operands

alexzautke:
## Summary
Fixes issue #1350 where interval construction with choice-typed operands (e.g., `Choice<DateTime, Interval<DateTime>>`) would fail at runtime with `InvalidCastException`. The binder was arbitrarily selecting an interval point type without proper type anchoring, resulting in invalid casts.

## Key Changes
- **Extracted `IntervalArgs()` method**: Refactored the inline interval argument construction into a dedicated method to handle choice-typed operands with proper type resolution.
- **Added type anchoring logic**: When interval operands are choice-typed (surfaced as `object` in C#), the code now:
  - Checks if either operand has a valid interval point type and uses it to anchor the other
  - Falls back to extracting the single valid interval point type from the choice type specifier
  - Converts choice-typed operands using 'as' semantics to the anchored point type
- **Added helper methods**:
  - `IsIntervalPointType()`: Validates whether a type is supported by `ICqlOperators` interval factory overloads (int, long, decimal, Quantity, Date, DateTime, Time)
  - `SingleIntervalPointTypeFromChoice()`: Extracts the single valid interval point type from a choice type specifier
- **Added comprehensive test**: `Interval_WithChoiceTypedOperands_AnchorsPointType()` validates the fix with a scenario mirroring CMS1173

## Implementation Details
The fix ensures that when an interval's ELM point type is a choice (e.g., from `FHIRHelpers.ToValue`), the compiler properly anchors the point type on a valid alternative rather than making arbitrary selections that lead to runtime failures.


ewoutkramer (approved):

ewoutkramer:
Conflicts solved in #1405.

PR #1353, opened 2026-07-12
Treat open null interval boundaries as unknown in interval comparisons

### LINKED ISSUE #1352: Interval operators treat open null boundaries as min/max instead of unknown
**Describe the bug**
Per the CQL spec, a null interval boundary is only interpreted as the minimum/maximum value of the point type when the boundary is **closed**; a null **open** boundary is *unknown*, and comparisons against it must result in null. Several interval operators substitute min/max unconditionally:

- `OverlapsHelper`, `OverlapsAfterHelper`, `OverlapsBeforeHelper` (`CqlOperators.IntervalOperators.cs` ~1885, ~1919, ~1952): `left.high ?? MaxValue<T>()` etc., ignoring the closed flags.
- `IntervalIncludesInterval` (~1236): same pattern.

`ToClosedHelper` handles this correctly (a null open boundary stays open-null, ~2373), and `In` handles it correctly too (returns null when a boundary is open-null, ~1208) — the substitution in the helpers above then discards that information.

**To Reproduce**
`QICoreCommon.prevalenceInterval()` builds `Interval[start of onset, abatementDate)` for a **non-active** condition; with no abatement, that's an open null high boundary. The overlap check should be null (unknown), so the condition drops out of `where` clauses — but with max-substitution the interval extends to infinity and the condition matches.

Two CMS347 integration test cases probe exactly this and fail against their official MADiE expectations even after correcting for the runner's multi-rate comparison (they are the only 2 of 149 CMS347 failures not explained by the harness):

- `695b64d8-8102-4109-89c2-9ca128d43f4d` ("RhabdomyolysisStatusInactive"): rhabdomyolysis inactive in the MP, no abatement — must not count as denominator exclusion; the SDK counts it (group 2 denominator-exclusion 1 vs expected 0).
- `a0202aaf-756f-4d08-8329-8fd585ddda63`: palliative-care diagnosis inactive before the MP — must not count as denominator exception; the SDK counts it (group 1 denominator-exception 1 vs expected 0).

**Expected behavior**
Comparisons involving a null open boundary yield null, combined with three-valued logic; a null closed boundary keeps the min/max interpretation.

**Version used:**
- Branch: develop

**Additional context**
Found while triaging the CMS347 failures (149 cases): 147 are the integration runner's multi-rate scoring comparison, these 2 are this engine defect. Related to the mixed-precision comparison semantics but distinct from it.

alexzautke:
Overlaps, OverlapsBefore, OverlapsAfter and IntervalIncludesInterval
substituted the type's minimum/maximum for null boundaries regardless of
whether the boundary was closed. Per the CQL spec that interpretation
only applies to closed boundaries; a null open boundary is unknown.
This made e.g. QICoreCommon.prevalenceInterval() for an inactive
condition without an abatement date - Interval[onset, null) - extend to
infinity, so the condition wrongly satisfied overlap checks against the
measurement period.

An unknown boundary is still a boundary of its own interval, so it is
constrained by the sibling boundary: an unknown high ranges over
[low, maximum], an unknown low over [minimum, high]. Comparisons are
evaluated over that range - true when every possible value satisfies
them, false when none does, and null otherwise - and combined with
three-valued logic. This keeps decidable cases decidable: an interval
starting inside another one still overlaps it no matter where it ends
(e.g. CMS1154's inactive prediabetes diagnosis inside the look back
period), while a genuine unknown - onset before the window, end unknown
- is null and drops out of where clauses (e.g. CMS347's inactive
palliative care diagnosis, expected not to count as an exception).

Comparisons between known boundaries are unchanged, including the
existing treatment of indeterminate mixed-precision comparisons.

Verified: new unit tests fail without the fix with the exact production
behavior; full CoreTests suite green; full gated integration suite
green (2369 passed, 0 failed) including CMS1154 bc9c82ca which relies
on the still-decidable overlap.

Fixes #1352.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Qp7JD2pmAV3KTXWmFkRgq1


baseTwo:

baseTwo:

baseTwo:

ewoutkramer (approved):

PR #1355, opened 2026-07-12
Fix tuple list union collapsing to null

### LINKED ISSUE #1354: Union of lists with different tuple types generates `as IEnumerable<object>` casts that are always null
## Summary

When a CQL `union` combines two lists whose element types are structurally compatible but not identical (typically two tuple types that differ in one element type), the operator binder casts both operands to `IEnumerable<object>` using a type-as conversion:

```csharp
// CqlOperatorsBinder.Specific.cs, Union(...)
var leftAsObjectEnumerable = left.NewTypeAsExpression<IEnumerable<object>>();
var rightAsObjectEnumerable = right.NewTypeAsExpression<IEnumerable<object>>();
return BindToBestMethodOverload(nameof(ICqlOperators.Union), [leftAsObjectEnumerable, rightAsObjectEnumerable], [])!;
```

This produces generated code like (CMS108, "No VTE Prophylaxis Medication Administered Or Ordered"):

```csharp
IEnumerable<object> bo_ = context.Operators.Union<object>(af_ as IEnumerable<object>, bn_ as IEnumerable<object>);
```

`IEnumerable<out T>` covariance only applies to **reference** type arguments. The generated tuple types are nullable **value** tuples (`(CqlTupleMetadata, string id, ...)?`), so `af_ as IEnumerable<object>` evaluates to `null` at runtime. `ICqlOperators.Union(null, null)` returns `null`, and the whole define silently evaluates to empty — no error, no diagnostic.

## Impact

Any define that unions tuple lists with mismatching element types is silently empty. In the dqm-content-qicore-2025 measure set the broken cast (tuple-typed operand) is generated in:

- CMS108FHIRVTEProphylaxis — numerator ("medical reason for no VTE prophylaxis" cases) scores 0
- CMS190FHIRVTEProphylaxisICU — same pattern, numerator scores 0
- CMS72FHIRSTKAntithromboticDay2 — denominator exception scores 0
- CMS2, CMS137, CMS871, CMS1218 — cast present, currently latent

Together this accounts for roughly 75 failing MADiE integration test cases.

## Reproduction

```csharp
IEnumerable<(CqlTupleMetadata, string id, CqlDateTime authoredOn)?> tuples =
    new List<(CqlTupleMetadata, string id, CqlDateTime authoredOn)?> { /* one element */ };

var asObject = tuples as IEnumerable<object>;   // null - covariance does not apply to value types
var union = ctx.Operators.Union<object>(asObject, asObject); // null
```

## Fix

In the mixed-element-type branch of `CqlOperatorsBinder.Union`, box the elements explicitly (e.g. bind a `Select` that converts each element to `object`) instead of emitting a reference conversion that cannot succeed for value-typed elements. Reference-typed elements can keep the covariant cast.

alexzautke:
## Description
A CQL `union` of two structurally compatible tuple lists whose element types differ (e.g. one branch's `authoredOn` materializes as `FHIR.dateTime`, the other as `System.DateTime`) was bound as `Union<object>` with both operands converted through `as IEnumerable<object>`. The C# code generator lowers the compiler-generated tuple types to value tuples, for which `IEnumerable<T>` covariance does not apply, so both casts yielded `null` at runtime and the whole define silently evaluated to empty. In the qicore-2025 measure set this broke the VTE prophylaxis "medical reason not done" numerators (CMS108, CMS190) and the CMS72 denominator exceptions, among others.

The fix converts union operands to the ELM-declared result type up front, so the union binds over a single element type and the object erasure disappears:

- **`CodeBuilderContext` (Union args)**: when compatible operands have different element types, both are converted to the declared result type, guarded by a directional `CanConvertForUnionOperation` check.
- **`ChangeType`**: list conversions involving tuple element types go through the element-wise `Select` path instead of a covariant cast that would not survive the value-tuple lowering; a null tuple converts to null instead of a tuple whose elements are all null; and the element-wise `Select` is reported as an operator conversion so the `As` handler (used for definitions typed as a list of choices) doesn't discard it in favor of a type-as cast.
- **`FhirTypeConverter`**: registers `Coding → CqlCode` and `CodeableConcept → CqlConcept` (the implicit FHIRHelpers `ToCode`/`ToConcept` conversions), needed to convert tuple elements typed via FHIR profile types such as `DoNotPerformReason`.
- **`CqlOperators.LateBoundProperty`**: resolves property names on lowered value tuples via the `CqlTupleMetadata` in their first item — consumers of choice-typed tuple unions access their elements late-bound, and value tuples have no named properties.
- **`CqlOperatorsBinder`**: the fallback erasure to `IEnumerable<object>` boxes value-typed and tuple-typed elements element-wise.

`GeneratorToolVersion` is bumped 5.1.4.0 → 5.1.5.0 and every checked-in `*.g.cs` regenerated. Only CMS108, CMS190 and CMS72 change content; files whose only diff was the version header are left at their previous version, matching develop's `172f26e63`. A corpus-wide scan confirms no tuple-typed `as IEnumerable<object>` casts remain anywhere in the checked-in generated C#. (The `Union<object>` calls that do remain are genuine choice-type unions over *reference*-typed operands, where covariance does apply.)

## Related issues
Fixes #1354.

## Testing
- New regression test `Union_OfCompatibleTuplesWithDifferentElementTypes_KeepsAllElements` compiles hand-built ELM through the full C#-generation + Roslyn pipeline and executes it; it fails on `develop` (union evaluates to null) and passes with this fix.
- CoreTests (825) and CqlToElmTests (2,455) pass on net10.0; the golden-file generation tests confirm unaffected measures produce byte-identical output.
- `dotnet build Cql-Sdk-All.sln --configuration Release` is green with the submodule at develop's pointer — this is the CI stage that was previously failing.

## Integration runner / MADiE status
The earlier revision of this PR pinned the `Firely.Cql.Sdk.Integration.Runner` submodule at `ec9f101`, which predates develop's runner work — in particular the `Hedis2025.GoldenTests` project that `Cql-Sdk-All.sln` already references. That made the *Build, Sign & Package* stage fail to restore the solution, and was the cause of the red CI on this PR.

This PR now keeps develop's submodule pointer (`f9a6194`). Consequently the earlier claim that **158 previously failing MADiE test cases now pass is not backed by this revision**: the runner's vendored library resources still embed C# generated by version 5.1.2.0, so a MADiE run does not exercise this fix. That is a pre-existing lag on develop, not a regression introduced here — regenerating the runner resources and re-deriving `passed.jsonl` is tracked in #1407.

## FirelyTeam Checklist


baseTwo:

baseTwo:

baseTwo:

ewoutkramer (dismissed):

baseTwo (approved):

PR #1358, opened 2026-07-12
Add stratifier support to generated FHIR Measures

alexzautke:
## Description
The packager now maps `@stratifier` ELM annotations onto `Measure.group.stratifier`; previously these annotations were silently ignored.

- All stratifier-tagged definitions of a group collapse into a single container stratifier (`<group>-Stratifier`, no top-level `criteria`) with one `component` per `(group, stratifier)` tag pair — including when a group has only one dimension.
- Each component gets id `<group>-StratifierComponent-<value>`, a text-only code from the tag value, an optional `@description`, and a `text/cql-identifier` criteria referencing the CQL definition.
- Empty `@stratifier` values, `@stratifier` without any `@group` tag, and duplicate component codes within a group throw with clear messages.
- `GetOrCreateGroup` is extracted from the population pass so groups can be established by stratifier-only definitions regardless of ELM statement order. Population behavior is unchanged.

Also includes a new "Measure Annotations" section in `docs/cql-packager.md` and a vNext release notes entry. All examples and test content are synthetic.

Follow-up: extend `LibrarySets/Demo/Cql/MeasureExample.cql` with stratifier definitions, which requires regenerating the committed demo ELM with the Java CQL-to-ELM translator.

## Related issues
None.

## Testing
- 10 new unit tests in `Cql/CoreTests/Packaging/FhirMeasureExtensionsTests.cs` covering the container/component shape, single-dimension groups, multi-group definitions, `@description` handling, multiple `@stratifier` tags on one definition, stratifier-only groups, and the three error paths. Pass on net8.0 and net10.0.
- Existing packaging and toolkit tests in CoreTests still pass (83 total).
- End-to-end: ran `cql-package elm` over a synthetic ELM library with stratifier annotations and verified the generated `Measure-*.json` stratifier subtree matches the documented mapping.

## FirelyTeam Checklist

https://claude.ai/code/session_01JUL8GUnH9ycowzPDmQ1gHe


Cql/Cql.Packaging/FhirMeasureExtensions.cs
```diff
@@ -192,6 +199,75 @@ from p in populations
...
+
+                container.Component.Add(new FhirMeasure.ComponentComponent
+                {
+                    ElementId = componentId,
+                    Code = new CodeableConcept { Text = tuple.Stratifier },
+                    Description = description?.value ?? tuple.Stratifier,
+                    Criteria = new Hl7.Fhir.Model.Expression
+                    {
+                        Language = "text/cql-identifier",
+                        ExpressionElement = new FhirString(def.name)
```
baseTwo, line 267: @alexzautke question: when a single CQL definition carries multiple `@stratifier` tags for the same group (e.g. `@stratifier: Region` and `@stratifier: AgeBand` on the same `define`, as in the `DefinitionWithMultipleStratifierTags_AddsOneComponentEach` test), this produces two separate `stratifier.component` entries — one per tag — but both `Criteria` point at the *same* CQL expression (`def.name`). Is that intentional? A consuming measure-evaluation engine would compute the identical stratum value for two differently-coded dimensions, which seems surprising for a "multi-dimensional stratifier" use case. If a CQL author accidentally copy-pastes a second `@stratifier` tag onto an existing definition instead of writing a new one, this would silently produce a component whose criteria doesn't actually stratify on that dimension. Should this require a distinct CQL definition per dimension instead, or is sharing one expression across multiple stratifier codes a supported pattern?

Cql/Cql.Packaging/FhirMeasureExtensions.cs
```diff
@@ -60,6 +60,7 @@ public static FhirMeasure CreateMeasureResource(
             measure.Group = [];
 
             FhirMeasureExtensions.AnnotateMeasurePopulations(measure, elmLibrary);
+            FhirMeasureExtensions.AnnotateMeasureStratifiers(measure, elmLibrary);
```
baseTwo, line 63: @alexzautke question: `AnnotateMeasurePopulations` and `AnnotateMeasureStratifiers` each scan every statement in the library independently, and nothing stops a single `define` from carrying both `@population` and `@stratifier` tags. Is that combination meant to be supported (a definition acting as both a population criterion and a stratifier dimension), or should it be rejected the same way the empty-value/missing-`@group` cases are? Right now it would silently succeed and produce both a `Measure.group.population` and a `Measure.group.stratifier.component` entry from the same definition. Worth a quick doc note or a guard either way, plus a test for whichever behavior is intended.

Cql/Cql.Packaging/FhirMeasureExtensions.cs
```diff
@@ -99,16 +100,40 @@ public bool TryCreateMeasure(
...
 
+    private static Tag[] GetAnnotationTags(Hl7.Cql.Elm.ExpressionDef def) =>
+        (def.annotation?
+            .OfType<ElmAnnotation>()
+            .SelectMany(a => a.t ?? Enumerable.Empty<Tag>())
+         ?? [])
+        .Where(t => t is not null)
+        .ToArray();
+
+    private static FhirMeasure.GroupComponent GetOrCreateGroup(FhirMeasure fhirMeasure, string rate)
```
baseTwo: @copilot please address the following:

1. Nit: the `rate` parameter on `GetOrCreateGroup` is leftover naming from when this logic only lived in `AnnotateMeasurePopulations`. Now that it's also called from `AnnotateMeasureStratifiers` (line 252) with a stratifier's `@group` value, `rate` is misleading — please rename it to something like `groupId` (and update the call sites accordingly).
2. Please re-check your earlier comment on this PR (about `GetAnnotationTags` returning null tags) against the current code — it looks like it was already fixed in a later commit (null entries are filtered via `.Where(t => t is not null)`), so confirm that's fully resolved and there's nothing left to do there.
3. Take another pass over the current diff for any other nitpicks (naming, redundant code, missed edge cases) and fix what you find directly.

Cql/Cql.Packaging/FhirMeasureExtensions.cs
alexzautke: @copilot This is a supported feature. Add a test that an expression can be used to define the population as well as be a criteria-based stratifier.

Cql/Cql.Packaging/FhirMeasureExtensions.cs
baseTwo: @alexzautke I did some digging in the FHIR spec to answer this myself.

**R4 base spec** (`Measure.group.stratifier.component`, https://hl7.org/fhir/R4/measure-definitions.html): comment says *"Stratifiers are defined either as a single criteria, or as a set of component criteria"*, and `component.criteria` is cardinality `1..1` — every component must carry its own criteria expression, which already hints components aren't meant to be interchangeable copies of one expression.

**The decisive text is in the Quality Measure IG** (build.fhir.org/ig/HL7/cqf-measures/measure-conformance.html, §3.4.8 Stratification, Conformance Requirement 3.17):

> If component stratifiers are used, all the component expressions SHALL return the same type within a stratifier

and, on how components combine:

> When they return stratum values, the combination of component values constitutes the stratum value. For example, in patient-based measures, combining gender and payer type expressions produces a composite stratum value for each patient.

So the intended pattern is: **each component is its own distinct CQL expression evaluating one dimension** (e.g. a "gender" define and a separate "payer type" define), and the *combination* of those independently-evaluated per-component values is what forms the compound stratum for a subject — not one shared expression relabeled under two codes.

That means `DefinitionWithMultipleStratifierTags_AddsOneComponentEach` (two `@stratifier` tags on the *same* `define`, producing two components that both point their `criteria` at that one expression) doesn't match the spec's model of a multi-component stratifier — both components would resolve to the exact same value at evaluation time, which isn't a meaningful "Region" + "AgeBand" combination the way "gender" + "payer type" is.

Suggest either:
- Requiring a distinct CQL definition per `@stratifier` tag when multiple tags target the same group (mirroring how `@population` requires one definition per population kind), or
- If the current behavior is intentional as a shorthand for something else, documenting explicitly in `docs/cql-packager.md` that multiple `@stratifier` tags on one definition share criteria, so authors don't assume spec-standard multi-dimensional combination semantics.

**Clarification:** to be precise about how much of this the spec actually mandates — Conformance Requirement 3.17 only requires component expressions to **share the same type**; it doesn't require them to be *distinct* from one another, and the "gender + payer type" text is introduced as an example ("For example..."), not a normative "SHALL NOT reuse an expression across components" rule. So nothing here is a strict spec violation — two components sharing one expression is structurally spec-legal. The case for requiring distinct expressions is pragmatic rather than spec-mandated: two components resolving to an identical value per subject is unlikely to be useful, and it's an easy trap when copy-pasting a `@stratifier` tag onto an existing definition. So this is really a design call for you to make, not a conformance bug — happy to defer to whichever behavior you intend, as long as it's what's actually wanted here.

alexzautke: Agree that this is practically not a realistic scenario. I think we should for now be forgiving here and accept it until we have more practical experience with these annotations

baseTwo (approved):
Both open discussion threads are resolved:
- `@population`+`@stratifier` co-occurrence: confirmed as a supported feature, now covered by `DefinitionWithPopulationAndStratifierTags_ProducesBoth`.
- Multi-`@stratifier`-tag shared-criteria question: confirmed structurally spec-legal (Quality Measure IG only requires component expressions to share the same type, not be distinct); accepted as-is for now per @alexzautke, to revisit with more real-world usage.

Verified locally: checked out the branch, built `Cql.Packaging`/`CoreTests` (net8.0), ran the 11 new `FhirMeasureExtensionsTests` plus the full Packaging suite (78 tests) — all green, no regressions. CI is also green on the latest commit.

PR #1357, opened 2026-07-12
Fix inverted null boundary handling in OverlapsAfter

alexzautke:
## Description
`OverlapsAfterHelper` substituted null interval boundaries with the wrong extremes: a null low boundary was treated as the *maximum* value and a null high boundary as the *minimum* (`left.low ?? MaxValue`, `left.high ?? MinValue`, `right.high ?? MinValue`). A null low boundary is the minimum value and a null high boundary the maximum — as `OverlapsBeforeHelper` directly below already has it — so an interval with an unbounded end never overlapped after anything.

Concretely, AHAOverall's `overlapsAfterHeartFailureOutpatientEncounter` builds `Interval[start of onset.toInterval(), lastOccurrence]` from an `AllergyIntolerance`. `lastOccurrence` is almost always absent, so the interval has a null high boundary and the `overlaps after` check always returned false — which silently disabled the "Has Allergy or Intolerance to ..." denominator exceptions of CMS135FHIRACEIorARBorARNIforHF and CMS144FHIRHFBetaBlockerForLVSD.

The fix swaps the substitutions to the correct extremes. It is a runtime-only change: no generated code is affected, so the integration-runner submodule bump only carries the updated pass list ([runner branch](https://github.com/FirelyTeam/Firely.Cql.Sdk.Integration.Runner/tree/claude/overlaps-after-passes)).

Note: #1353 (open null boundaries are unknown, #1352) rewrites the same helper on its branch and preserves these substitutions in its known-boundary comparisons — whichever PR merges second needs a small rebase adopting the corrected extremes there as well.

## Related issues
Fixes #1356.

## Testing
- New `OverlapsAfterNullBoundaryTests`: the two null-boundary tests fail on `develop` and pass with the fix; four more tests pin the unchanged behavior for bounded and decidable cases.
- Full CoreTests (740) and CqlToElmTests (2,426) suites pass.
- Full MADiE integration run: **15 previously failing test cases now pass** (CMS144 7, CMS135 6, CMS646 2) with **zero regressions**.

## FirelyTeam Checklist
- [x] **Update the title** of the PR to be succinct and less than 50 characters
- [ ] Mark the PR with the label **breaking change** when this PR introduces breaking changes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Qp7JD2pmAV3KTXWmFkRgq1

PR #1360, opened 2026-07-13
Fix interval Union meets handling and tail drop

### LINKED ISSUE #1359: Interval Union: meeting intervals return null and the right-starts-first branch drops the left interval's tail
`Union<T>(CqlInterval<T>?, CqlInterval<T>?)` in `CqlOperators.IntervalOperators.cs` has two defects:

## 1. Meeting intervals return null

The implementation only merges genuinely overlapping intervals (`left.high >= right.low`). Per the CQL specification, interval `union` returns null only when the arguments *neither overlap nor meet*:

```cql
Interval[2, 2] union Interval[3, 3]  // expected [2, 3] - the intervals meet; actual: null
```

Detecting "meets" requires a successor-aware comparison (`successor(left.high) >= right.low`) on `ToClosed`-normalized intervals. The current implementation reads raw boundaries and never normalizes, so open boundaries are also mishandled beyond flag propagation.

## 2. Self-comparison drops the left interval's tail

In the branch where the right operand starts first:

```csharp
if (Comparer.Compare(right.high!, left.low!, null) >= 0)
{
    if (Comparer.Compare(left.high!, left.high!, null) > 0)  // compares left.high with itself - always false
        return new CqlInterval<T>(right.low, left.high, right.lowClosed, left.highClosed);
    else return right;
}
```

The condition should compare `left.high` with `right.high`. As written it is never true, so whenever the right operand starts first the method returns `right` unchanged:

```cql
Interval[3, 10] union Interval[1, 5]  // expected [1, 10]; actual: [1, 5] - silently discards [5, 10]
```

This is a wrong-answer bug rather than an over-conservative null.

## Notes

Found while reviewing the interval operators for #1356. Not covered by #1352 (open null boundaries) or #1354 (tuple list unions), which touch different operators. Runtime-only fix - no generated code is affected.

alexzautke:
## Description
`Union<T>(CqlInterval<T>?, CqlInterval<T>?)` had two defects:

1. **Meeting intervals returned null.** The implementation only merged genuinely overlapping intervals, but the CQL specification returns null only when the intervals *neither overlap nor meet*: `Interval[2, 2] union Interval[3, 3]` must be `Interval[2, 3]`. Detecting "meets" requires the point-type specific `ToClosed` and `Successor` overloads, so the operator now dispatches on the runtime point type, normalizes open boundaries, and merges when `successor(first.high)` reaches `second.low`. The successor is only computed when the overlap check fails, so it is never asked for the successor of the maximum value.

2. **The right-starts-first branch dropped the left interval's tail.** That branch compared the left interval's high boundary *with itself* (`Compare(left.high, left.high) > 0`, never true), so whenever the right operand started first the method returned the right operand unchanged: `Interval[3, 10] union Interval[1, 5]` returned `Interval[1, 5]` instead of `Interval[1, 10]`. The rewrite orders the operands by start before merging, so the result no longer depends on argument order.

Null boundaries are treated as the minimum/maximum extremes, consistent with #1357. Runtime-only change: no generated code is affected and the integration-runner submodule is untouched.

## Related issues
Fixes #1359.

## Testing
- New `IntervalUnionTests` (10 tests): the 4 covering the two defects fail on `develop` (meets in both argument orders, date-typed meets, tail drop) and pass with the fix; 6 more pin unchanged behavior for overlap, containment, open-boundary normalization, unbounded high boundaries and null arguments.
- Full CoreTests (744) and CqlToElmTests (2,426) suites pass.
- Full MADiE integration run is byte-for-byte identical to the develop baseline (zero regressions, zero changes) — no measure in the corpus exercises these paths, consistent with the defects surfacing in direct SDK usage rather than the measure tests.

## FirelyTeam Checklist


baseTwo:
@copilot 
1. Add a release notes entry. docs/releases/vnext-release-notes.md wasn't touched, but the two sibling fixes in this exact file (#1354, #1350) both got a "Fixes" entry even though they were pure runtime behavior changes with no API signature change. This should follow the same pattern.
2. Either add a decimal/CqlQuantity meets test, or justify the >= check. The new meets check uses `Comparer.Compare(successor(firstHigh), secondLow, null) >= 0`, but the existing MeetsHelper elsewhere in this same file uses exact ==. For int/long/date-time types that distinction doesn't matter, but for decimal/CqlQuantity, Successor adds a fixed 0.00000001m epsilon — if two decimal intervals have a real gap smaller than that epsilon, this >= check would report "meets" when the stricter existing helper wouldn't. None of the 10 new tests exercise the decimal or CqlQuantity dispatch arms, so this is currently untested either way. A quick decimal-specific test (or just a comment explaining why >= is intentional here) would close that out.

baseTwo:
@copilot , merge from develop and fix conflicts

baseTwo:
> > copilot
> > 
> > 1. Add a release notes entry. docs/releases/vnext-release-notes.md wasn't touched, but the two sibling fixes in this exact file ([Union of lists with different tuple types generates `as IEnumerable<object>` casts that are always null #1354](https://github.com/FirelyTeam/firely-cql-sdk/issues/1354), [Choice-typed interval operands lowered as Interval<CqlDate> with hard casts, crashing CMS1173 #1350](https://github.com/FirelyTeam/firely-cql-sdk/issues/1350)) both got a "Fixes" entry even though they were...
> 
> Addressed in [bfc5cdc](https://github.com/FirelyTeam/firely-cql-sdk/commit/bfc5cdca24a9edc1a5b7e353dfcf1ec2862331ab).
> 
> * Added a `Fixes` entry for [Interval Union: meeting intervals return null and the right-starts-first branch drops the left interval's tail #1359](https://github.com/FirelyTeam/firely-cql-sdk/issues/1359) in `/docs/releases/vnext-release-notes.md`.
> * Updated interval union meets detection to require `successor(first.high) == second.low` (matching existing `MeetsHelper` semantics), and added decimal/CqlQuantity coverage in `IntervalUnionTests` (`Union_MeetingDecimalIntervals_Merges`, `Union_MeetingQuantityIntervals_Merges`, plus decimal gap coverage).

One correction to my own review, though: I went further and tested the actual sub-epsilon scenario I was worried about (a gap smaller than the 0.00000001m successor step, e.g. `Union([2,2], [2.000000005,3]))`. My assumption was that this could still slip through as a false "meets." It doesn't — but not because of the == fix per se. I traced it into `DecimalCqlComparer.CompareValues`, which truncates both operands to 8 decimal digits before comparing (CQL's Decimal type has exactly 8 digits of scale by spec). 2 and 2.000000005 both truncate to 2.00000000 and compare as literally equal — so the "sub-epsilon gap" I was worried about can't exist between two distinguishable CQL Decimal values in the first place; the type's own precision quantization rules it out. So the >=→== change is still the right, more consistent fix (matching MeetsHelper elsewhere in the file), but it wasn't closing a reachable bug — my original concern was based on treating the comparer as exact/infinite-precision, which it isn't.

Spec alignment: Confirmed against the CQL spec (Union returns null only when arguments neither overlap nor meet) — the fix's logic (only test `successor(first.high) == second.low` once the plain overlap check has already failed) is a faithful, minimal implementation of that rule.

baseTwo (approved):

PR #1367, opened 2026-07-14
Compile 'with'/'without' clauses as semi-joins

### LINKED ISSUE #1366: CQL 'with' relationship compiled as multiplying join instead of semi-join (over-counts source rows)
## Summary

The code generated for a CQL `with` relationship clause implements the semi-join as a `SelectMany` + `Select` that projects each matching relationship row back to the source element. This multiplies source rows: a source element with **N** matching relationship rows is emitted **N** times, instead of at most once.

A CQL `with` is a semi-join (existence filter). Each surviving source element must appear **exactly once**, regardless of how many relationship rows satisfy the `such that` condition. The correct semantics are `Where(source, s => relationship.Any(r => condition(s, r)))`.

This is a latent code-generation bug, independent of #1365 (Retrieve returning duplicate resources), though the two compound each other.

## Example

CQL (`PCMaternal."Delivery Encounter With Age Range"`):

```cql
"Encounter With Age Range" EncounterWithAge
  with ["Procedure": "Delivery Procedures"] DeliveryProcedure
  such that DeliveryProcedure.status = 'completed'
    and DeliveryProcedure.performed.toInterval() starts during EncounterWithAge.hospitalizationWithEDOBTriageObservation()
```

Generated C# (`PCMaternal-5.25.000.g.cs`, `Delivery_Encounter_With_Age_Range_Compute`):

```csharp
IEnumerable<Procedure> g_ = context.Operators.Where<Procedure>(e_, f_);   // matching procedures
Encounter h_(Procedure DeliveryProcedure) => EncounterWithAge;            // project back to the encounter
IEnumerable<Encounter> i_ = context.Operators.Select<Procedure, Encounter>(g_, h_);
return i_;
// ...
IEnumerable<Encounter> c_ = context.Operators.SelectMany<Encounter, Encounter>(a_, b_);
```

This is a cross-join projected back to the source, not a semi-join:

| matching relationship rows | emitted | correct? |
|---|---|---|
| 0 | dropped | ✅ |
| 1 | once | ✅ (accidentally) |
| N ≥ 2 | N times | ❌ should be once |

## Expected

A `with` relationship should compile to an existence filter over the source, emitting each source element at most once, e.g.:

```csharp
bool exists = context.Operators.Where<Procedure>(e_, f_).Any();
// keep EncounterWithAge only if exists
```

i.e. a `Where` over the source, not a `SelectMany` that fans out over relationship rows.

## Notes

- The defect is in the ELM → C# compilation of query `with` relationships (not in the generated `.g.cs` file itself, which is just the symptom). The exact compiler location has not yet been pinned down.
- Discovered while investigating #1365 (`CMS0334FHIRPCCesareanBirth` Initial Population = 2 instead of 1). Fixing #1365 resolves that specific test because it removes the duplicate procedure, but this bug remains: any `with` where a single source element genuinely has two or more matching relationship rows will still be over-counted.

alexzautke:
## Description
A CQL `with` relationship clause is a semi-join (existence filter): each source element must be emitted at most once, no matter how many related elements satisfy the `such that` condition. The compiler previously emitted `SelectMany(source, s => related.Where(suchThat).Select(r => s))`, which emits a source element once per matching related element — over-counting whenever an element has two or more matches.

`without` had related defects: it was compiled as `Except(source, <the multiplied matches>)`, whose set semantics also deduplicated the surviving source elements.

Both clauses now compile to an existence filter over the source, exactly as proposed in the issue:

```csharp
// with
Where(source, s => Exists(Where(related, r => suchThat)))
// without
Where(source, s => Not(Exists(Where(related, r => suchThat))))
```

Changes:
- `Cql/Cql.Compiler/ExpressionBuilderContext.cs`: `WithToSelectManyBody` replaced by `WithToExistenceCheck`; the `Query` relationship handling now applies a single `Where` per relationship clause for both `with` and `without`.
- `LibrarySets/dqm-content-qicore-2025/Extracted/CSharp/CMS56FHIRFuncStatHipReplacement-1.0.000.g.cs`: golden file regenerated (the RR23 golden corpus contains no relationship clauses and regenerates byte-identical).
- `Demo/Measures.dqm-content-qicore-2025/CSharp/*.g.cs`: demo measure corpus (used by the integration runner) regenerated with PackagerCLI from the checked-in ELM; 61 of 89 libraries contain relationship clauses and change, the rest regenerate byte-identical.
- Submodule `submodules/Firely.Cql.Sdk.Integration.Runner` bumped to [`82d7eac79`](https://github.com/FirelyTeam/Firely.Cql.Sdk.Integration.Runner/commits/claude/with-clause-semijoin-passes) (branch `claude/with-clause-semijoin-passes`, see FirelyTeam/Firely.Cql.Sdk.Integration.Runner#102), which refreshes `passed.jsonl` with 81 additional passing cases (see Testing).

Regenerating `PCMaternal` (the library quoted in the issue) produces the expected shape for `Delivery_Encounter_With_Age_Range_Compute`: a `Where` over the source encounters whose predicate is `Exists` over the matching procedures.

## Related issues
Fixes #1366.

## Testing
- New `Cql/CoreTests/RelationshipClauseTests.cs`: end-to-end tests (CQL → ELM → C# → execution) covering the issue's semantics table — 0 matches drops the element, 1 match emits once, N ≥ 2 matches emits once — plus `without`, duplicate-preservation for both clauses, and chained `with` clauses. Three of the five tests fail against the previous implementation.
- Full `CoreTests` suite (including the C# generation golden tests): 742 passed on net10.0; key tests also verified on net8.0. `CqlToElmTests`: 2427 passed.
- **Integration suite** (all 3857 cases, `Write` mode, run twice): the fixed SDK passes 2447 cases with zero regressions against the previous `passed.jsonl` (2366 entries). A baseline run against develop *without* this fix passes the identical 2447 cases — so the fix flips no integration outcome, and the 81 newly listed cases (13 measures, e.g. CMS71/CMS104/CMS190) come from SDK fixes already on develop that predate the last `passed.jsonl` refresh. The refreshed list was re-validated in `ReadAndOnlyRunPreviousPasses` mode: 2447/2447 pass.
- `Cql-Sdk.slnf` and the demo measure projects build cleanly. (`Test.Measures.Demo` has one pre-existing, unrelated failure caused by FHIR Library resources that are not checked into the repo.)

## FirelyTeam Checklist

https://claude.ai/code/session_01DDy7nDXVN5AYZMPa7X3eB9

### DIFF
--- a/Cql/CoreTests/RelationshipClauseTests.cs
+++ b/Cql/CoreTests/RelationshipClauseTests.cs
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2026, Firely, NCQA and contributors
+ * See the file CONTRIBUTORS for details.
+ *
+ * This file is licensed under the BSD 3-Clause license
+ * available at https://raw.githubusercontent.com/FirelyTeam/firely-cql-sdk/main/LICENSE
+ */
+
+#nullable enable
+
+using Hl7.Cql.CqlToElm;
+using Hl7.Cql.CqlToElm.Toolkit;
+using Hl7.Cql.CqlToElm.Toolkit.Extensions;
+using Hl7.Cql.Fhir;
+using Hl7.Cql.Invocation.Toolkit;
+using Hl7.Cql.Invocation.Toolkit.Extensions;
+using Hl7.Cql.Runtime;
+
+namespace CoreTests;
+
+/// <summary>
+/// Tests for the query 'with'/'without' relationship clauses (see issue #1366).
+/// A 'with' is a semi-join and a 'without' an anti-semi-join: each source element must be
+/// emitted at most once, no matter how many related elements satisfy the 'such that' condition,
+/// and duplicates in the source must be preserved.
+/// </summary>
+[TestClass]
+public class RelationshipClauseTests
+{
+    private static LibrarySetInvoker _invoker = null!;
+    private static CqlVersionedLibraryIdentifier _libraryIdentifier;
+
+    [ClassInitialize]
+    public static void Initialize(TestContext context)
+    {
+        var cqlLibrary = CqlLibraryString.Parse(
+            """
+            library RelationshipClauseTest version '1.0.0'
+
+            define "Sources": { 1, 2, 3 }
+            define "SourcesWithDuplicates": { 1, 1, 3 }
+            define "Related": { 10, 20, 30 }
+
+            // 1 matches 20 and 30, 2 matches 10, 3 matches nothing
+            define "With Clause":
+              "Sources" S
+                with "Related" R
+                  such that (S = 1 and R > 10) or (S = 2 and R = 10)
+
+            define "Without Clause":
+              "Sources" S
+                without "Related" R
+                  such that (S = 1 and R > 10) or (S = 2 and R = 10)
+
+            define "With Clause Preserves Duplicates":
+              "SourcesWithDuplicates" S
+                with "Related" R
+                  such that R = 10 * S
+
+            define "Without Clause Preserves Duplicates":
+              "SourcesWithDuplicates" S
+                without "Related" R
+                  such that R = 100 * S
+
+            define "Multiple With Clauses":
+              "Sources" S
+                with "Related" R1
+                  such that R1 = 10 * S
+                with "Related" R2
+                  such that R2 > 10 * S
+            """);
+        _libraryIdentifier = cqlLibrary.LibraryIdentifier;
+        _invoker = new CqlToolkit()
+                   .AddCqlLibraries(cqlLibrary)
+                   .CreateLibrarySetInvoker();
+    }
+
+    [ClassCleanup]
+    public static void Cleanup()
+    {
+        _invoker?.Dispose();
+    }
+
+    private static IEnumerable<int?> Invoke(string definition) =>
+        ((IEnumerable<int?>)_invoker.InvokeLibraryDefinition(FhirCqlContext.ForBundle(), _libraryIdentifier, definition)!);
+
+    [TestMethod]
+    public void With_EmitsEachSourceElementAtMostOnce()
+    {
+        // Element 1 has two matching related elements but must be emitted only once.
+        Invoke("With Clause").Should().Equal(1, 2);
+    }
+
+    [TestMethod]
+    public void Without_KeepsOnlySourceElementsWithNoMatch()
+    {
+        Invoke("Without Clause").Should().Equal(3);
+    }
+
+    [TestMethod]
+    public void With_PreservesDuplicateSourceElements()
+    {
+        // Both 1s match related element 10 and must both be kept.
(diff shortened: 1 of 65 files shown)


baseTwo (commented):
Two write-up (PR description) issues worth fixing before merge — the code change itself looks correct:

1. **Factual error re: the "byte-identical" golden file claim.** The description says `LibrarySets/dqm-content-qicore-2025/Extracted/CSharp/CMS56FHIRFuncStatHipReplacement-1.0.000.g.cs` "regenerates byte-identical" because "the RR23 golden corpus contains no relationship clauses." That's not accurate — diffing this file directly shows the same `SelectMany`→`Where`/`Exists` transformation as every other changed measure (e.g. `bool? b_(Procedure THAProcedure)` replacing `IEnumerable<Procedure> b_(...)`, `SelectMany`→`Where` at the call site). Please correct the description so it doesn't misrepresent the blast radius of this change.

2. **Missing `docs/releases/vnext-release-notes.md` entry.** This repo has an established convention (see commits `66983181e`, `58eeb6497`) of adding a one-line `## Fixes` entry with the issue number for behavior-affecting fixes like this one, since it changes the output of previously-generated measure code. Suggest adding something like:
   > - CQL `with`/`without` relationship clauses now compile as an existence filter (semi-join/anti-semi-join) instead of `SelectMany`/`Except`, so a source element with multiple matching related elements is emitted exactly once instead of duplicated (or spuriously deduplicated by `Except`'s set semantics for `without`) (#1366).

baseTwo:
> Two write-up (PR description) issues worth fixing before merge — the code change itself looks correct:
> 
> 1. **Factual error re: the "byte-identical" golden file claim.** The description says `LibrarySets/dqm-content-qicore-2025/Extracted/CSharp/CMS56FHIRFuncStatHipReplacement-1.0.000.g.cs` "regenerates byte-identical" because "the RR23 golden corpus contains no relationship clauses." That's not accurate — diffing this file directly shows the same `SelectMany`→`Where`/`Exists` transformation as every other changed measure (e.g. `bool? b_(Procedure THAProcedure)` replacing `IEnumerable<Procedure> b_(...)`, `SelectMany`→`Where` at the call site). Please correct the description so it doesn't misrepresent the blast radius of this change.
> 2. **Missing `docs/releases/vnext-release-notes.md` entry.** This repo has an established convention (see commits `66983181e`, `58eeb6497`) of adding a one-line `## Fixes` entry with the issue number for behavior-affecting fixes like this one, since it changes the output of previously-generated measure code. Suggest adding something like:
>    > * CQL `with`/`without` relationship clauses now compile as an existence filter (semi-join/anti-semi-join) instead of `SelectMany`/`Except`, so a source element with multiple matching related elements is emitted exactly once instead of duplicated (or spuriously deduplicated by `Except`'s set semantics for `without`) (#1366).

@copilot , please address these

baseTwo (approved):

PR #1368, opened 2026-07-14
Fix duplicate Retrieve results when multiple codings match

### LINKED ISSUE #1365: Retrieve by value set returns duplicate resources when multiple codings match (inflates population counts)
## Summary

`Retrieve` by value set returns the **same resource multiple times** when the resource has more than one coding that is a member of the value set. This causes downstream population counts to be inflated.

Discovered while tracing the `CMS0334FHIRPCCesareanBirth` measure test: `Initial Population` returns **2** for a patient with a single delivery encounter, where the expected count is **1**.

## Root cause

`Cql/Cql.Firely/IndexedBundle.cs` yields a candidate once **per matching coding** instead of once per resource:

```csharp
// IndexedBundle.cs:45
public IEnumerable<T> FilterByType<T>(Predicate<Coding> filter)
{
    var candidates = FilterByType<T>();
    foreach (var candidate in candidates)
    {
        if (candidate is ICoded codedCandidate)
        {
            var codings = codedCandidate.ToCodings();
            foreach (var coding in codings)
                if (filter(coding)) yield return candidate;   // <-- yields once PER matching coding
        }
    }
}
```

The `getCodes` overload at `IndexedBundle.cs:60` has the identical flaw.

A retrieve returns a *set of resources*; a resource must appear at most once regardless of how many of its codes match the value set.

## Reproduction (CMS0334FHIRPCCesareanBirth)

The test Procedure fixture has three codings, two of which are members of the "Delivery Procedures" value set (`2.16.840.1.113762.1.4.1045.59`):

| system | code | in value set? |
|---|---|---|
| `http://www.cms.gov/Medicare/Coding/ICD10` | `10D00Z0` | ✅ |
| `http://snomed.info/sct` | `4504004` | ✅ |
| `http://www.icd10data.com/icd10pcs` | `10D00Z0` | ❌ (system not in VS) |

Two codings match → `Retrieve<Procedure>` yields the same Procedure twice. That duplicate flows through the `with ["Procedure": "Delivery Procedures"]` semi-join in `PCMaternal."Delivery Encounter With Age Range"`, and the single Encounter is emitted twice → `Initial Population` = 2 (expected 1).

## Proposed fix

Yield each candidate at most once if **any** coding matches:

```csharp
public IEnumerable<T> FilterByType<T>(Predicate<Coding> filter)
{
    foreach (var candidate in FilterByType<T>())
        if (candidate is ICoded codedCandidate && codedCandidate.ToCodings().Any(c => filter(c)))
            yield return candidate;
}
```

(and the analogous change to the `getCodes` overload).

## Secondary observation (separate issue candidate)

The generated code for a CQL `with` relationship (e.g. `PCMaternal-5.25.000.g.cs` `Delivery_Encounter_With_Age_Range_Compute`) implements the semi-join as `SelectMany` + `Select` (one source row per matching relationship row) rather than an exists-based filter. A `with` clause should never multiply source rows even if the relationship yields multiple matches. Fixing the retrieve duplication above resolves this test, but this codegen behavior is a latent bug worth a separate look.

alexzautke:
## Description
`IndexedBundle.FilterByType<T>` yielded a candidate once *per matching coding* instead of once per resource, so a resource with several codings in the same value set appeared multiple times in `Retrieve` results, inflating downstream population counts (observed in the `CMS0334FHIRPCCesareanBirth` measure, where a single delivery encounter was counted twice in the Initial Population).

Both filter overloads — the primary-code-path one (`Predicate<Coding>`) and the explicit-code-property one (`Predicate<Coding>, Func<T, IEnumerable<Coding>>`), which had the identical defect — now yield each candidate at most once when any of its codings matches, using `.Any(coding => filter(coding))`.

## Related issues
Fixes #1365.

## Testing
- Added two regression tests in `DataSourceTests` covering both code paths (primary code path via `ICoded`, and explicit `codeProperty`), using an Observation with three codings of which two match the filter — asserting it is returned exactly once.
- Verified both new tests fail against the unfixed code (each reproducing the duplicate) and pass with the fix.
- Full CoreTests suite (net10.0): 739 passed, 0 failed, 9 skipped (pre-existing skips).

## FirelyTeam Checklist

https://claude.ai/code/session_01BvCvWDbnWWt4m5vW4d7vq2

---
_Generated by [Claude Code](https://claude.ai/code/session_01BvCvWDbnWWt4m5vW4d7vq2)_


baseTwo:
@copilot Please add an entry for this fix to `docs/releases/vnext-release-notes.md` under `## Fixes`, following the same convention used for the #1366 fix. Suggested wording:

- `IndexedBundle.FilterByType<T>` (used by `Retrieve`) no longer yields a candidate resource more than once when several of its codings match the filter, preventing inflated `Retrieve` results / population counts for resources with multiple matching codings (#1365).

baseTwo:

baseTwo (approved):

Issue #1389, opened 2026-07-21
Periodically resync CQL/FHIR conformance tests and condensed specs against upstream


baseTwo:
## Summary

Three spec-derived assets that AI-assisted spec-conformance work in this repo relies on were either imported once with no resync process, or don't exist locally at all:

1. **Vendored CQL conformance test suite** — `Cql/CqlToElmTests/Input/DQIC/*.xml`
2. **Condensed CQL spec for AI parsing** — `spec/condensed/*.md` (generated by `tools/condense_spec/condense_spec.py`)
3. **FHIR R4 specification** (hl7.org/fhir) — has **no local cache at all**, unlike the CQL spec

Items 1 and 2 have already drifted from upstream — this isn't a hypothetical future risk.

## Impact

- Silent test-suite drift means we can't tell whether a gap versus the official CQL test suite is an intentional decision or an accidental omission from the original 2023 import.
- The condensed CQL spec has no refresh trigger, so any future CQL errata release goes unnoticed and AI-assisted spec-conformance work ends up reasoning against an outdated copy without anyone realizing it.
- FHIR spec work (e.g. `Measure` resource generation in `Cql.Packaging`) currently has no local reference at all, so any AI-assisted analysis has to fetch pages from hl7.org live, ad hoc, per conversation — no shared, reviewable snapshot the way the CQL spec has one.

### DQIC test suite drift (verified 2026-07-21)

`Input/DQIC/*.xml` was first committed 2023-06-09 and has never been systematically resynced against the official suite. I downloaded the current `https://cql.hl7.org/tests.zip` and diffed it against the local files:

| File | Official test count | Local test count | Notes |
|---|---|---|---|
| `CqlAggregateTest.xml` | 2 | **missing entirely** | Whole file never imported |
| `CqlDateTimeOperatorsTest.xml` | 295 | 293 | Missing `DateTimeComponentFromTimezone`, `TimeDurationBetweenHourDiffPrecision` |
| `CqlComparisonOperatorsTest.xml` | 183 | 190 | 7 tests added locally, beyond upstream |
| `CqlListOperatorsTest.xml` | 207 | 208 | 1 test added locally |
| `CqlStringOperatorsTest.xml` | 81 | 88 | 7 tests added locally (matches the Split-semantics fix work) |
| `CqlAgeTest.xml` | — | 8 | Entirely custom, doesn't exist upstream |
| all other files | — | — | Match exactly |

[...]

Issue #1193, opened 2026-02-19
CQL Specification Conformance Issues


baseTwo:
## Important 
* The issues listed here was generated by AI (see https://github.com/FirelyTeam/firely-cql-sdk/issues/1190) by letting it compare the codebase to the CQL specification (see /spec/condensed)
* Each issue must be evaluated for accurateness by a human, before being picked up

### Critical Priority

- https://github.com/FirelyTeam/firely-cql-sdk/issues/1194
  - **Bug:** Substring uses wrong length calculation
  - **Impact:** Can throw exceptions or return incorrect results
  - **Fix:** Change `source.Length` to `source.Length - startIndex`

- https://github.com/FirelyTeam/firely-cql-sdk/issues/1206
  - **Bug:** Date.Add() allows time-based units (hours, minutes, seconds, milliseconds)
  - **Impact:** Violates spec line 3226, allows semantically invalid operations
  - **Fix:** Throw ArgumentException for time-based units on Date values

### High Priority

- https://github.com/FirelyTeam/firely-cql-sdk/issues/1195
  - **Bug:** Missing bounds check, throws on oversized suffix
  - **Impact:** Throws exceptions for valid inputs
  - **Fix:** Add bounds check and empty string handling

- https://github.com/FirelyTeam/firely-cql-sdk/issues/1196
  - **Spec Violation:** Null elements should be considered equal
  - **Impact:** Incorrect equality results for lists with nulls
  - **Fix:** Treat `null == null` as `true` in list comparisons

- https://github.com/FirelyTeam/firely-cql-sdk/issues/1197
  - **Test Issue:** Implementation is CORRECT, tests are WRONG
  - **Impact:** None - just fix test expectations
  - **Fix:** Update test XML to expect `-1.0` instead of `0.0`

- https://github.com/FirelyTeam/firely-cql-sdk/issues/1198
  - **Bug:** Returns `0` instead of `null` for fractional results
  - **Impact:** Loses precision, violates spec
  - **Fix:** Return `null` when result cannot be represented
  - **Affects:** 10 skipped tests

- https://github.com/FirelyTeam/firely-cql-sdk/issues/1207
  - **Bug:** UCUM units "a" (year) and "mo" (month) use fractional days instead of calendar math
  - **Impact:** Produces incorrect results, loses value magnitude with Math.Sign()
  - **Fix:** Merge "a"/"mo" with "year"/"month" to use calendar semantics

### Medium Priority

[...]

PR #1391, opened 2026-07-21
Add DQIC test suite comparator tool

baseTwo:
## Description
Part of #1389: adds `tools/dqic_sync/compare_dqic_tests.py`, a review tool that compares the vendored `Cql/CqlToElmTests/Input/DQIC/*.xml` against the current official suite at https://cql.hl7.org/tests.zip.

- Classifies every `<test name="...">` (matched by name, per file) as missing locally, a Firely-only addition, or differing from upstream.
- For every difference, runs `git log -S<test name>` against the local file so a reviewer can see which commit last touched that specific test — the tool surfaces evidence, it doesn't guess intent.
- Does **not** modify any XML files. The official suite's own `tests.html` page states its tests are "informative, not normative" — many local differences are intentional spec-conformance fixes (see epic #1193) where upstream is the one that's wrong, so this has to be a manual review + judgment call, not an auto-sync.
- Handles a real data-quality bug found in the official zip itself: `CqlArithmeticFunctionsTest.xml` upstream has a malformed stray `</group>` tag that breaks strict XML parsing. The tool catches per-file parse errors and reports them separately instead of crashing the whole comparison.

Ran it against the current suite as part of this ticket's investigation — found 1 file (`CqlAggregateTest.xml`, 2 tests) and 2 individual tests (`DateTimeComponentFromTimezone`, `TimeDurationBetweenHourDiffPrecision`) missing entirely, 18 Firely-only additions, and 111 tests with differing expectations (heavily concentrated in DateTime timezone-handling and List null-equality semantics — both look like deliberate historical decisions, not accidental drift, but that's exactly the kind of thing this tool exists to surface for a human to confirm rather than silently overwrite).

## Testing
- Ran end-to-end against a live download of `tests.zip` and produced a correct, reviewed report (initial run had a bug misclassifying the malformed upstream file as "local-only" — fixed and re-verified).
- `python -m py_compile` / `ast.parse` clean.

## Related issues
Part of #1389.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -5,6 +5,7 @@ Utility scripts for the Firely CQL SDK.
 ## Directory Structure
 
 - **condense_spec/** - CQL Specification HTML to Markdown converter
+- **dqic_sync/** - Compares the vendored CQL conformance test suite against the official upstream suite
 - **XsdToCSharpConverter/** - XSD to C# code generation tool
 - **XsdToCSharpConverterTests/** - Tests for XSD converter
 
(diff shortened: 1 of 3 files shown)


baseTwo:
Reviewed all 6 of Copilot's comments — agreed with every one:

1. Local XML parse failures were silently dropped and mid-report showed up as "missing locally," which is misleading (a broken file ≠ a never-imported one).
2. `USER_AGENT` was copy-pasted from `condense_spec.py` and didn't identify this tool.
3–5. Canonical test content (joined with `\n` for comparison) was rendered raw inside single-line Markdown inline code spans, breaking/squashing the report's formatting.
6. `identical_count` was computed and never used.

@copilot's agent pushed a fix for all 6 while I was independently writing the same fixes locally — compared the two implementations (functionally equivalent: same UA rename, same local-error tracking, same newline-escaping, same identical-count column) and reconciled onto its commit rather than push a duplicate. Re-ran the tool end-to-end against a live copy of `tests.zip` to confirm the reconciled version still produces a correct report. Resolving all 6 threads.

alexzautke (approved):

Issue #1342, opened 2026-07-10
Includes/IncludedIn compute rightElementType from the left operand (dead mismatch check)


ewoutkramer:
## Where

`Cql\Cql.Compiler\ExpressionBuilderContext.cs`, `Includes` (~line 1424) and `IncludedIn` (~line 1461), and faithfully preserved in the phase-4 IR port (`Cql\Cql.Compiler\Ir\IrExpressionBuilderContext.Operators.cs`, marked `NOTE(phase4)`).

## What

In the nested-list branch of both methods, `rightElementType` is computed from `left.Type` instead of `right.Type` (an apparent copy-paste of the `leftElementType` line). As a result the element-type mismatch check compares the left element type against itself and can never fire.

## Impact

The guard is dead code: mismatched operand element types fall through to operator binding, which either resolves via conversions or fails there with a less specific error. No known incorrect binding results from this — the effect is a missing early diagnostic.

## Why not fixed yet

Same policy as #1341: the Linq.Expressions removal requires old and new pipelines to behave bug-for-bug identically until golden parity is proven (phase 5). Fix once, in one place, after phase 6 deletes the old builder — and add a test with deliberately mismatched element types.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1341, opened 2026-07-09
Operator binder generic inference indexes methodParameters with the wrong loop variable


ewoutkramer:
## Where

`Cql\Cql.Compiler\CqlOperatorsBinder.Bind.cs`, inside `ResolveMethodInfosWithPotentialArgumentConversions` -> `GetGenericTypeArgs` (~line 228), and faithfully preserved in the phase-3 IR port (`Cql\Cql.Compiler\Ir\IrCqlOperatorsBinder.Bind.cs`, marked `NOTE(phase3)`).

## What

The local function probes the first two *arguments* for a usable generic type argument, looping with `argIndexForGenericMethod` — but the parameter it inspects is indexed with `i`, which is the *outer* retry-pass index (the "try once with all arguments, once without a trailing null constant" loop), not the argument index:

```csharp
for (int argIndexForGenericMethod = 0;
     argIndexForGenericMethod < Math.Min(args.Length, 2);
     argIndexForGenericMethod++)
{
    var argType = args[argIndexForGenericMethod].Type;
    var parameterType = methodParameters[i].ParameterType;   // <-- `i` is the outer retry index
    ...
}
```

So while probing argument 0 and argument 1, both probes check the genericity of `methodParameters[0]` on the first pass and `methodParameters[1]` on the retry pass, regardless of which argument is being probed. Almost certainly `methodParameters[argIndexForGenericMethod]` was intended.

## Impact

Low in practice: this only influences which *candidate* generic type arguments are attempted; every candidate still passes through `MakeGenericMethod` + `TryBindArguments`, which validates the full signature, so a wrong candidate is discarded rather than mis-bound. The realistic failure mode is a **missed binding** ("no suitable method could be bound") for overload shapes where the generic parameter is not in the position the wrong index happens to check — not an incorrect binding.

## Why not fixed yet

The Linq.Expressions removal (see `docs/linq-expression-removal-plan.md`) requires the old and new binders to behave bug-for-bug identically until golden parity is proven (phase 5). Changing inference behavior mid-migration would add noise to the parity diff. Fix once, in one place, after phase 6 deletes the old binder — at that point also remove the `NOTE(phase3)` marker and add a regression test for an overload shape that the current indexing misses.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1392, opened 2026-07-21
feat!: delete the Expression-based pipeline — the typed IR is the pipeline (phase 6)

### LINKED ISSUE #1345: Old operator binder: Retrieve dispatch can index out of range, plus three diagnostic defects (fixed in the IR port only)
## Background

Copilot review of the phase-3 IR binder port (#1340) surfaced four defects that turned out to be verbatim heritage from the **old** `CqlOperatorsBinder`. Per the migration policy (see `docs/linq-expression-removal-plan.md` and #1341), they were fixed in the IR copy only (commit `0081270c2`); the old binder was deliberately left untouched so the two pipelines stay bug-for-bug comparable until golden parity is proven.

This issue exists so the fixes can be applied to the **old** binder if that ever becomes necessary — i.e. if a release ships from the old pipeline before phase 6 deletes it, or if the migration is paused.

## The four defects in the old binder

1. **`CqlOperatorsBinder.cs` ~line 75** — the name-dispatch pattern `("Retrieve", 0, >=3) => Retrieve(args[0], args[1], args[2], args[3])` accepts 3 arguments but indexes `args[3]`, throwing `IndexOutOfRangeException` for a 3-argument call. Fix: guard `>=4`. (Only the crash path changes; all working call sites pass 4 arguments.)
2. **`CqlOperatorsBinder.Specific.cs` ~line 102** — `ResolveValueSet` throws "Expression should be a constant CqlValueSet", but the code actually checks for a `New` (constructor) expression of `CqlValueSet` — misleading when debugging binding failures.
3. **`CqlOperatorsBinder.Specific.cs` ~line 194** — the Retrieve `PropertyInfo` validation message says "Second parameter", but `codePropertyExpression` is the third parameter.
4. **`CqlOperatorsBinder.Specific.cs` ~line 276** — `Where` captures a misspelled, unused pattern variable `lamdaExpr`.

## Resolution paths

- **Normal course**: phase 6 deletes the old binder; the IR copy already contains all four fixes; close this issue then.
- **If needed earlier**: apply the four fixes above to the old binder. Items 2-4 are diagnostics-only. Item 1 changes only a crash path, but for strict old-vs-new parity during phase 5 the same guard (`>=4`) is already what the IR copy uses — so fixing the old side actually *reduces* divergence.

Related: #1341, #1342, #1343 (other preserved-as-is findings from the migration).

ewoutkramer:
Phase 6 of the [Linq.Expressions removal plan](../blob/linq-expr-removal-phase6/docs/linq-expression-removal-plan.md): with byte-identical output proven across all three corpora in phase 5, this deletes the Expression-based pipeline outright. The typed IR is now the only pipeline (**−8,097 lines**, 48 files deleted).

## What's deleted

- `ExpressionBuilderContext` + all partials, `CqlOperatorsBinder.*`, `CqlContextBinder` (the ELM→Expression builder and binders)
- The 5 custom `Expression` subclasses (`Expressions\`), `CqlExpressions.cs`, `IExpressionMutator`
- The 4 visitor passes (`Visitors\`: `RedundantCastsTransformer`, `SimplifyExpressionsVisitor`, `RenameVariablesVisitor`, `LocalVariableDeduper`) and `DebugViewHelper`
- `LibrarySetCSharpCodeGenerator` + `LambdaDefinitionWriter` (the old writer)
- `CqlDefinition` / `CqlLambdaDefinition`

The old pipeline's sources remain reachable at its final commit, 85207efd5, which is the permanent parity reference.

## Decisions to review

- **`UseIrPipeline` is removed entirely** (config property + `PublicAPI.Unshipped.txt` entries) rather than deprecated: the flag never shipped in a release, so flipping the default and removing it are the same act.
- **No `GeneratedCodeAttribute` tool-version bump** — a deliberate deviation from the plan doc, which anticipated bumping at the flip. The IR pipeline's output is byte-identical to the old writer's, so a version bump would claim a difference that doesn't exist. Happy to bump anyway if you prefer the "different generator, different version" reading.
- **Phase-6 `FIXME` sites resolved in-flight**: `CannotBindToCqlOperatorError` generalized from `Expression[]` to `Type[]`; `CqlOperatorsMethodsCache` relocated to `Cql.Compiler\Ir\`; the duplicated assign-to-type helpers consolidated onto `IrExpressionExtensions`; `GeneratorToolVersion` moved to `IrLibrarySetCSharpCodeGenerator`.

## Emitter fixes exposed by the deletion

Deleting the old pipeline rewired every test (including the XML-driven CQL test runner, which previously used `Expression.Compile()`) through the IR → C# → assembly path, exposing six print gaps that the golden corpora never exercised — five small ones (constant `is null` folds, `default(T)?.member` receivers, tuple-typed member-init redirect, printed-type recursion through boxing-elided casts, zero-length array bounds) and one that took real adjudication:

The inline-conditional outer-cast trigger had been re-derived from its failure mode instead of ported verbatim, and overfired. A dual-commit comparison of the generated C# for all 382 HEDIS 2025 libraries (this branch vs. 85207efd5) showed 13 diverging files, then 5 after a first narrowing; the literal port of the old writer's rule (branch node type ≠ conditional type) gives a clean 0/382. That literal rule exposed one shape no corpus can contain — the XML test harness wraps `IsNull` around the expression under test, and for constant conditionals that prints `(a > b ? 5 : 10) is null`, whose ternary has no nullable natural type (CS0037). The old writer emitted the same broken text for this shape (which is why it can't occur in any compiling corpus), so `PrintBinary` gained a narrowly-scoped safety net: an `is null` operand that is an inline ternary of bare constant literals is cast to its IR type. It is documented as having no old-writer counterpart.

## Submodule

`Firely.Cql.Sdk.Integration.Runner` → `2a8fb74e`: the golden-*parity* test is retired (its comparison target no longer exists); the guard is now `HEDIS_2025_CompilesToAssemblies`, exercising the IR pipeline end to end over the full corpus.

## Deferred follow-ups (post-merge, tracked in the plan doc checklist)

- Drop the now-redundant `Ir` prefix from type names (`IrLibrarySetCSharpCodeGenerator` etc.)
- Fix the tracked upstream bugs in one place: #1341, #1342; close #1345
- Revisit `IrCodeDefinition.ReturnType` (parity-preserved `typeof(CqlCodeDefinition)`)
- Replace the hoisted zero-parameter local functions for conditional chains with native `if`/`else`

## Verification

- CoreTests **793/0**, CqlToElmTests **2453/0** (XML-driven runner 1493/0), at the final commit
- Generated C# for all **382 HEDIS 2025 libraries byte-identical** to the old pipeline's output at 85207efd5, and all 382 compile to assemblies
- RR23 + CMS56 golden files regenerate byte-identically (part of CoreTests)
- develop drift re-checked: no compiler-directory changes since the merge-base, nothing to port


baseTwo:
Awesome!!!

PR #1401, opened 2026-07-23
feat!: drop the three ledgered old-writer quirks from the generated C#

ewoutkramer:
Second post-migration cleanup PR (checklist in `docs/linq-expression-removal-plan.md`), stacked on #1398 — **the first output-changing one**, so it carries the `GeneratedCodeAttribute` version bump deferred at phase 6 and regenerated goldens as the review artifact.

Drops the three old-writer quirks that were preserved bug-for-bug for golden parity and ledgered for exactly this PR:

## 1. Burned-letter naming gaps — with a real dedup win

The old pipeline allocated a name for every deduplicated statement before removing it, leaving gaps in the `a_`, `b_`, … sequence. Worse than cosmetic: because the burned name keyed the duplicate for its *parents*, identical parent expressions never deduplicated with each other. Duplicates now reuse the original's local outright — names stay contiguous, **and previously-blocked deduplication kicks in**: 52 redundant pure computations collapse in CMS56 alone, **1,028 across the HEDIS 2025 corpus** (e.g. the same `ToDateTime`/`DateFrom`/`ToConcept` call computed 2–6× in one definition now computes once).

## 2. The stray `;` after a conditional function's final `else` block

Gone, along with the parameter that existed to thread it through.

## 3. The accidental `as object` survivors

Casts to object never print now — the conversion is always implicit in C# (boxing or implicit reference conversion). The old pipeline stripped these too, *except* the ELM `as`-operator casts its single-pass `RedundantCastsTransformer` couldn't see (they materialized lazily, after the pass) — a visitor-ordering accident. The `CodeCast.FromCqlAsOperator` flag, which existed solely to replicate that accident, is deleted.

## Version bump

`GeneratorToolVersion` 5.1.2.0 → **5.1.3.0** (patch: only expression internals change, no shape the invoker binds to; stays inside `LibraryInstanceInvoker_5_0`'s supported `[5.1.0.0, 5.2.0.0)` range, so no new invoker).

## How the output diff was reviewed

Not by eyeballing 1,300 changed lines — by **classification**: normalize local names (`\b[a-z]{1,2}_\b` → `X`), fold the five expected transforms (version attribute, `};` → `}`, ` as object` removal, its parenthesis fallout, renumbering), then multiset-compare old vs new lines per file.

- RR23 + CMS56 goldens (in this diff): every changed line classified; the 52 removed-only lines are all collapsed duplicate assignments.
- HEDIS 2025 (dumped locally, both sides): **382/382 files fully classified, zero unexplained lines**, 1,028 collapsed duplicates. All 382 still compile to assemblies (the CI guard re-proves this).

## Also

- Integration-runner submodule bumped to runner main's merge of FirelyTeam/Firely.Cql.Sdk.Integration.Runner#105 (contains the previously-referenced branch commit).
- Plan-doc checklist updated. Remaining cleanup: the native `if`/`else` PR (next), under the agreed readability license.

Verified: CoreTests 823/0 (including the regenerated goldens), CqlToElmTests 2455/0, HEDIS 382/382.


baseTwo:
@copilot Please check:

1. That `GeneratorToolVersion` in `Cql/CodeGeneration.NET/_CODE GENERATOR VERSION_.cs` is bumped again at the patch level to reflect this PR's emitter changes.
2. That all checked-in `*.g.cs` files affected by these emitter changes (e.g. under `Cql/CoreTests/CSharp/`, `Demo/Measures.Authoring/CSharp/`, `Demo/Measures.Demo/CSharp/`, `Demo/Measures.dqm-content-qicore-2025/CSharp/`, in addition to `LibrarySets/`) are regenerated and committed — not just the ones where only the `GeneratedCodeAttribute` version string changes.

baseTwo (approved):

PR #1404, opened 2026-07-25
feat!: emit native if/else statements for multi-branch conditionals

ewoutkramer:
The last ledgered post-migration cleanup (checklist in `docs/linq-expression-removal-plan.md`), done under the agreed readability license: the old writer's shapes were heuristics, not contracts, so this doesn't replicate the zero-parameter-function form — it replaces it with what a human would write.

## What changes in the generated C#

Multi-branch conditionals (`if`/`case … when`) emitted as invoked zero-parameter local functions:

```csharp
IEnumerable<Claim> a_() {
    if ((this.Check_data_source(context)) ?? false)
    { ... return c_; }
    else
    { ... return d_; }
}

return a_();
```

now emit as native statements — in tail position the wrapper disappears entirely and branches return directly:

```csharp
if ((this.Check_data_source(context)) ?? false)
{ ... return b_; }
else
{ ... return c_; }
```

Elsewhere, a result local is declared up front and every branch assigns it (or throws); definite assignment is compiler-verified.

Two rules govern conditions:

- Conditions within the old isSimpleWhen budget (≤ one hoist-worth) still print **inline**, keeping flat `else if` chains for the common case — HEDIS's ubiquitous `(this.Foo(context)) ?? false` guards stay one-liners.
- A condition that needs its own statements may only evaluate after all earlier conditions tested false, so the chain continues **nested inside the `else` block**, where its statements print before a fresh `if`. This preserves the as-late-as-possible evaluation the old pipeline got from zero-argument when-functions — which are gone too, along with the whole isSimpleWhen function-wrapping machinery.

## Size and readability

- HEDIS 2025: **−13,285 lines (−4.2%)** across 382 libraries.
- dqm-content-qicore-2025 demo corpus alone: net −6,773 lines.
- The blank-line-wrapped function definitions and their invocations are gone from every corpus.

## Semantics

One accepted delta, flagged for review: a conditional's interior now evaluates at its linearization position (canonical left-to-right argument order) instead of deferred to the consumer statement. For CQL's pure expressions this is unobservable; it can shift error/`Message` timing in shapes like a conditional on the lazy side of a coalesce. The full runtime suites (which execute the new form), the HEDIS compile-all, and CI's integration tests (measure evaluation against real data) all pass.

## Version and regeneration

- `GeneratorToolVersion` 5.1.4.0 → **5.1.5.0** (patch: no shape the invoker binds to changes; stays inside `LibraryInstanceInvoker_5_0`'s range).
- All checked-in generated C# regenerated per the regenerate-then-revert-version-only convention from #1401's merge round: **77 files carry real changes, 72 reverted** (their only diff was the version header). The three pre-existing `cql`-path conversion failures (TJCOverallFHIR, VTEFHIR4, TestCodeSystemConceptRetrieve — front-end issues, unrelated) were routed around by regenerating from the checked-in ELM, as the precedent did.
- `docs/releases/vnext-release-notes.md` gains the generated-output breaking-change entry — covering #1401's output changes too, which had missed the release notes.

## Verification

- Full Debug + Release builds: 0 errors (all 149 regenerated files compile in both).
- CoreTests **825/0** — including the golden regeneration tests over the new form and the runtime tests that *execute* it.
- CqlToElmTests **2455/0**.
- HEDIS 2025: **382/382** compile to assemblies; representative diffs eyeballed (tail-position collapse, flat guard chains, nested statement-bearing conditions).


baseTwo:
@copilot resolve the merge conflicts in this pull request.

baseTwo (approved):

Issue #1403, opened 2026-07-24
Make TypeConverter.AddConversion public so hosts can extend/override FhirTypeConverter conversions


alexzautke:
**Is your feature request related to a problem? Please describe.**

`Hl7.Cql.Conversion.TypeConverter` is a public class and `FhirTypeConverter.Create(...)` / `FhirTypeConverter.Default` hand hosts a `TypeConverter` instance, but both `AddConversion` overloads are `internal`. A host application (e.g. a FHIR server exposing `$cql`) that needs a different or additional CQL→FHIR conversion — for example an adjusted `CqlInterval<decimal?>` → `Range` encoding, or support for a type the SDK doesn't convert yet — cannot register or override a conversion. The only options are bypassing the converter with duplicated host-side conversion logic, or forking the package.

**Describe the solution you'd like**

Make the conversion registration API public (or provide an equivalent public extension point), e.g.:

- `public void AddConversion<TFrom, TTo>(Func<TFrom, TTo> conversion)`
- optionally a `ReplaceConversion`/`TryAddConversion` variant with defined semantics when a conversion for the type pair already exists, so hosts can deliberately override SDK defaults.

**Describe alternatives you've considered**

- Host-side bypass: pattern-match the affected types before delegating to `FhirTypeConverter.Default`. Works, but duplicates SDK mapping logic and drifts as the SDK evolves.
- Reflection into the internal method — fragile and defeats the public API contract.

**Additional context**

Motivated by implementing the FHIR-56226 interval→Range mapping downstream (see #1402, VONK-9988): with a public `AddConversion`, hosts could ship such mapping changes without waiting for an SDK release, then drop their override once the SDK adopts the behavior.

alexzautke:
Closing this in favor of moving the conversion logic into the SDK itself rather than exposing a public `AddConversion` API.

The conversions Firely Server needs are all spec-driven (FHIR-56226, CQL IG) rather than host-specific, so they should be the SDK's default behavior for all consumers instead of per-host extensions. A public `AddConversion` would invite hosts to diverge from spec-mandated encodings while still leaving the maintenance burden on each host. Some of the required behavior (e.g. converting open interval endpoints to closed equivalents via `successor`/`predecessor` at full engine precision) can't be implemented correctly outside the engine anyway.

[...]

Issue #1402, opened 2026-07-24
Map Interval<Decimal>/Interval<Integer>/Interval<Long> to FHIR Range with quantity-precision extensions (FHIR-56226)


alexzautke:
**Is your feature request related to a problem? Please describe.**

Running the [cql-tests-runner](https://github.com/cqframework/cql-tests-runner) against a `$cql` endpoint built on this SDK surfaces failures for interval-typed results whose element type has no native FHIR datatype (e.g. `CqlIntervalOperatorsTest / DecimalIntervalExcept1to3`, expression `Interval[1.0, 10.0] except Interval[4.0, 10.0]`, expected `Interval [ 1.0, 3.99999999 ]`).

Three gaps in the current interval → FHIR conversion (verified against 2.9.2):

1. **No `quantity-precision` extension.** The `FhirTypeConverter` conversions for `CqlInterval<decimal?>` → `Range` and `CqlInterval<int?>` → `Range` emit bare unit-less UCUM bounds (`code: "1"`, `system: http://unitsofmeasure.org`) without the [quantity-precision](http://hl7.org/fhir/StructureDefinition/quantity-precision) extension proposed by [FHIR-56226](https://jira.hl7.org/browse/FHIR-56226) ("Add mapping for Intervals of Integer, Decimal, and Long" against the [CQL IG FHIR type mapping](https://hl7.org/fhir/uv/cql/conformance.html#fhir-type-mapping)).
2. **`lowClosed`/`highClosed` are ignored at serialization.** FHIR `Range` bounds are always inclusive, but an interval that reaches the converter with an open endpoint (e.g. a library directly returning `Interval[1.0, 1.4)`) has its open bound value emitted as-is instead of being converted to its closed equivalent via `successor`/`predecessor`.
3. **`Interval<Long>` is unsupported.** `CqlTypeToFhirTypeMapper.TypeEntryFor` returns `null` for an interval with a `Long` point type, and no `CqlInterval<long?>` → `Range` conversion is registered, so such results cannot be serialized at all.

**Describe the solution you'd like**

Per FHIR-56226:

- `Interval<Decimal>`, `Interval<Integer>`, and `Interval<Long>` all map to a FHIR `Range` whose bounds are unit-less Quantities (UCUM `code: "1"`).
[...]

PR #1408, opened 2026-07-25
feat: add host-side CQL↔FHIR conversions to FhirTypeConverter

### LINKED ISSUE #1406: Move remaining host-side CQL↔FHIR conversions from Firely Server into FhirTypeConverter

Sibling of #1402, follow-up to #1403.

Firely Server's `$cql` / `Library/$evaluate` implementation currently carries several CQL↔FHIR conversions locally, because `Hl7.Cql.Fhir.FhirTypeConverter` doesn't provide them and `TypeConverter.AddConversion` is `internal`. All of these are spec-driven (CQL IG [FHIR type mapping](https://hl7.org/fhir/uv/cql/conformance.html#fhir-type-mapping)) rather than host-specific, so they belong in the SDK where every consumer gets them and the encoding is defined once.

**Describe the solution you'd like**

Add the missing conversions to `FhirTypeConverter` (and the corresponding entries in `Hl7.Cql.Packaging.CqlTypeToFhirTypeMapper.TypeEntryFor` where a type mapping is missing too).

*CQL → FHIR*

1. **`CqlInterval<CqlTime>` → `Period`** — no `Interval<Time>` conversion exists in the SDK. Firely Server anchors the time on the minimum FHIR date (`0001-01-01T{time}`); whatever encoding is chosen should be defined once, SDK-wide.
2. **`CqlCode` → `code` vs `Coding` disambiguation** — when a `CqlCode` has only `code` populated (no `system`, `version` or `display`), emit a bare FHIR `code`; otherwise a `Coding`. The SDK currently only registers `CqlCode` → `Coding`.
3. **`long` → FHIR** — currently unmapped (the type mapper returns `null` for `Long`), so Firely Server falls back to `FhirString`. Overlaps with the `Interval<Long>` gap in #1402.

*FHIR → CQL*

4. **`code`/`Coding` → `CqlCode`** — no inbound conversions exist (mirror of item 2).
5. **Type-hinted `Period` conversion** — `Period` maps ambiguously to `Interval<Date>` or `Interval<DateTime>`. Hosts resolve this from the `cqf-cqlType` extension on the `Library` parameter (per the CQL IG), but then have to pattern-match around the converter. A `Convert` overload accepting the target CQL type (name) would let hosts pass the hint and delegate the conversion itself.

**Describe alternatives you've considered**

- Keeping the conversions host-side, as today. This duplicates type-mapping logic that belongs to the SDK's CQL-on-FHIR layer, drifts as the SDK evolves, and leaves every other consumer without the mappings.
- Exposing `TypeConverter.AddConversion` publicly so hosts can register these themselves — see #1403, closed as not planned: a public extension point would invite hosts to diverge from spec-mandated encodings while still leaving the maintenance burden on each host.

**Additional context**

Together with #1402 (quantity-precision extensions, `lowClosed`/`highClosed` handling, `Interval<Long>`), this would let hosts route all value conversion through `FhirTypeConverter.Default` and delete their local converters.

alexzautke:
Moves the CQL↔FHIR conversions that hosts (e.g. Firely Server's `$cql` / `Library/$evaluate`) previously implemented locally into the SDK, so every consumer gets the same spec-driven encoding.

Refs #1406. Sibling of #1402, follow-up to #1403 (VONK-9988).

## Primary Work

New conversions in `Hl7.Cql.Fhir.FhirTypeConverter`:

- `CqlInterval<CqlTime>` → `Period`, anchoring the times on the minimum FHIR date (`0001-01-01`). The `dateTime` values are composed from the time's ISO 8601 string rather than built via `DateTimeOffset`, which preserves the time's original precision, adds no timezone offset when the time has none, and sidesteps the fact that the minimum date combined with a positive UTC offset (e.g. `0001-01-01T00:30:00+02:00`) has no `DateTimeOffset` representation even though it is a valid FHIR `dateTime`. A zero offset (ISO 8601 `Z`) is rewritten to the explicit `+00:00` form.
- `CqlCode` → `Code`, alongside the existing `CqlCode` → `Coding`. The new `TypeConverter.ConvertCqlCodeToFhir(CqlCode?)` extension picks between the two: a bare `code` when only the code element is populated, a `Coding` otherwise, and `null` when no element is populated at all (avoiding a `Code` with a null value).
- `long` → `FhirString`, matching the CQL `Long` → FHIR `string` mapping in `CqlTypeToFhirTypeMapper` on R4 (R5 hosts can move to `integer64` once that mapping is enabled).
- `Code` → `CqlCode` and `Coding` → `CqlCode` for the inbound direction.
- `TypeConverter.ConvertPeriodToCqlInterval(Period?, string?)` extension, which resolves the ambiguous `Period` mapping to `CqlInterval<CqlDate>` or `CqlInterval<CqlDateTime>` from a CQL point type name (typically read from the `cqf-cqlType` extension on a `Library` parameter).

Covered by the new `FhirTypeConverterHostConversionTests`, and recorded in `docs/releases/vnext-release-notes.md`.

---

## Auxiliary Work

- **ISO 8601 timezone offsets are now signed as a whole, not only on their hour component** (surfaced while testing the `Interval<Time>` → `Period` conversion): `TimeIso8601` read an offset of `-05:30` as `-04:30` (`DateTimeIso8601` already handled this), and both types rendered a zero-hour offset without its sign (`00:30` instead of `+00:30`) and could not represent `-00:30` at all. `OffsetMinute` is now negative for negative offsets on both types, and the strict-validation checks and exception messages accept the signed range `[-59,59]`. The new `Iso8601OffsetTests` cover construction, parsing (including non-zero-hour negative offsets such as `-05:30`), and rendering on both types.


Cql/Cql.Firely/FhirTypeConverter.cs
```diff
@@ -385,6 +410,59 @@ internal static TypeConverter ConvertCqlPrimitivesToFhir(this TypeConverter conv
...
+        /// to disambiguate the otherwise ambiguous Period mapping.
+        /// </summary>
+        /// <param name="converter">the type converter</param>
+        /// <param name="period">the period to convert, or <see langword="null"/></param>
+        /// <param name="cqlPointTypeName">the CQL point type name; "Date" yields an Interval&lt;Date&gt;, anything else an Interval&lt;DateTime&gt;</param>
+        /// <returns>a <see cref="CqlInterval{CqlDate}"/> or <see cref="CqlInterval{CqlDateTime}"/>, or <see langword="null"/> when <paramref name="period"/> is <see langword="null"/></returns>
+        public static object? ConvertPeriodToCqlInterval(this TypeConverter converter, M.Period? period, string? cqlPointTypeName) =>
+            cqlPointTypeName switch
+            {
+                "Date" => converter.Convert<CqlInterval<CqlDate>>(period),
```
alexzautke: @copilot Can we reuse an existing constant for this?

Cql/Iso8601/TimeIso8601.cs
```diff
@@ -180,7 +180,7 @@ internal TimeIso8601(string @string, int hour, int? minute, int? second, int? ms
                         throw new ArgumentException("Offset hours must between [-14,14]", nameof(osHour));
                     if (osMinute.HasValue)
                     {
-                        if (osMinute.Value < 0 || osMinute.Value > 59)
+                        if (Math.Abs(osMinute.Value) > 59)
                             throw new ArgumentException("Offset minutes must between [0,59]", nameof(osMinute));
```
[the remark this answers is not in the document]
alexzautke: Fixed in 3290b27 — the exception messages in `TimeIso8601` and `DateTimeIso8601` now say `[-59,59]`, matching the signed-minute range the `Math.Abs` check actually allows.

---

Cql/Cql.Firely/FhirTypeConverter.cs
```diff
@@ -385,6 +410,59 @@ internal static TypeConverter ConvertCqlPrimitivesToFhir(this TypeConverter conv
...
+                ? "." + milliseconds.ToString("D3", CultureInfo.InvariantCulture).TrimEnd('0')
+                : string.Empty;
+            var offset = t.Offset;
+            var absoluteOffset = offset.Duration();
+            var value = string.Format(CultureInfo.InvariantCulture,
+                "0001-01-01T{0:D2}:{1:D2}:{2:D2}{3}{4}{5:D2}:{6:D2}",
+                t.Hour, t.Minute ?? 0, t.Second ?? 0, fraction,
+                offset < TimeSpan.Zero ? '-' : '+', absoluteOffset.Hours, absoluteOffset.Minutes);
+            return new M.FhirDateTime(value);
+        }
```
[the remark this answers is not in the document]
alexzautke: Fixed in 3290b27 — `CqlTimeToFhirDateTime` now composes the value from the time's ISO 8601 string, so the original precision is preserved and no offset is appended when the time has none; only the `Z` UTC form is rewritten to `+00:00`. Added `ConvertCqlIntervalOfTime_Period_PreservesPrecisionAndAbsentOffset` to lock this in.

---

Cql/Cql.Firely/FhirTypeConverter.cs
```diff
@@ -385,6 +410,59 @@ internal static TypeConverter ConvertCqlPrimitivesToFhir(this TypeConverter conv
...
+        /// <param name="converter">the type converter</param>
+        /// <param name="code">the CQL code to convert, or <see langword="null"/></param>
+        /// <returns>a <see cref="M.Code"/> or <see cref="M.Coding"/>, or <see langword="null"/> when <paramref name="code"/> is <see langword="null"/></returns>
+        public static M.DataType? ConvertCqlCodeToFhir(this TypeConverter converter, CqlCode? code) =>
+            code switch
+            {
+                null => null,
+                { system: null, version: null, display: null } => converter.Convert<M.Code>(code),
+                _ => converter.Convert<M.Coding>(code)
+            };
```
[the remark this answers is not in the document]
alexzautke: Fixed in 3290b27 — `ConvertCqlCodeToFhir` now returns `null` when the code value is null/empty and no other element is populated, instead of producing a `Code` with `Value == null`. Covered by `ConvertCqlCodeToFhir_Null_WhenNothingPopulated`.

---

Cql/CoreTests/Iso8601OffsetTests.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    [TestMethod]
+    public void DateTimeIso8601_PositiveOffsetWithoutHours_RendersSign()
+    {
+        var dateTime = new DateTimeIso8601(2024, 1, 1, 10, 30, 0, null, 0, 30);
+
+        Assert.AreEqual(new TimeSpan(0, 30, 0), dateTime.DateTimeOffset.Offset);
+        Assert.AreEqual("2024-01-01T10:30:00+00:30", dateTime.ToString());
+    }
+}
```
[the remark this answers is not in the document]
alexzautke: Added in 3290b27 — `DateTimeIso8601_ParsesNegativeOffsetWithMinutes` parses `2024-01-01T10:30:00-05:30` and asserts the offset, both signed components, and the round-tripped string.

---

Cql/Cql.Firely/FhirTypeConverter.cs
```diff
@@ -144,6 +144,8 @@ internal static TypeConverter ConvertFhirToCqlPrimitives(
             add((M.Integer c) => new M.UnsignedInt(c.Value));
             add((M.Integer c) => new M.PositiveInt(c.Value));
             add((M.Code c) => c.Value);
+            add((M.Code c) => new CqlCode { code = c.Value });
+            add((M.Coding c) => new CqlCode { code = c.Code, system = c.System, version = c.Version, display = c.Display });
```
baseTwo: **Blocker (correctness):** This `M.Coding -> CqlCode` conversion duplicates one already merged into `develop` by PR #1354 (`add((M.Coding c) => new CqlCode(c.Code, c.System, c.Version, c.Display));`, added elsewhere in this same file). `TypeConverter.AddConversion` throws `ArgumentException("Conversion from ... is already defined.")` on a duplicate (from, to) pair, and `FhirTypeConverter.CreateImpl` runs this registration unconditionally on every `Default`/`Create(...)` call.

I verified this isn't theoretical: merging `origin/develop` into this branch in a scratch worktree auto-merges `FhirTypeConverter.cs` with **no conflict markers** (the only textual conflict is in `docs/releases/vnext-release-notes.md`), so a developer resolving that one visible conflict would believe the branch is clean. But building and running `FhirTypeConverter.Default` against the merged result throws immediately: `System.ArgumentException: Conversion from Hl7.Fhir.Model.Coding to Hl7.Cql.Primitives.CqlCode is already defined.` — breaking every FhirTypeConverter-dependent test and all runtime CQL↔FHIR conversion.

Please drop this PR's `Coding -> CqlCode` registration (line 148) in favor of the one already on `develop` from #1354 as part of rebasing this branch.

Cql/CoreTests/Iso8601OffsetTests.cs
```diff
@@ -0,0 +1,114 @@
...
+        Assert.IsTrue(DateTimeIso8601.TryParse("2024-01-01T10:30:00-05:30", out var dateTime));
+
+        Assert.AreEqual(new TimeSpan(-5, -30, 0), dateTime!.DateTimeOffset.Offset);
+        Assert.AreEqual(-5, dateTime.OffsetHour);
+        Assert.AreEqual(-30, dateTime.OffsetMinute);
+        Assert.AreEqual("2024-01-01T10:30:00-05:30", dateTime.ToString());
+    }
+
+    [TestMethod]
+    public void DateTimeIso8601_ParsesNegativeOffsetWithoutHours()
```
baseTwo, line 99: **Blocker (required test-coverage fix):** This is the test for the exact bug these release notes describe ("a negative, zero-hour offset like `-00:30` could not be represented"), but it only asserts `dateTime.DateTimeOffset.Offset`, never `dateTime.ToString()` — so it never actually exercises the `Format()` fix.

Worse, `DateTimeIso8601.TryParse` reconstructs the result via the internal string-preserving constructor, which stores the original parsed string verbatim as `ToString()`'s output rather than recomputing it through `Format()` — so even adding a `ToString()` assertion *to this specific test* would pass trivially without ever touching the fixed code path. I confirmed this empirically: reverting only the `Format()` fix (keeping the parsing fix intact) and running the whole suite leaves 9/10 tests passing, and this one among them — it catches nothing.

Please add a test that goes through the raw constructor for the zero-hour/negative-minute case, e.g.:
```csharp
var dateTime = new DateTimeIso8601(2024, 1, 1, 10, 30, 0, null, 0, -30);
Assert.AreEqual("2024-01-01T10:30:00-00:30", dateTime.ToString());
```
(mirroring the existing `DateTimeIso8601_PositiveOffsetWithoutHours_RendersSign` test just below, but with a negative minute).

Cql/CoreTests/FhirTypeConverterHostConversionTests.cs
```diff
@@ -0,0 +1,195 @@
...
+
+            var converted = FhirTypeConverter.ConvertCqlCodeToFhir(code);
+
+            var fhirCode = converted as Code;
+            Assert.IsNotNull(fhirCode);
+            Assert.AreEqual("123", fhirCode.Value);
+        }
+
+        [TestMethod]
+        public void ConvertCqlCodeToFhir_Null_WhenNothingPopulated()
```
baseTwo, line 118: **Required test-coverage fix (major):** `ConvertCqlCodeToFhir`'s null-guard (`FhirTypeConverter.cs`) special-cases `{ system: null, version: null, display: null, code: null or "" }` to return null. This test only constructs `new CqlCode()` (code defaults to null), so it exercises the `code: null` half of that pattern but never the `code: ""` half. If the `or ""` clause were ever dropped/narrowed in a future refactor, a `CqlCode` with an explicit empty-string code would fall through to producing a `Code` with an effectively-empty value — the exact defect this branch exists to prevent — with nothing catching it. Please add a case with `new CqlCode { code = "" }` asserting the same null result.

Cql/CoreTests/FhirTypeConverterHostConversionTests.cs
```diff
@@ -0,0 +1,195 @@
...
+        {
+            var code = new CqlCode();
+
+            var converted = FhirTypeConverter.ConvertCqlCodeToFhir(code);
+
+            Assert.IsNull(converted);
+        }
+
+        [TestMethod]
+        public void ConvertCqlCodeToFhir_Coding_WhenSystemPopulated()
```
baseTwo, line 138: **Required test-coverage fix (minor):** Only the "all three of system/version/display populated" permutation of the Coding fallback branch is tested (despite the test name implying system-only). The guard is `{ system: null, version: null, display: null } => Code` / `_ => Coding`, so a version-only or display-only `CqlCode` also falls into the `Coding` branch, but nothing tests those permutations — a future narrowing of the guard to check only `system` would pass this suite while silently discarding version/display-only metadata. Consider adding version-only and display-only cases.

Cql/Iso8601/TimeIso8601.cs
```diff
@@ -227,6 +227,11 @@ internal TimeIso8601(string @string, int hour, int? minute, int? second, int? ms
                 seconds: second ?? 0,
                 milliseconds: ms ?? 0);
             OffsetHour = osHour;
+            // The sign of an offset applies to the whole offset, not just its hour component:
+            // -05:30 is -(5h + 30m), not -5h + 30m. Store the minutes signed so that composing
+            // the two components yields the right TimeSpan.
+            if (osHour < 0 && osMinute > 0)
```
baseTwo: **Required test-coverage fix (minor):** This only re-signs the minute for the negative-hour/positive-minute direction; the opposite direction (positive hour, negative minute, e.g. `osHour=5, osMinute=-30`) is left untouched. I traced this through: `new TimeIso8601(10, 30, 0, null, 5, -30)` ends up with `.Offset` == `TimeSpan(5,-30,0)` == **+04:30** (positive), while `.ToString()` renders **"-05:30"** (negative) — an internally inconsistent object. This exact asymmetry was already flagged by Copilot's own suppressed low-confidence review comment on this line and was left unaddressed by the follow-up commit. No test (positive or negative) currently exercises a mixed-sign raw-constructor pair. Please either fix the normalization for this direction too, or add a test that pins down the current (arguably still-buggy) behavior so a future change doesn't silently alter it further.

Cql/Cql.Firely/FhirTypeConverter.cs
```diff
@@ -385,6 +410,56 @@ internal static TypeConverter ConvertCqlPrimitivesToFhir(this TypeConverter conv
...
+
+        /// <summary>
+        /// Converts a <see cref="CqlCode"/> to the most appropriate FHIR type: a bare <see cref="M.Code"/>
+        /// when only the code element is populated, otherwise a <see cref="M.Coding"/>.
+        /// </summary>
+        /// <param name="converter">the type converter</param>
+        /// <param name="code">the CQL code to convert, or <see langword="null"/></param>
+        /// <returns>a <see cref="M.Code"/> or <see cref="M.Coding"/>, or <see langword="null"/> when <paramref name="code"/> is
+        /// <see langword="null"/> or has no elements populated at all</returns>
+        public static M.DataType? ConvertCqlCodeToFhir(this TypeConverter converter, CqlCode? code) =>
```
baseTwo, line 415: **Required fix (major, build-breaking in Release):** This new public method (and `ConvertPeriodToCqlInterval` below it) has no corresponding entry in `Cql/Cql.Firely/PublicAPI.Unshipped.txt`, even though every other public member of this class is tracked there. I reproduced the consequence directly: `dotnet build Cql/Cql.Firely/Cql.Firely.csproj -c Release` fails with two RS0016 errors ("Symbol ... is not part of the declared public API") for both net8.0 and net10.0, since `Microsoft.CodeAnalysis.PublicApiAnalyzers` + `TreatWarningsAsErrors` are wired for Release builds. This PR's CI currently shows green only because of an unrelated pipeline quirk (the restore step defaults to Debug, so the Release-only analyzer package never restores) — that's pre-existing and not something to fix here, but it means CI won't catch this. Please add the two missing entries to `PublicAPI.Unshipped.txt`.

Cql/Iso8601/DateTimeIso8601.cs
```diff
@@ -1,8 +1,8 @@
 ﻿#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
-/*
+/*
```
baseTwo, line 2: **Nit:** This touches two lines inside the copyright-header block (here and the blank `*` line below) — turns out to be a CRLF-normalization artifact (the pre-existing file had LF-only line endings on just these two lines, now normalized to CRLF to match the rest of the file), not a content change; the attribution text itself is untouched. Still, this repo's convention is to leave an existing file's header alone entirely. Same applies to `Cql/Iso8601/TimeIso8601.cs`'s header. Not blocking, but worth reverting those two lines' line-endings to keep the diff minimal, if easy to do.

baseTwo (changes requested):
@copilot Found a genuine blocker below (a duplicate type-conversion registration that will throw at runtime once this branch is rebased onto current `develop` — verified by actually merging, compiling, and running it), plus a required test-coverage gap on the exact bug this PR's release notes claim to fix, a Release-build-breaking omission (also empirically reproduced), and a couple of narrower coverage gaps. Per this repo's standing convention, test-coverage gaps are required fixes, not nice-to-haves — please address all of these.

baseTwo:
@copilot Two things to handle together when merging `origin/develop` in (this branch conflicts with it — one real conflict in `docs/releases/vnext-release-notes.md`, plus a silent, non-conflicting duplicate that also needs fixing).

## 1. Release notes: use the new fragment-file convention, and migrate #1409's entry too

This repo now splits release notes into one fragment file per PR under `docs/releases/vnext/` (see [`docs/releases/vnext/README.md`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/releases/vnext/README.md)) instead of everyone editing the shared `vnext-release-notes.md` and hitting merge conflicts — which is exactly the conflict this PR just hit. Please:

1. **Create `docs/releases/vnext/1406-host-side-conversions.md`** (issue #1406 is this PR's tracking issue) containing exactly this PR's own two entries (the ones currently in this branch's copy of `vnext-release-notes.md`):

```md
## Features

- `Hl7.Cql.Fhir.FhirTypeConverter` now covers conversions that CQL-on-FHIR hosts previously had to
  implement themselves:
  - `CqlInterval<CqlTime>` → `Period`, anchoring the times on the minimum FHIR date (`0001-01-01`).
  - `CqlCode` → `Code`, alongside the existing `CqlCode` → `Coding`. The new
    `TypeConverter.ConvertCqlCodeToFhir(CqlCode?)` extension picks between the two: a bare `code`
    when only the code element is populated, a `Coding` otherwise.
  - `long` → `FhirString`, matching the CQL `Long` → FHIR `string` mapping in
    `CqlTypeToFhirTypeMapper` on R4.
  - `Code` → `CqlCode` and `Coding` → `CqlCode` for the inbound direction.
  - `TypeConverter.ConvertPeriodToCqlInterval(Period?, string?)` extension, which resolves the
    ambiguous `Period` mapping to `CqlInterval<CqlDate>` or `CqlInterval<CqlDateTime>` from a CQL
    point type name (typically read from the `cqf-cqlType` extension on a `Library` parameter).

## Fixes

- ISO 8601 timezone offsets are now signed as a whole rather than only on their hour component.
  `TimeIso8601` read an offset of `-05:30` as `-04:30` (`DateTimeIso8601` already handled this), and
  both types rendered an offset with a zero hour component without its sign (`00:30` instead of
  `+00:30`) and could not represent `-00:30` at all. Note that `TimeIso8601.OffsetMinute` is now
  negative for negative offsets, matching `DateTimeIso8601.OffsetMinute`.
```

2. **Also create `docs/releases/vnext/1402-interval-range-mapping.md`** for PR #1409's entries (#1409 already merged to `develop` directly into the shared file, before this migration — since you're already touching this file to resolve the conflict, please pull its entries out into a fragment too rather than leaving them as the one straggler in the shared file):

```md
## Breaking Changes

- FHIR serialization of `Interval<Integer>` and `Interval<Decimal>` results changed, per [FHIR-56226](https://jira.hl7.org/browse/FHIR-56226): every `Range` bound now carries the [`quantity-precision`](http://hl7.org/fhir/StructureDefinition/quantity-precision) extension (`valueInteger` = number of digits after the decimal point), and because FHIR `Range` bounds are always inclusive, an open endpoint is now emitted as its closed equivalent instead of as-is (successor/predecessor at the engine's minimum precision, so `Interval[1.0, 4.0)` becomes `[1.0, 3.99999999]`). Consumers comparing exact `Range` output need to account for both (#1402).

## Features

- `Interval<Long>` results now map to a FHIR `Range` of unit-less Quantities (UCUM `1`), per [FHIR-56226](https://jira.hl7.org/browse/FHIR-56226). Previously `CqlTypeToFhirTypeMapper` had no mapping for an interval with a `Long` point type and no `CqlInterval<long?>` → `Range` conversion was registered, so such results could not be serialized at all (#1402).
```

3. **Resolve the conflict on `docs/releases/vnext-release-notes.md` by taking `origin/develop`'s version verbatim (`git checkout origin/develop -- docs/releases/vnext-release-notes.md`), then delete the two `(#1402)` entries** (the Breaking Changes bullet and the Features bullet) now that they live in the fragment file above. Don't add this PR's own entries to the shared file at all — they belong only in the new `1406-host-side-conversions.md` fragment.

## 2. Revert the `hasCodingToCqlCodeConversion()` guard — it doesn't actually fix the crash, and it's the wrong kind of fix anyway

I checked this empirically: after merging `develop` in, `FhirTypeConverter.cs` ends up with **two** `M.Coding -> CqlCode` registrations — this one (guarded, at the earlier line) and PR #1354's pre-existing one (unguarded, later in the same method). Since registrations run top-to-bottom in a single method body, your guarded line executes *first*, finds nothing registered yet (correctly), and registers anyway — then PR #1354's *unguarded* line executes second and throws `ArgumentException` on the now-duplicate registration. The guard doesn't prevent the crash, it just moves which line throws.

More importantly, even a version of this that worked would be the wrong shape of fix — this repo's conventions are against adding runtime validation/guards for a scenario that "can't happen" in the final, correctly-merged codebase (see CLAUDE.md: "Don't add error handling, fallbacks, or validation for scenarios that can't happen"). The actual fix is simpler: once merged with `develop`, there will only be one `Coding -> CqlCode` registration left in the file (PR #1354's), so this PR's own duplicate line and the entire `hasCodingToCqlCodeConversion()` helper should just be deleted — not guarded. Please:
- Remove this PR's own `add((M.Coding c) => new CqlCode { ... })` registration (and the `if (!hasCodingToCqlCodeConversion())` wrapper) entirely.
- Remove the `hasCodingToCqlCodeConversion()` local function entirely.
- Confirm only one `Coding -> CqlCode` registration remains in the merged file (PR #1354's).

Please run the full `CoreTests` suite before pushing and confirm it passes, given the merge-conflict history on this queue.

baseTwo (approved):
Independently verified everything empirically (built and ran the full CoreTests suite against this exact commit, not just read the diff):

- **Release notes**: both fragment files (`docs/releases/vnext/1406-host-side-conversions.md` and `1402-interval-range-mapping.md`) are byte-for-byte correct, and the shared `vnext-release-notes.md` now matches develop's tip minus exactly the two #1402 entries — no other content lost.
- **Duplicate registration**: exactly one `M.Coding -> CqlCode` registration remains (develop's original), the `hasCodingToCqlCodeConversion()` guard is fully removed, and `FhirTypeConverter.Default` construction no longer throws — confirmed via an actual Release build plus a full `CoreTests` run (927 passed, 0 failed, 9 pre-existing skips) in an isolated worktree.

All 7 original findings are resolved. Approving.

PR #1409, opened 2026-07-25
Map Interval<Integer|Decimal|Long> to FHIR Range per FHIR-56226

### LINKED ISSUE #1402: Map Interval<Decimal>/Interval<Integer>/Interval<Long> to FHIR Range with quantity-precision extensions (FHIR-56226)

alexzautke:
Interval-typed results whose element type has no native FHIR datatype were serialized without the precision information needed to round-trip them, and intervals of `Long` could not be serialized at all. This aligns the mapping with [FHIR-56226](https://jira.hl7.org/browse/FHIR-56226) and the CQL IG's FHIR type mapping.

Fixes #1402

## Primary Work

**`FhirTypeConverter` — `Interval<Integer|Decimal|Long>` → FHIR `Range`**

- Every `Range` bound now carries the [`quantity-precision`](http://hl7.org/fhir/StructureDefinition/quantity-precision) extension (`valueInteger` = number of digits after the decimal point), so precision no longer depends on a serializer preserving trailing zeros.
- FHIR `Range` bounds are always inclusive, so an open endpoint is emitted as its **closed equivalent** — successor of an open low bound, predecessor of an open high bound. The step is the engine's minimum precision value for the point type (`10^-8` for Decimal, `1` for Integer/Long), matching `ToClosed()`, so `Interval[1.0, 4.0)` serializes as `[1.0, 3.99999999]` as the official CQL test suite expects.
- Registered the missing `CqlInterval<long?>` → `Range` conversion.
- The three near-identical conversion lambdas collapsed into one `NumericIntervalToRange` + `UnitlessQuantity` pair.

**`CqlTypeToFhirTypeMapper`** — an interval with a `Long` point type now maps to FHIR `Range` instead of returning `null`.

**`CqlOperators`** — extracted the Decimal minimum precision value into `MinDecimalPrecisionValue`, now shared between `Successor`/`Predecessor` and the Range conversion instead of being repeated as a literal.

The serialization change is a breaking one for consumers comparing exact `Range` output, and is recorded in [`vnext-release-notes.md`](docs/releases/vnext-release-notes.md) along with the new `Interval<Long>` support.

No `GeneratorToolVersion` bump: nothing here changes the C# emitted for CQL libraries, the `ICqlOperators` surface, or the calls `CqlOperatorsBinder` emits.

### Review follow-up: how decimal scale is computed

Review flagged `value.Scale` as not compiling on `net8.0`. That turned out to be incorrect — `Decimal.Scale` has existed since .NET 7 and is available on both `net8.0` and `net10.0` — so the property stays.

The check did surface something real, though: two pre-existing sites computed the same thing by hand, both via `BitConverter.GetBytes(decimal.GetBits(value)[3])[2]`, which only lands on the scale byte on a little-endian machine:

- `CqlComparers.DecimalCqlComparer.GetPrecision` (equivalence comparison)
- a local `getDecimalPrecision` function in `CqlOperators.ListOperators` (`expand`'s `per`)

Both now read `decimal.Scale` directly and the two helpers are gone, so the duplication and the endianness assumption go with them. Behavior is unchanged on any little-endian platform, so no test expectations moved.

### Tests

- `FhirTypeConverterTests` — closed bounds keep their precision, open bounds convert to closed (Decimal, Integer and Long), null endpoints omit the bound entirely, plus the `DecimalIntervalExcept1to3` case from the CQL test suite.
- `CqlTypeToFhirTypeMapperTest` / `Packaging.CqlTypeToFhirTypeMapperTests` — `Interval<Long>` maps to `Range` with a `Long` element type.

---

## Auxiliary Work

- Vendored the CQL IG conformance page (FHIR type mapping) as [`spec/fhir/condensed/fhir-uv-cql-conformance.md`](spec/fhir/condensed/fhir-uv-cql-conformance.md), with its row added to the [`spec/fhir/README.md`](spec/fhir/README.md) ground-truth table.


Cql/Cql.Firely/FhirTypeConverter.cs
```diff
@@ -385,6 +356,45 @@ internal static TypeConverter ConvertCqlPrimitivesToFhir(this TypeConverter conv
...
+        /// <summary>
+        /// Creates the unit-less (UCUM <c>1</c>) Quantity used for the bounds of a Range converted from an interval
+        /// of Integer, Decimal or Long. The <see cref="QuantityPrecisionExtensionUrl"/> extension is always added, so
+        /// that the number of digits after the decimal point does not depend on a serializer preserving trailing zeros.
+        /// </summary>
+        private static M.Quantity UnitlessQuantity(decimal value)
+        {
+            var quantity = new M.Quantity(value, "1");
+            quantity.Extension.Add(new M.Extension(QuantityPrecisionExtensionUrl, new M.Integer(value.Scale)));
+            return quantity;
```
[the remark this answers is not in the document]
alexzautke: Applied in a572319 — the scale is now computed via `decimal.GetBits`, with a comment explaining the bit extraction.

One correction on the premise though: `decimal.Scale` *is* available on `net8.0`. It was added in .NET 7 ([API docs](https://learn.microsoft.com/en-us/dotnet/api/system.decimal.scale) list net7.0 through net11.0), so `value.Scale` would have compiled fine for both targets — there was no build break to fix.

The change is still worth keeping for consistency: `GetBits`-based scale extraction is the idiom already used elsewhere in this repo (`CqlComparers.DecimalCqlComparer.GetPrecision` and `CqlOperators.ListOperators`), so all three sites now read the same way.

---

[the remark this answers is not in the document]
alexzautke: Follow-up: reverted to `decimal.Scale` in 2e31349, so the `GetBits` form I mentioned above is no longer used anywhere.

Since `Scale` is available on both targets, the hand-rolled extraction had nothing to offer. The two pre-existing runtime sites (`CqlComparers.DecimalCqlComparer` and `CqlOperators.ListOperators`) were also reading the scale via `BitConverter.GetBytes(decimal.GetBits(value)[3])[2]`, which only picks the right byte on a little-endian machine — those now use `.Scale` too, so the duplication and the endianness assumption are both gone.

---

Cql/Cql.Firely/FhirTypeConverter.cs
```diff
@@ -385,6 +356,47 @@ internal static TypeConverter ConvertCqlPrimitivesToFhir(this TypeConverter conv
...
+        /// </summary>
+        internal const string QuantityPrecisionExtensionUrl = "http://hl7.org/fhir/StructureDefinition/quantity-precision";
+
+        /// <summary>
+        /// Converts an interval of Integer, Decimal or Long to a FHIR Range of unit-less Quantities (FHIR-56226).
+        /// FHIR Range bounds are always inclusive, so an open endpoint is emitted as its closed equivalent, i.e.
+        /// the successor of an open low bound and the predecessor of an open high bound, stepping by <paramref name="step"/> -
+        /// the same minimum precision value the engine's ToClosed() applies for the interval's point type.
+        /// </summary>
+        private static M.Range NumericIntervalToRange(decimal? low, decimal? high, bool? lowClosed, bool? highClosed, decimal step)
```
baseTwo, line 385: **Correctness (minor, boundary-only):** `NumericIntervalToRange`'s open→closed step for `Interval<Integer>`/`Interval<Long>` is computed in `decimal` arithmetic (`l + step` / `h - step`, `step = 1m`), but the engine's own `ToClosed()` is built on `Successor(int?)`/`Predecessor(int?)`/`Successor(long?)`/`Predecessor(long?)` (`CqlOperators.ArithmeticOperators.cs`), which are plain unchecked `argument +/- 1` — no `checked` context anywhere in this repo (confirmed: no `CheckForOverflowUnderflow` in any csproj/Directory.Build.props).

At the extreme ends of the range this diverges: for `CqlInterval<int?>(int.MaxValue, high, lowClosed:false, highClosed:true)`, the engine's `ToClosed()`/`Successor(int?)` wraps to `int.MinValue` (-2147483648), but this converter computes `(decimal)int.MaxValue + 1m = 2147483648m` — no wraparound, since `decimal` has far more range than `int`/`long`. The emitted `Range.Low` therefore contradicts what the engine's own `ToClosed()` would produce for the same interval, which is the PR's own stated design goal (matching `ToClosed()`). Same issue applies to `Interval<Long>` at `long.MaxValue`/`long.MinValue`.

This is a narrow edge case (only triggered at `int.MaxValue`/`int.MinValue`/`long.MaxValue`/`long.MinValue`), but worth either matching the engine's wraparound behavior explicitly, or documenting the intentional divergence if the wraparound behavior is considered a bug in the engine that this converter shouldn't reproduce.

Cql/CoreTests/FhirTypeConverterTests.cs
```diff
@@ -657,8 +657,46 @@ public void ConvertCqlDecimalInterval_Range()
...
+            Assert.IsNotNull(converted);
+
+            Assert.AreEqual(1.0m, converted.Low.Value);
+            Assert.AreEqual(1, QuantityPrecisionOf(converted.Low));
+            Assert.AreEqual(10.000m, converted.High.Value);
+            Assert.AreEqual(3, QuantityPrecisionOf(converted.High));
+        }
+
+        [TestMethod]
+        public void ConvertCqlDecimalInterval_Range_ConvertsOpenBoundsToClosed()
```
baseTwo, line 685: **Test-coverage gap (required fix — blocker):** No test exercises an open (exclusive) low bound for the `Interval<Decimal>` → `Range` conversion. The decimal successor branch in `NumericIntervalToRange` (`l + step`, `step = CqlOperators.MinDecimalPrecisionValue = 0.00000001m`) is never executed by any test in this file:
- `ConvertCqlDecimalInterval_Range_ConvertsOpenBoundsToClosed` uses `(1.0m, 4.0m, true, false)` — low is closed, only the high-open branch runs.
- `ConvertCqlInterval_Range_OmitsBoundsForNullEndpoints` has a null (not open) low.
- `DecimalIntervalExcept1to3`'s result interval is already fully closed.

Only the Integer/Long 'both open' tests reach a successor branch, and those use `step=1m`, not `MinDecimalPrecisionValue`. Please add a test with a non-null, open (`lowClosed:false`) decimal low bound (e.g. `Interval(1.05, 4.0]`) asserting the successor value and its precision extension — a sign error, wrong step constant, or broken precision calc specific to this path would currently pass the whole suite undetected.

Cql/CoreTests/FhirTypeConverterTests.cs
```diff
@@ -672,9 +710,97 @@ public void ConvertCqlIntInterval_Range()
...
+            var converted = FhirTypeConverter.Convert<Hl7.Fhir.Model.Range>(longInterval);
+
+            Assert.IsNotNull(converted);
+
+            Assert.AreEqual(2, converted.Low.Value);
+            Assert.AreEqual(9, converted.High.Value);
+        }
+
+        [TestMethod]
+        public void ConvertCqlInterval_Range_OmitsBoundsForNullEndpoints()
```
baseTwo, line 840: **Test-coverage gap (required fix — major):** No test covers a missing/null HIGH bound of the Range conversion for any of the three numeric types (Decimal, Integer, Long) — only a missing LOW bound (Decimal only, via `ConvertCqlInterval_Range_OmitsBoundsForNullEndpoints`) is tested. `NumericIntervalToRange` has two independent conditionals (`if (low is {} l)` / `if (high is {} h)`); a bug isolated to the high-bound branch (e.g. a copy-paste/variable-swap mistake) would ship undetected. Please add a null-high-bound test for at least one type (ideally all three, alongside the Integer/Long gap noted separately below).

Cql/CoreTests/FhirTypeConverterTests.cs
```diff
@@ -672,9 +710,97 @@ public void ConvertCqlIntInterval_Range()
...
+            Assert.IsNotNull(converted);
+
+            Assert.AreEqual(2, converted.Low.Value);
+            Assert.AreEqual(0, QuantityPrecisionOf(converted.Low));
+            Assert.AreEqual(9, converted.High.Value);
+            Assert.AreEqual(0, QuantityPrecisionOf(converted.High));
+        }
+
+        [TestMethod]
+        public void ConvertCqlLongInterval_Range()
```
baseTwo, line 780: **Test-coverage gap (required fix — major):** `Interval<Integer>` and `Interval<Long>` have zero coverage for a null/missing endpoint (unbounded interval) — all four tests (`ConvertCqlIntInterval_Range`, `ConvertCqlIntInterval_Range_ConvertsOpenBoundsToClosed`, `ConvertCqlLongInterval_Range`, `ConvertCqlLongInterval_Range_ConvertsOpenBoundsToClosed`) supply fully-populated low/high values. Since Int/Long use a different step constant (`1m`) than Decimal (`MinDecimalPrecisionValue`), a regression isolated to the int/long null-endpoint path wouldn't be caught by the Decimal-only null-endpoint test. Please add null-endpoint tests for Integer and Long as well.

Cql/CoreTests/FhirTypeConverterTests.cs
baseTwo, line 840: **Test-coverage gap (required fix — minor):** No test supplies both bounds as null simultaneously (a fully unbounded interval) — `ConvertCqlInterval_Range_OmitsBoundsForNullEndpoints` only nulls the low bound while keeping a real high value. `NumericIntervalToRange` always constructs a new `Range` and only conditionally populates `Low`/`High`; the fully-unbounded path (both conditions false) is currently untested. Please add a both-bounds-null case for at least one type.

Cql/CoreTests/FhirTypeConverterTests.cs
```diff
@@ -672,9 +710,97 @@ public void ConvertCqlIntInterval_Range()
...
+
+            Assert.AreEqual(1, converted.Low.Value);
+            Assert.AreEqual("1", converted.Low.Unit);
+            Assert.AreEqual(Hl7.Fhir.ElementModel.Types.Quantity.UCUM, converted.Low.System);
+            Assert.AreEqual(0, QuantityPrecisionOf(converted.Low));
+            Assert.AreEqual(10, converted.High.Value);
+            Assert.AreEqual(0, QuantityPrecisionOf(converted.High));
+        }
+
+        [TestMethod]
```
baseTwo, line 797: **Test-coverage gap (required fix — minor):** `ConvertCqlLongInterval_Range_ConvertsOpenBoundsToClosed` only asserts `converted.Low.Value`/`High.Value` and omits the `QuantityPrecisionOf` checks that its Integer counterpart (`ConvertCqlIntInterval_Range_ConvertsOpenBoundsToClosed`) performs. The quantity-precision extension on successor/predecessor-derived Long bounds is therefore unverified — please add the same `QuantityPrecisionOf(converted.Low)`/`QuantityPrecisionOf(converted.High)` assertions here for symmetry with the Integer test.

Cql/Cql.Firely/FhirTypeConverter.cs
```diff
@@ -297,44 +297,15 @@ internal static TypeConverter ConvertCqlPrimitivesToFhir(this TypeConverter conv
...
-                    }
-
-                    if (interval.high is { } high)
-                    {
-                        range.High = new M.Quantity(high, "1");
-                    }
-                    return range;
-                }
-            });
+            converter.AddConversion((CqlInterval<decimal?> interval) => interval is null
```
baseTwo, line 307: **Test-coverage gap (required fix — minor):** The whole-object-null path (`(CqlInterval<decimal?>)null`/int/long → `Convert<Range>` returns null) is untested for all three numeric types, even though this PR rewrote that null-check for decimal/int (block if/else → ternary) and added it new for the long overload. Please add a test calling `Convert<Range>` with a null `CqlInterval<...>` instance for at least one of the three types to lock in the ternary's short-circuit behavior.

baseTwo (changes requested):
@copilot Thorough review of the Interval→Range mapping. One narrow correctness issue plus several test-coverage gaps below. Per this repo's standing convention, test-coverage gaps are required fixes, not nice-to-haves — please address all of them before this is ready to merge.

baseTwo:
@copilot The fix commit looks good — all 7 review comments are addressed and verified. This branch now conflicts with `develop` though (two other PRs merged while this was in review). Please merge `origin/develop` in, but **use the exact resolutions below rather than picking either side** — this is not a "take one side" conflict:

## `Cql/Cql.Runtime/Operators/CqlOperators.ListOperators.cs`

Develop's side of this hunk is PR #1421's actual Expand-overshoot bug fix (the `needsTruncation`/`TruncateToScale`/`UnitAtScale` logic, broadened from integer-only to any precision) — it must be kept, not discarded. This branch's side is only the pre-#1421 narrow `needsIntegerTruncation` logic with a cosmetic `.Scale`-property swap. Taking "HEAD" here would silently reintroduce the overshoot bug; taking "origin/develop" wholesale would keep the bug-fixed logic but also keep its `DecimalScale()` bit-manipulation helper that this PR's own commit (`2e31349f`, "use Decimal.Scale for decimal precision everywhere") is specifically trying to eliminate everywhere else in the codebase.

Resolve the conflicted hunk to exactly this (keep develop's truncation logic and naming, but use the built-in `.Scale` property instead of the `DecimalScale()` helper):

```csharp
                    var listItem = interval.low!.Value;
                    var highBoundary = interval.high!.Value;
                    var perValue = per.value ?? 1m;
                    var usesDefaultDecimalUnit = string.IsNullOrEmpty(per.unit) || per.unit == UCUMUnits.Unary;
                    var perScale = perValue.Scale;

                    // Boundaries more precise than per are truncated to per's scale, which may broaden the input range.
                    var needsTruncation = usesDefaultDecimalUnit
                        && (listItem.Scale > perScale || highBoundary.Scale > perScale);

                    if (needsTruncation)
                    {
                        listItem = TruncateToScale(listItem, perScale);
                        highBoundary = TruncateToScale(highBoundary, perScale);
                    }
```

Then delete the now-unused `DecimalScale` private helper further down in the same file (keep `UnitAtScale` and `TruncateToScale` — those aren't precision-detection helpers, they're part of the truncation fix itself and aren't touched by this PR's refactor):

```csharp
        /// <summary>
        /// The number of digits after the decimal point.
        /// </summary>
        private static int DecimalScale(decimal value) =>
            BitConverter.GetBytes(decimal.GetBits(value)[3])[2];
```

## `docs/releases/vnext-release-notes.md`

Purely additive on both sides — no semantic conflict. Keep **both** sides' entries under `## Breaking Changes` and `## Features` (this branch's new entries plus develop's existing ones), don't drop either.

## Before pushing

Please run the full `CoreTests` suite (or at minimum the `Expand`-related tests and `FhirTypeConverterTests`/`FhirTypeConverterHostConversionTests`) locally and confirm they pass before pushing the resolution — given a prior conflict resolution on a sibling PR in this queue silently reintroduced a regression, I'd like this one double-checked before I re-review.

baseTwo (approved):
All 7 findings are fixed and independently verified (the Int/Long overflow divergence is now documented as an intentional design choice; all 6 test-coverage gaps have new tests covering exactly the scenarios flagged). The develop-conflict resolution is also verified byte-for-byte correct: `CqlOperators.ListOperators.cs` keeps develop's #1421 overshoot-fix logic while adopting this PR's `.Scale`-based refactor, the now-redundant `DecimalScale()` helper is removed, and the same modernization was correctly extended to `IntervalOperators.cs`'s analogous point-form logic (a good catch beyond what I originally flagged). Release notes correctly combine both sides. Approving.

PR #1411, opened 2026-07-25
test: remove 14 stale XML-suite skip entries and correct outdated skip reasons

### LINKED ISSUE #1410: Remove stale skip entries from XML test suite (14 skipped DQIC tests already pass)
# Summary

`Cql/CqlToElmTests/(tests)/SkippedTests.cs` contains skip entries for the DQIC XML test suite (`Cql/CqlToElmTests/Input/DQIC/*.xml`) whose reasons are no longer true: an audit on `develop` (running every skipped test with the skip lists disabled, on net10.0) shows **14 of the 45 skipped tests pass today**. Several of the remaining entries have reason strings that no longer match the actual failure.

# Impact

- 14 passing conformance tests are reported as `Inconclusive` instead of contributing coverage, so regressions in those operators (`Power`, `ProperContains`/`ProperIn`, `Descendents`, list `Expand`) would go unnoticed.
- Stale reason strings ("Not working", "Power returns integers, not decimals") mislead anyone triaging the remaining skips.

# Observed Behavior

With the skip lists temporarily disabled, each of these skipped tests passes:

**`DoesNotCompile` list (2):**
- `ExpandIntervalPer2` (reason was "Not working")
- `DescendentsEmptyList` (reason was "Not implemented")

**`DoesNotMatchExpectation` list (12):**
- `DecimalNegOneStep`, `DecimalNegTenStep`, `DecimalNegTwoStep`, `DecimalOneStep`, `DecimalPosOneStep`, `DecimalPosTenStep`, `DecimalPosTwoStep`, `DecimalTenStep`, `DecimalTwoStep`, `Power2ToNeg2` (reason was "Power returns integers, not decimals" — obsolete: `SystemLibrary.Power` declares a Decimal result and `CqlOperators.Power(int?, int?)` returns `decimal?`)
- `ProperContainsTimeNull`, `ProperInTimeNull` (reason was "There is no spec language justifying null instead of false")

Additionally, some entries that still fail have reasons that no longer describe the actual error, e.g.:

- `ExpandPer0D1` ("Not working") — actually hangs the test host at runtime (infinite loop in `Expand` for an integer interval with a decimal `per`).
- `CodeToConcept1` ("The expectation requires ListPromotion to be enabled to evaluate properly.") — the expectation now dies with a bare `NotImplementedException` in `InvocationTermVisitor.VisitInstanceSelector` when coercing `codes:` from a single `Code`.
- `EquivEqCM1M01` ("Mixed unit equivalence is not supported.") — sits in `DoesNotCompile`, but the expression compiles and evaluates; it fails at runtime (`false` instead of `true`) because `CqlQuantityCqlComparer.EquivalentValues` does not canonicalize UCUM units the way `CompareValues` does.

# Expected Behavior

Skip lists only contain tests that genuinely fail, each with a reason string describing the current failure mode.

# Acceptance Criteria

- [x] The 14 tests listed above are removed from `SkippedTests.cs` and run green as part of `XmlTest`.
- [x] Reason strings for the remaining entries are corrected where they no longer match the observed failure (`ExpandPer0D1`, `CodeToConcept1`, `EquivEqCM1M01` moved/annotated as described, and any others found while editing — `ReplaceMatchesSpaces` also moved to `DoesNotMatchExpectation`, since it compiles and runs but produces Java-flavored replacement output).
- [x] Full `XmlTest` suite passes with no new failures (1506 passed, 0 failed, 31 inconclusive on net10.0).

alexzautke:
An audit running every skipped DQIC XML test with the skip lists disabled
showed 14 entries pass today (Power/Decimal step tests, ProperContains/
ProperIn time tests, DescendentsEmptyList, ExpandIntervalPer2). Remove
them so they contribute coverage again: XmlTest now runs 1507 passing
cases (was 1492), 30 inconclusive (was 45).

Also correct reasons that no longer match the observed failure:
- ExpandPer0D1 compiles but hangs at runtime (infinite loop in Expand)
- CodeToConcept1 dies with NotImplementedException in VisitInstanceSelector
- EquivEqCM1M01 and ReplaceMatchesSpaces compile and run; EquivEqCM1M01's
  underlying bug was separately fixed by #1415 before this branch's base,
  so it's removed outright rather than moved; ReplaceMatchesSpaces moves
  from DoesNotCompile to DoesNotMatchExpectation with an accurate reason

Fixes #1410

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_016gKrSSmxraX96iFCrM1Fk8


alexzautke:
@copilot Fix the merge conflict

Cql/CqlToElmTests/(tests)/SkippedTests.cs
```diff
@@ -35,11 +35,7 @@ internal static class SkippedTests
             { "ExpandPer0D1", "Throws NotSupportedException: a fractional per over integer intervals would produce Decimal intervals (value-dependent typing); the reference Java translator rejects this expression at compile time." },
             { "ExpandIntervalPer2", "Not working" },
```
baseTwo: **Blocker:** This PR's original commit (`9555f3e46`) correctly removed `ExpandIntervalPer2` from `DoesNotCompile` as one of the 14 now-passing tests. Develop independently merged an unrelated fix (#1412, commit `c946a2a43`) that only updated the adjacent `ExpandPer0D1` reason and left `ExpandIntervalPer2` untouched (since it predates this PR). When resolving the resulting conflict, taking develop's side of this hunk wholesale silently reinstated `{ "ExpandIntervalPer2", "Not working" }`.

I independently built this exact head commit and ran the full `XmlTest` suite twice: as-is it's 1505 passed / 32 skipped (matching your own PR comment); with only this entry removed it's 1506 passed / 31 skipped, 0 failed — the test genuinely passes today (`expand { Interval[1, 10] } per 2`). Your PR comment's claim that "ExpandIntervalPer2 remains skipped since it still fails after that fix" doesn't hold — fix #1412 only concerns non-advancing (zero/negative/fractional) `per` values; `per 2` is unaffected by it and was never broken. Please remove this entry again.

Cql/CqlToElmTests/(tests)/SkippedTests.cs
```diff
@@ -55,18 +51,8 @@ internal static class SkippedTests
...
-            { "DecimalOneStep", "Power returns integers, not decimals" },
-            { "DecimalPosOneStep", "Power returns integers, not decimals" },
-            { "DecimalPosTenStep", "Power returns integers, not decimals" },
-            { "DecimalPosTwoStep", "Power returns integers, not decimals" },
-            { "DecimalTenStep", "Power returns integers, not decimals" },
-            { "DecimalTwoStep", "Power returns integers, not decimals" },
-            { "Power2ToNeg2", "Power returns integers, not decimals" },
-            { "ProperContainsTimeNull", "There is no spec language justifying null instead of false" },
-            { "ProperInTimeNull", "There is no spec language justifying null instead of false" },
+            { "EquivEqCM1M01", "Returns false: CqlQuantityCqlComparer.EquivalentValues does not canonicalize UCUM units like CompareValues does." },
```
baseTwo: **Blocker:** This is a new skip entry this PR is adding, but the bug it describes (`CqlQuantityCqlComparer.EquivalentValues` not canonicalizing UCUM units) was already fixed *before this PR's own base commit* by commit `8641fc02d` ("fix: quantity Equivalent (~) now canonicalizes UCUM units", Fixes #1415), which explicitly removed `EquivEqCM1M01` from the skip dictionaries entirely because the test now passes. I confirmed `8641fc02d` is an ancestor of this PR's actual merge-base (`daef38dfe`), and that `EquivEqCM1M01` appears nowhere in either dictionary at that base commit.

This PR was authored against an older snapshot (before #1415 landed) and still believes the test needs to move to `DoesNotMatchExpectation` with a reason describing the now-fixed bug. Please drop this entry entirely — the test currently passes and should stay unskipped.

Cql/CqlToElmTests/(tests)/SkippedTests.cs
```diff
@@ -35,11 +35,7 @@ internal static class SkippedTests
             { "ExpandPer0D1", "Throws NotSupportedException: a fractional per over integer intervals would produce Decimal intervals (value-dependent typing); the reference Java translator rejects this expression at compile time." },
```
baseTwo, line 35: **Nit (informational, not blocking):** The file content here is currently correct (it kept develop's accurate "Throws NotSupportedException..." text from #1412's fix), but this PR's own body/first-commit-message still describes this as "compiles but hangs at runtime (infinite loop in Expand)" — that description was superseded by #1412 shortly after this PR's first commit and is no longer accurate. Worth updating the PR description so it doesn't mislead future readers, even though no file change is needed here.

baseTwo (changes requested):
@copilot The merge-conflict resolution you pushed (`73c964745`) silently reintroduced two stale skips that this PR itself was supposed to remove — both are CI-invisible regressions (a `DoesNotCompile`/`DoesNotMatchExpectation` entry causes `Assert.Inconclusive`, not a failure, so CI stays green). I verified both empirically by running the suite with each entry removed. Also: the PR description's headline result counts ("1506 passing ... 31 inconclusive") are now stale — they describe the original single commit, not the current head (which is actually 1505/32 until the `ExpandIntervalPer2` regression below is fixed). Please correct both entries and update the PR description's counts once done.

baseTwo (approved):
Both blockers are fixed and verified against the actual file content: `ExpandIntervalPer2` and `EquivEqCM1M01` are both fully removed from the skip dictionaries (commit `c35e5896`). The reconciled counts (1507 passing / 30 inconclusive = 1492+14+1 / 45-14-1) check out against the original baseline plus this PR's 14 fixes plus the additional EquivEqCM1M01 fix. I've corrected the PR description to reflect these verified counts (the earlier claim of having updated it hadn't actually taken effect). Approving.

Issue #1418, opened 2026-07-25
CqlQuantity hash code disagrees with equality/equivalence under unit conversion


alexzautke:
# Summary

`CqlQuantityCqlComparer.GetHashCodeValue` hashes `CqlQuantity.ToString()`, so two quantities that compare equal/equivalent after UCUM unit conversion (`1 'cm'` and `0.01 'm'`) land in different hash buckets. Hash-based operations built on `CqlComparers` — `Distinct`, `Union`, `Except`, and other `HashSet`-backed paths — will fail to deduplicate unit-converted duplicates while the pairwise comparer says they are equal.

The inconsistency predates #1415 (it already applied to the `=`/`Compare` path); the #1415 fix widens its surface by making the equivalence path unit-converting as well.

# Repro Steps

```cql
distinct { 1 'cm', 0.01 'm' }  // returns both elements; the comparer says they are equal
```

# Expected Behavior

The hash contract holds: values that compare equal/equivalent hash identically. A spec-correct implementation hashes the canonicalized value + canonical unit, falling back to `ToString()` only when canonicalization fails.

# Acceptance Criteria

- [x] `GetHashCodeValue` hashes the canonicalized quantity (scale-normalized value + canonical unit) when canonicalization succeeds. It also fixes the subtler same-unit case: `1.0 'cm'` vs `1.00 'cm'` hashed differently because `decimal.ToString()` preserves trailing zeros.
- [x] `distinct`/`union` over unit-converted duplicates deduplicate correctly, with tests (end-to-end `ICqlOperators.Distinct` over `{ 1 'cm', 0.01 'm' }` and `{ 1.0 'cm', 1.00 'cm' }`).
- [x] Hash behavior for non-canonicalizable units: falls back to the original unit with a scale-normalized value (better than `ToString()` — fixes trailing-zero splits there too), and quantities with null value/unit hash without throwing.
- [x] Full comparer and `XmlTest` suites green (CoreTests 840 passed / 0 failed; XmlTest 1493 / 44 / 0, unchanged from branch baseline).

Documented limits: CQL's rounding-based decimal equivalence is non-transitive and has no consistent hash, and the `'1'` unit compares equal to every unit — extra collisions/splits there are inherent. The sub-1e-8 truncation gap is shared with bare decimals and can be tightened separately (collisions are safe).

Fixed in PR #1423 (stacked on #1419).

Issue #1417, opened 2026-07-25
Quantity Compare/Equal is dimension-blind: 1 'cm' = 0.01 'g' returns true


alexzautke:
# Summary

`CqlQuantityCqlComparer.CompareValues` (Cql/Cql.Runtime/Comparers/CqlComparers.CqlQuantityCqlComparer.cs) canonicalizes each quantity independently via `TryCanonicalize` and then compares only the canonicalized *values*, never checking that the canonical *units* agree. Canonicalization succeeds per-quantity for any valid UCUM unit, so quantities of entirely different dimensions are compared numerically.

Surfaced while fixing #1415 — the equivalence path got a canonical-unit guard there, but the Compare/Equal path deliberately kept its existing behavior and needs the same guard.

# Repro Steps

```cql
1 'cm' = 0.01 'g'   // true today: cm canonicalizes to 0.01 'm', g stays 0.01 'g', values compare equal
1 'cm' < 1 'g'      // also answers as if both were dimensionless
```

# Observed Behavior

`1 'cm' = 0.01 'g'` → `true`; ordering comparisons between incommensurable quantities return a numeric answer instead of failing.

# Expected Behavior

**Corrected after a first implementation attempt** (the original text here prescribed routing incommensurable units to the existing `NotSupportedException` path — that contradicts the spec and broke the CMS integration tests on PR #1425, where real measure data compares `%` vs `/min` (CMS144) and `cm` vs `weeks` (CMS0334)).

The spec states, identically in every comparison operator section (§9.B Equal, Less, Greater, Greater Or Equal, between): "For comparisons involving quantities, the dimensions of each quantity must be the same, but not necessarily the unit. … **Attempting to operate on quantities with invalid units will result in a null.**"

So: `CompareValues` checks that both quantities canonicalize to the same base unit before comparing values (mirroring the #1415 guard in `EquivalentValues`); when they don't — different dimensions or non-canonicalizable units — the comparison result is **null**, which propagates as unknown through measure logic instead of crashing evaluation. Equivalence remains `false` for such pairs (equivalence is never null and never errors).

# Acceptance Criteria

- [x] `CompareValues` only compares canonicalized values when the canonical units agree.
- [x] Incommensurable or non-canonicalizable units yield **null** from comparison (`=`, `<`, `>`, `<=`, `>=`), per the spec sentence above — no exception (the pre-existing `NotSupportedException` for invalid UCUM units is removed too; the spec sentence covers both).
[...]

PR #1419, opened 2026-07-25
Fix quantity Equivalent to canonicalize UCUM units

### LINKED ISSUE #1415: Quantity Equivalent (~) ignores UCUM unit conversion that Equal/Compare already support
# Summary

`CqlQuantityCqlComparer` (Cql/Cql.Runtime/Comparers/CqlComparers.CqlQuantityCqlComparer.cs) canonicalizes UCUM units in `CompareValues` — when units differ it calls `TryCanonicalize` on both quantities and compares the normalized values — but `EquivalentValues` skips that step and returns `false` outright for unlike units. As a result `1 'cm' ~ 0.01 'm'` is `false` while `1 'cm' = 0.01 'm'` is `true`, which is internally inconsistent and the opposite of the spec's intent.

# Impact

- Equivalence-based logic (`~`, `!~`, and everything built on equivalence semantics such as list membership via equivalence) silently misses matches between convertible quantities, e.g. cm vs m, mg vs g.
- The DQIC XML conformance test `EquivEqCM1M01` (`1'cm' ~ 0.01'm'` → `true`) is skipped because of this.

# Observed Behavior

`1 'cm' ~ 0.01 'm'` returns `false`. The comparer's `EquivalentValues` only consults `UnitComparer.Equivalent` and the default-unit (`'1'`) special case before giving up; the `TryCanonicalize` fallback exists a few lines up in `CompareValues` but is not applied.

# Expected Behavior

CQL spec (§9.B, Equivalent): "For quantities, equivalent means the values are the same quantity when considering unit conversion (e.g. 100 'cm' ~ 1 'm') and using decimal equivalent semantics for the value." The spec allows implementations without unit-conversion support to return `false`, but this SDK supports UCUM conversion (Fhir.Metrics) and already uses it in `CompareValues` — `EquivalentValues` should apply the same canonicalization fallback, returning `true` for convertible equal quantities and `false` (not an error) when canonicalization fails (incommensurable units).

Calendar/definite-duration equivalences that already work through `UnitComparer.Equivalent` (e.g. `1 year ~ 1 'a'`) must be unaffected — the canonicalization fallback only runs when the unit comparer says the units differ.

# Acceptance Criteria

- [x] `EquivalentValues` canonicalizes unlike units the same way `CompareValues` does; equal-after-conversion quantities are equivalent (`1 'cm' ~ 0.01 'm'` → `true`), incommensurable units remain `false` without throwing (implementation additionally guards that both quantities canonicalize to the same base unit, since `TryCanonicalize` succeeds per-quantity — the equivalent latent gap in `CompareValues` is tracked in #1417).
- [x] `EquivEqCM1M01` is removed from `SkippedTests.cs` and runs green in `XmlTest` (1493 passed / 44 skipped / 0 failed).
- [x] Unit tests cover: convertible units equivalent (`~` true), incommensurable units (`~` false, including the equal-magnitude case), calendar-duration equivalences (`1 year ~ 1 'a'` — now true via the canonicalization fallback; the unit comparer alone never handled these), and no-throw behavior for invalid units.
- [x] Full `XmlTest` and comparer test suites pass with no new failures (CoreTests 835 passed / 0 failed).
- [x] A note is added under `## Fixes` in `docs/releases/vnext-release-notes.md`.

Fixed in PR #1419. Follow-ups filed: #1417 (dimension-blind `CompareValues`), #1418 (hash code disagrees with unit-converted equality).

alexzautke:
## Description

**Primary Work**

`CqlQuantityCqlComparer.EquivalentValues` returned `false` for any two quantities whose units weren't textually equivalent, even though `CompareValues` (and therefore `=` and the ordering operators) already canonicalizes via UCUM — so `1 'cm' = 0.01 'm'` was `true` while `1 'cm' ~ 0.01 'm'` was `false`. Spec §9.B requires quantity equivalence to consider unit conversion.

`EquivalentValues` now falls back to the same `TryCanonicalize` when the unit comparer reports unlike units, comparing the canonicalized values **only when both quantities canonicalize to the same base unit**. Equivalence never signals an error: invalid units and different base metrics are simply not equivalent. The base-unit guard matters because canonicalization succeeds per-quantity — without it, equal-magnitude quantities of different dimensions (`1 'cm'` vs `0.01 'g'`) would compare equivalent.

Side effect (spec-aligned): calendar/UCUM duration pairs (`1 year ~ 1 'a'`, `1 month ~ 1 'mo'`, …) are now equivalent via `Ucum.CalendarDurationMapping`; previously the unit comparer treated them as plain strings. Interval- and list-of-quantity equivalence pick the fix up automatically through the comparer.

Unskips the DQIC conformance test `EquivEqCM1M01`, plus a release-note bullet under Fixes.

**Auxiliary Work**

Two pre-existing defects surfaced during this work are tracked separately rather than fixed here: `CompareValues` is dimension-blind (`1 'cm' = 0.01 'g'` returns `true` today) — #1417 — and the quantity hash code disagrees with unit-converted equality, breaking hash-based dedup like `distinct` — #1418.

## Related issues

Fixes #1415. Refs #1417, #1418.

## Testing

13 new comparer tests in `CqlComparersTests.cs`: convertible units equal/unequal magnitude, incommensurable units (equal magnitude included — the guard case), non-UCUM units (no throw), all 8 calendar/UCUM duration pairs, and the default-`'1'` short-circuit. Verified on net10.0: XmlTest **1493 passed / 44 skipped / 0 failed** (exactly the +1/−1 delta from unskipping `EquivEqCM1M01`), full CoreTests **835 passed / 0 failed**.

## FirelyTeam Checklist

https://claude.ai/code/session_016gKrSSmxraX96iFCrM1Fk8

---
_Generated by [Claude Code](https://claude.ai/code/session_016gKrSSmxraX96iFCrM1Fk8)_


baseTwo:
@copilot please:

1. Merge `develop` into this branch and resolve the conflicts (`mergeStateStatus: DIRTY`).
2. Move this PR's release-note entry out of `docs/releases/vnext-release-notes.md` and into a new fragment file `docs/releases/vnext/1415-quantity-equivalent-ucum.md` instead, following the new convention from #1432 (see [docs/releases/vnext/README.md](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/releases/vnext/README.md)). Same `## Fixes` section content, just in its own file.

Note: PRs #1423 and #1425 are stacked directly on top of this branch, so once this merges I'll rebase those myself — no action needed on them for this.

This review has already passed on my end — verified the base-unit guard (`left1.unit == right1.unit`) correctly avoids the dimension-blind bug tracked separately as #1417, ran all 13 new comparer tests plus 2 of my own confirming list-of-quantity/interval-of-quantity equivalence pick up the fix automatically through the comparer composition. Just need the rebase + release-note move before this is ready to merge.

baseTwo (dismissed):
Approving. Verified:

- The fix correctly adds a base-unit-match guard (`left1.unit == right1.unit`) after canonicalizing both quantities — this is exactly the guard the sibling `CompareValues` method is *missing* (tracked separately as #1417, the dimension-blind bug). The PR deliberately doesn't repeat that bug here; confirmed by tracing both methods side by side.
- Ran all 13 new `CqlComparersTests` cases: pass. Total suite 21/21.
- Verified empirically (not just architecturally) that list-of-quantity and interval-of-quantity equivalence pick up this fix automatically through the comparer composition — wrote two throwaway tests (`Equivalent` on `List<CqlQuantity>` and `CqlInterval<CqlQuantity>` with convertible-but-different units) and both pass, confirming the PR body's claim.
- Calendar/UCUM duration pairs (`year`/`a`, `month`/`mo`, etc.) all 8 tested and pass as a side effect of the same fix, matching spec §9.B.

No outstanding concerns. Just needs the develop merge + release-note fragment move (see my other comment) before this is mergeable.

baseTwo (approved):
Re-approving after the develop merge + release-note fragment move — verified the actual fix diff (`CqlComparers.CqlQuantityCqlComparer.cs`) is byte-identical to what I already reviewed, and the release note correctly moved to `docs/releases/vnext/1415-quantity-equivalent-ucum.md`.

PR #1423, opened 2026-07-25
Fix quantity hash code to agree with unit-converted equality

### LINKED ISSUE #1418: CqlQuantity hash code disagrees with equality/equivalence under unit conversion

alexzautke:
## Description

> **Stacked PR**: based on `claude/fix-quantity-equivalent-ucum-1415` (#1419), since it modifies the same comparer and removes a comment that PR introduced. Only the topmost commit is this PR's change; it should merge after #1419 (GitHub will retarget to `develop`).

**Primary Work**

`CqlQuantityCqlComparer.GetHashCodeValue` hashed `CqlQuantity.ToString()`, so quantities that compare **equal** after UCUM unit conversion (`1 'cm'` vs `0.01 'm'`) — or that differ only in decimal scale (`1.0 'cm'` vs `1.00 'cm'`) — landed in different hash buckets. The `HashSet`-backed list operators (`distinct`, `union`, `except`), which route through `EqualityComparerBridge` → `CqlComparers.GetHashCode` → this comparer, therefore failed to deduplicate values the pairwise comparer said were equal.

The hash is now `HashCode.Combine` over the canonicalized quantity's scale-normalized value and canonical unit, falling back to the original (still scale-normalized) value and unit when UCUM cannot canonicalize. Quantities with a null value or unit reach this method (the comparer's `IsNull` only rejects null references) and now hash without throwing. `NormalizeDecimalScale` lives in `CqlComparerSharedMethods` so other comparers can share it.

Documented limits (in code): CQL's rounding-based decimal equivalence is non-transitive (`0.15 ~ 0.2`, `0.2 ~ 0.24`, `0.15 !~ 0.24` — verified against the actual comparer) and therefore has no consistent hash; the `'1'` unit compares equal against every other unit and is equally unhashable. Unit conversion + scale normalization is the whole of what a hash can cover, and it covers all of `Equals`.

**Auxiliary Work**

The sub-1e-8 truncation `DecimalCqlComparer` applies in comparisons is a pre-existing hash gap shared with bare decimals (extra collisions are safe, so it can be tightened separately). The dimension-blind `Equals` cases (#1417) still hash apart — correctly so once #1417 lands.

## Related issues

Fixes #1418. Refs #1417.

## Testing

Five new tests in `CqlComparersTests.cs`: hash agreement for unit-converted equals and for scale-only differences, no-throw hashing for uncanonicalizable/partly-null quantities, non-collision for everyday unequal cases, and an end-to-end `ICqlOperators.Distinct` dedup over `{ 1 'cm', 0.01 'm' }` and `{ 1.0 'cm', 1.00 'cm' }`. Verified on net10.0: CoreTests **840 passed / 0 failed** (835 baseline + 5), XmlTest **1493 / 44 / 0** — identical to the branch baseline, no regressions.

## FirelyTeam Checklist

https://claude.ai/code/session_016gKrSSmxraX96iFCrM1Fk8

---
_Generated by [Claude Code](https://claude.ai/code/session_016gKrSSmxraX96iFCrM1Fk8)_

### DIFF
--- a/Cql/CoreTests/CqlComparersTests.cs
+++ b/Cql/CoreTests/CqlComparersTests.cs
@@ -417,21 +417,31 @@ public class CqlComparersTests
     }
 
     /// <summary>
-    /// Quantities that are not equal are still allowed to collide, but the everyday cases must not:
-    /// a different magnitude in the same unit, and the same magnitude in an incommensurable unit.
+    /// Quantities that are not equal should be kept as separate elements by <c>Distinct</c>.
+    /// Verifies semantic behavior (via <c>Distinct</c>) rather than asserting hash inequality —
+    /// hash functions are permitted to collide, and asserting inequality would over-specify
+    /// and risk flakiness.
     /// </summary>
     [TestMethod]
-    public void CqlQuantity_UnequalQuantities_HashCodesDiffer()
+    public void CqlQuantity_UnequalQuantities_AreNotDeduplicated()
     {
-        var comparers = new CqlComparers();
+        var operators = FhirCqlContext.WithDataSource().Operators;
 
-        Assert.AreNotEqual(
-            comparers.GetHashCode(new CqlQuantity(1.0m, "cm")),
-            comparers.GetHashCode(new CqlQuantity(1.01m, "cm")));
+        // Different magnitude in the same unit must not be collapsed.
+        var differentMagnitude = operators.Distinct<CqlQuantity>(
+        [
+            new CqlQuantity(1.0m, "cm"),
+            new CqlQuantity(1.01m, "cm"),
+        ])!.ToList();
+        Assert.AreEqual(2, differentMagnitude.Count);
 
-        Assert.AreNotEqual(
-            comparers.GetHashCode(new CqlQuantity(1m, "cm")),
-            comparers.GetHashCode(new CqlQuantity(1m, "g")));
+        // Incommensurable units (cm vs g) must not be collapsed.
+        var incommensurable = operators.Distinct<CqlQuantity>(
+        [
+            new CqlQuantity(1m, "cm"),
+            new CqlQuantity(1m, "g"),
+        ])!.ToList();
+        Assert.AreEqual(2, incommensurable.Count);
     }
 
     /// <summary>
--- a/Cql/CoreTests/CqlComparersTests.cs
+++ b/Cql/CoreTests/CqlComparersTests.cs
@@ -9,6 +9,7 @@
 #nullable enable
 
 using Hl7.Cql.Comparers;
+using Hl7.Cql.Fhir;
 using Hl7.Cql.Primitives;
 
 namespace CoreTests;
@@ -236,10 +237,7 @@ public class CqlComparersTests
     // Regression tests for #1415: CqlQuantityCqlComparer.EquivalentValues used to return false as
     // soon as the two units weren't textually equivalent, even though CompareValues (and therefore
     // = and the comparison operators) already canonicalized via UCUM. Spec §9.B requires quantity
-    // equivalence to consider unit conversion.
-    // Note: GetHashCodeValue still hashes CqlQuantity.ToString(), so two quantities that are now
-    // equivalent (or equal) via unit conversion hash differently. That inconsistency predates this
-    // fix -- it applies equally to the = path -- and is deliberately left alone here.
+    // equivalence to consider unit conversion. The matching hash code fix is #1418, tested below.
 
     /// <summary>
     /// Different but convertible units with the same magnitude are equivalent -- the spec's own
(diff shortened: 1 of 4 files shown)


baseTwo (dismissed):
Verified (in addition to the stacked-PR rebase I already did to get this onto `develop`):

- Traced `NormalizeDecimalScale` by hand — correctly strips only numerically-redundant trailing zeros, and confirmed it's genuinely needed on both the canonicalized and fallback hash paths (decimal arithmetic accumulates scale during canonicalization, so 1.0 'cm' and 1.00 'cm' can canonicalize to different raw scales despite being numerically equal).
- Test coverage is thorough, including an end-to-end `Distinct` test proving real dedup across both the unit-conversion and scale-only cases, not just the raw comparer.
- Ran the full rebased branch: CoreTests 869/0, `CqlComparersTests` 26/26.

Left two follow-up observations as separate inline review comments (tagging @copilot) rather than folding them in here:

- The hash including the canonical unit means it correctly disagrees with the still-open #1417 bug (`CompareValues`'s missing dimension guard) for that one case — worth confirming this resolves itself once #1417 lands.
- `GetHashCodeValue` always attempts `TryCanonicalize`, unlike `Compare`/`Equivalent` which short-circuit first — a possible micro-optimization, not a correctness issue.

See those threads for the full discussion before merging.

Cql/Cql.Runtime/Comparers/CqlComparers.CqlQuantityCqlComparer.cs
```diff
@@ -78,7 +78,25 @@ protected override bool EquivalentValues(
...
+            // and must land in the same bucket for the HashSet-based operators (Distinct, Union,
+            // Except) to deduplicate them. Value normalization covers the same-unit case, where
+            // 1.0 'cm' and 1.00 'cm' are equal but have different decimal representations.
+            //
+            // Unit conversion and scale are all that can be covered here. Equivalence rounds both
+            // operands to the least precise of the two, which is not transitive (0.15 ~ 0.2 and
+            // 0.2 ~ 0.24, but 0.15 !~ 0.24) and therefore has no consistent hash; neither does the
+            // '1' unit, which compares equal against every other unit.
+            if (value.TryCanonicalize(out var canonical))
+                return combine(canonical!.value, canonical.unit);
```
baseTwo, line 102: @copilot 
Worth flagging explicitly (not blocking, and this is already partly disclosed in the PR body): since this hash includes the canonical **unit**, it correctly does *not* agree with the still-open #1417 bug (`CompareValues`'s missing dimension guard, e.g. `1 'cm' == 0.01 'g'` today). That means hash and equals technically diverge for that one already-known-buggy case until #1417 lands — the failure mode is a missed `Distinct`/`Union`/`Except` dedup, not an incorrect merge, so I read this as a sound forward-looking choice. Flagging as an inline note so it's easy to find once #1417 is being worked, to confirm the hash naturally becomes consistent again at that point (it should, since the unit comparison would then agree with `CompareValues`).

Cql/Cql.Runtime/Comparers/CqlComparers.CqlQuantityCqlComparer.cs
```diff
@@ -78,7 +78,25 @@ protected override bool EquivalentValues(
...
+            // so the hash has to be taken over the canonical form: 1 'cm' and 0.01 'm' are equal
+            // and must land in the same bucket for the HashSet-based operators (Distinct, Union,
+            // Except) to deduplicate them. Value normalization covers the same-unit case, where
+            // 1.0 'cm' and 1.00 'cm' are equal but have different decimal representations.
+            //
+            // Unit conversion and scale are all that can be covered here. Equivalence rounds both
+            // operands to the least precise of the two, which is not transitive (0.15 ~ 0.2 and
+            // 0.2 ~ 0.24, but 0.15 !~ 0.24) and therefore has no consistent hash; neither does the
+            // '1' unit, which compares equal against every other unit.
+            if (value.TryCanonicalize(out var canonical))
```
baseTwo: @copilot 
Minor, non-blocking: this always attempts `TryCanonicalize` regardless of unit, unlike `Compare`/`Equivalent` above which short-circuit on a textually-matching unit or the default `'1'` unit before ever canonicalizing. Since hashing is O(1) per item (not O(n²) like pairwise comparison), this isn't a correctness or scaling concern — just a possible micro-optimization for a future pass if quantity hashing shows up in profiling.

Cql/Cql.Runtime/Comparers/CqlComparers.CqlQuantityCqlComparer.cs
baseTwo: @copilot could you take a look at this and consider addressing it if it's a quick fix (not blocking merge)?

Cql/Cql.Runtime/Comparers/CqlComparers.CqlQuantityCqlComparer.cs
baseTwo: @copilot could you take a look at this and consider addressing it if it's a quick fix (not blocking merge)?

baseTwo: Fixed in 187196cc7 — confirmed the short-circuit (`value.unit != null && value.unit != "1"`) skips canonicalization for null/wildcard units, and re-ran the tests: `CqlComparersTests` 26/26 still pass.

Cql/Cql.Runtime/Comparers/CqlComparers.CqlQuantityCqlComparer.cs
baseTwo: Addressed in 187196cc7 — the expanded comment now documents the '1'-wildcard non-transitivity, the rounding-based equivalence non-transitivity, and the #1417 interaction explicitly in code, not just in review discussion. Good clarification, thanks.

Cql/Cql.Runtime/Comparers/CqlComparers.CqlQuantityCqlComparer.cs
```diff
@@ -78,7 +78,25 @@ protected override bool EquivalentValues(
...
+                return combine(canonical!.value, canonical.unit);
+
+            // A unit UCUM cannot canonicalize -- and a quantity whose value or unit is null, which
+            // this comparer does not treat as a null quantity -- must still hash without throwing.
```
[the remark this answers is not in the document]
baseTwo: Fixed in 187196cc7 — good catch. This is already the accepted, documented trade-off (see the expanded comment in `GetHashCodeValue`): the '1'-wildcard and #1417 dimension-blind cases are non-transitive and have no consistent hash by construction, so they're intentionally left to diverge from Equals until #1417 lands. Not something further code can fix without breaking the equality semantics themselves.

Cql/CoreTests/CqlComparersTests.cs
```diff
@@ -347,4 +345,119 @@ public void CqlQuantity_DefaultUnit_ComparesValuesDirectly()
...
+        var comparers = new CqlComparers();
+
+        Assert.AreNotEqual(
+            comparers.GetHashCode(new CqlQuantity(1.0m, "cm")),
+            comparers.GetHashCode(new CqlQuantity(1.01m, "cm")));
+
+        Assert.AreNotEqual(
+            comparers.GetHashCode(new CqlQuantity(1m, "cm")),
+            comparers.GetHashCode(new CqlQuantity(1m, "g")));
+    }
```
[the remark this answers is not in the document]
baseTwo: Fixed in 187196cc7 — the test was renamed to `CqlQuantity_UnequalQuantities_AreNotDeduplicated` and now asserts via `Distinct` semantic behavior instead of hash inequality directly, exactly as suggested. Verified it passes.

baseTwo (dismissed):
Re-approving after the review-comment fixes (commit 187196cc7) — verified both my observations and GitHub Copilot's two review comments are addressed: the short-circuit skips canonicalization for null/wildcard units, the hash-contract gap is now documented in code, and the flaky hash-inequality test was replaced with a Distinct-based semantic assertion. All 4 threads resolved. `CqlComparersTests` 26/26 pass.

baseTwo:
@copilot please revert the `CompareValues` change from the last commit (bb3b966bf) — it's out of scope for this PR and, more importantly, reintroduces a regression that #1425 (the dedicated PR for #1417) already found and fixed.

Adding `&& left1!.unit == right1!.unit` to the canonicalization fallback in `CompareValues` means incommensurable units now fall through to the existing `throw new NotSupportedException(...)` at the end of the method. But #1425's own PR description explains that exact approach was tried first and broke the CMS integration suite (real measure data compares things like `'%'` vs `'/min'`), which is why #1425 changed the result to `null` (per CQL spec §9.B: "Attempting to operate on quantities with invalid units will result in a null") instead of throwing, and also updated `CqlComparisonToEquivalence`'s null-mapping to match.

This PR (#1423) should stay scoped to the hash fix (#1418) only. Please:
1. Revert the `CompareValues` change (restore it to just `x.TryCanonicalize(out var left1) && y.TryCanonicalize(out var right1)` without the dimension guard).
2. Restore the doc comment noting the #1417 dimension-blind gap as a known, deferred issue (the one you removed in this same commit).
3. Leave #1417 entirely to #1425, which already has the correct, tested fix.

baseTwo:
Confirmed in 7971ade03 — `CompareValues` is back to identical to `develop`'s current state (verified via diff), the #1417 doc-comment note is restored, and this PR is correctly scoped to #1418 only. Re-ran `CqlComparersTests`: 26/26 pass. Thanks for catching and fixing this quickly.

baseTwo (approved):
Re-approving after the scope revert (7971ade03) — confirmed `CompareValues` is now identical to develop's, #1417 is correctly left entirely to #1425, and `CqlComparersTests` 26/26 still pass.

PR #1430, opened 2026-07-27
Upgrade Java cql-to-elm-cli/elm-fhir tooling: 4.0.0 -> 5.0.0

### LINKED ISSUE #1334: Upgrade cql-to-elm tooling past 4.0.0 (4.9.0 or 5.0.0) - requires ELM-shape verification
## Summary

Follow-up to #1333, which lands the CQL-to-ELM tooling upgrade at `org.cqframework 4.0.0` (Stage 1 of that ticket's plan). This ticket covers Stage 2: upgrading from `4.0.0` to the current `5.0.0`.

## Impact

Unlike the 3.29.0 -> 4.0.0 jump (which was purely cosmetic plus one additive field), `5.0.0` introduces real compilation-strategy changes in the cql-to-elm translator itself, not just ELM formatting or engine-side API changes. Investigation found this out empirically by regenerating ELM for all 4 library sets at `5.0.0` and diffing against `4.0.0` output.

## Observed Behavior

Regenerated ELM at `5.0.0` for all 4 library sets and structurally compared (order-independent, `localId`-stripped, `choice`-array-unordered) against `4.0.0` output:

| Library set | Files affected (of total) |
|---|---|
| CoreTests.HL7 | 0 / 13 |
| Measures.Authoring | 1 / 7 |
| Demo | 4 / 38 |
| dqm-content-qicore-2025 | **61 / 91** |

Three distinct real (non-cosmetic) differences found:

1. **Interval-literal boundary constant-folding**: `Interval[low, high).high` is now compiled directly to the `high` expression itself, instead of `Property(path="high", source=Interval{...})`. Semantically equivalent optimization, lowest risk of the three.
2. **Retrieve -> Query restructuring** (20 files): retrieves filtering on a non-default code property (e.g. `[MedicationRequest: medication in "ValueSet"]`) are no longer folded into the `Retrieve` node's native `codeFilter`. They now compile as a plain `Retrieve` plus an explicit `Query`/`where` clause instead.
3. **`asType`/`isType` vs `asTypeSpecifier`/`isTypeSpecifier` representation swap** (44 files): `as`/`is` casts to `Interval`-typed values sometimes serialize as the short-form string property (`asType`/`isType`) instead of the full specifier object (`asTypeSpecifier`/`isTypeSpecifier`), inconsistently between otherwise-similar case branches in the same file.

## Root Cause

Not investigated in depth yet - these appear to be intentional compiler/optimizer changes in the `org.cqframework/cql-to-elm` 5.0.0 release (a major version bump), not bugs. The upstream 5.0.0 release notes describe it as "a major version bump with breaking API changes affecting both Kotlin and Java implementations" primarily at the *engine* level (sealed `Value` interface). The differences found here are additionally at the *cql-to-elm* (ELM-shape) level, which wasn't called out as explicitly in the release notes reviewed during #1333's research.

## Expected Behavior

Our `Cql.Compiler`/`CodeBuilder` should correctly compile all valid ELM-5.0.0 shapes it can produce, matching evaluation semantics with the CQL source.

## Evidence Gathered So Far

- Full solution build (`Cql-Sdk-All.sln`) with 5.0.0-generated ELM: 0 errors.
- `CoreTests` full suite (734 tests) against 5.0.0-generated ELM: all pass.
- `Test.Measures.Demo` (3 tests) against 5.0.0-generated ELM: all pass.
- **Update**: re-ran the same investigation bumping only to the latest `4.x` release (`4.9.0`, one step before `5.0.0`) instead of `5.0.0` directly. The exact same three patterns appear, with the exact same per-library-set file counts (0/13, 1/7, 4/38, 61/91) — meaning these are **not specific to the `5.0.0` major version bump**; they were already introduced somewhere within the `4.0.0` -> `4.9.0` line. Whatever upgrade path we take past `4.0.0` will hit these changes.
- **This time a real test failure surfaced**: `CSharpGenerationGoldenTests.Regenerated_DqmQiCore2025_CMS56_CSharp_Matches_CheckedInFiles` fails against `4.9.0`-generated ELM (733/734 `CoreTests` passing instead of 734/734). Diffing the expected (checked-in golden `.g.cs`) vs. actual (regenerated) C# shows the generator now emits `is`/`as` type-check branches in the **opposite order** for a `case when ... is CqlInterval<CqlDateTime> ... when ... is CqlQuantity ...` construct — i.e. this is the concrete downstream consequence of the `asType`/`asTypeSpecifier` representation swap pattern (#3 below), not just an abstract ELM-shape curiosity. `CqlInterval<CqlDateTime>` and `CqlQuantity` are mutually exclusive types, so the swap is likely semantically safe for any single input, but this needs explicit verification (and the golden file needs regenerating) rather than being assumed.
- This resolves the earlier "gap" (no dedicated `dqm-content-qicore-2025` test coverage) — the golden-file test for CMS56 *is* effectively that coverage, and it directly caught an impact from this change once actually run against the new ELM.

## Acceptance Criteria

- [x] Trace the `Retrieve` -> `Query` restructuring and `asType`/`asTypeSpecifier` swap through `Cql.Compiler`'s `CodeBuilder` to confirm both old and new shapes are handled correctly (not just "doesn't throw", but produces equivalent runtime results).
- [x] Specifically verify the `CqlInterval<CqlDateTime>`/`CqlQuantity` case-branch reordering found in CMS56 (`Regenerated_DqmQiCore2025_CMS56_CSharp_Matches_CheckedInFiles`) doesn't change behavior for any real input, then regenerate the checked-in golden `.g.cs` file to match.
- [x] Add or identify a test that specifically exercises a `[Resource: nonDefaultProperty in "ValueSet"]`-style retrieve (triggers the Retrieve -> Query pattern), run it against both `4.0.0` and the target version's ELM to confirm equivalent behavior.
- [x] Decide the target version: since the ELM-shape changes already appear by `4.9.0`, consider whether to land on `4.9.0` (still avoids the engine-level sealed `Value` interface break at `5.0.0`) or go straight to `5.0.0` — both require the same ELM-shape verification work, so `5.0.0` may not add meaningfully more risk once that's done. Re-run the same regenerate-and-diff process against whichever version is chosen, since intermediate `4.x` releases weren't individually checked.
- [x] Version and groupId bumped from `4.0.0` to the target version across the same 4 locations as #1333: `Demo/Cql/Build/pom.xml`, `Demo/Cql/Build/Java-Dependencies-Vars.ps1`, `Demo/Cql/Build/Java-Dependencies-Vars.sh`, `Cql/PackagerCLI/Program.cs`.
- [x] ELM regenerated for all 4 library sets (CoreTests.HL7, Measures.Authoring, Demo, dqm-content-qicore-2025) at the target version, and the `CMS56` golden `.g.cs` file regenerated to match.
- [x] Full `CoreTests` suite (including `CSharpGenerationGoldenTests`) passes 100% against the new ELM — do not proceed on partial/mostly-passing results.
- [x] If landing on `5.0.0`: v4-to-v5 migration notes (linked from the [5.0.0 release notes](https://github.com/cqframework/clinical_quality_language/releases/tag/v5.0.0)) reviewed for anything affecting our engine-level integration beyond the ELM-shape changes found here.

## Related

- #1333 (Stage 1: 3.29.0 -> 4.0.0, this ticket's prerequisite)

baseTwo:
## Primary Work

Upgrades the upstream Java `cql-to-elm-cli`/`elm-fhir` tooling (`org.cqframework`) from `4.0.0` to `5.0.0`, one version at a time (4.1.0 → 4.2.0 → ... → 5.0.0), per this ticket's request for an incremental, verify-at-every-step approach rather than a single jump.

**Each of the 10 commits on this branch bumps the pinned version in the same 4 places** (`Demo/Cql/Build/pom.xml`, `Demo/Cql/Build/Java-Dependencies-Vars.ps1`/`.sh`, `Cql/PackagerCLI/Program.cs`), regenerates ELM (Java) → C# for all 4 library sets (`CoreTests.HL7`, `Demo`, `Measures.Authoring`, `dqm-content-qicore-2025`), and runs `CoreTests` + `Test.Measures.Demo` + the `Firely.Cql.Sdk.Integration.Runner` submodule's `IntegrationRunner` suite before committing.

**No logic in the pipeline tooling itself changed.** `Cql.Compiler`, `CodeGeneration.NET`, and the rest of the C#-generation pipeline are untouched by this PR — the existing code already handles everything the new Java tooling versions produce. The diff is: the 4 version-pin locations, the regenerated `.g.cs`/ELM `.json` output that reflects those version bumps, one regenerated test golden file (see below), a `vnext-release-notes.md` entry, and the `Firely.Cql.Sdk.Integration.Runner` submodule pointer.

### Only 4.6.0 changed anything observable

Versions 4.1.0, 4.2.0, 4.3.0, 4.4.0, 4.5.0, 4.7.0, 4.8.0, 4.9.0, and 5.0.0 were all no-ops: regenerating every library set at each of those versions produced byte-identical output except the embedded `translatorVersion`/`GeneratedCode` version stamp, which was the only diff reverted before each of those commits. This is consistent with each version's own release notes (engine-only fixes, internal refactors explicitly marked behavior-preserving, additive features this pipeline doesn't invoke).

**4.6.0** introduced two real ELM-shape changes:
- **Interval-literal boundary constant-folding** — `Interval[low, high).high` now compiles directly to the `high` expression instead of a `Property`-extraction node reading it back off a constructed `Interval`.
- **Choice-type alphabetical sorting** — ELM `TypeSpecifier` `choice` lists are now serialized alphabetically instead of in source/insertion order.

Both are simplifications/reorderings with no semantic effect, and the existing `Cql.Compiler`/`CodeGeneration.NET` pipeline already produces correct (if textually different) C# for both the old and new ELM shapes.

One of these differences did surface as a test failure that needed fixing: `CSharpGenerationGoldenTests.DqmQiCore2025_CMS56_CSharp_Matches_CheckedInFiles` failed because the choice-type reordering changed the emitted order of a 4-way, mutually-exclusive type-check chain (`CqlDateTime` / `CqlInterval<CqlDateTime>` / `CqlQuantity` / `CqlInterval<CqlQuantity>`, resolving FHIR `Procedure.performed`'s choice type). Before/after:

```diff
 bool n_() {
     DataType t_ = THAProcedure?.Performed;
     object u_ = FHIRHelpers_4_4_000.Instance.ToValue(context, t_);
-    bool v_ = u_ is CqlInterval<CqlDateTime>;
+    bool v_ = u_ is CqlQuantity;
     return v_;
 }
```

Since the four candidate types are mutually exclusive (no runtime value can ever match more than one), swapping which `is` check runs first cannot change which branch fires for any input — this is a safe, behavior-preserving reorder. Verified, then regenerated the checked-in golden file: `LibrarySets/dqm-content-qicore-2025/Extracted/CSharp/CMS56FHIRFuncStatHipReplacement-1.0.000.g.cs`.

### How the Retrieve -> Query / asType-asTypeSpecifier equivalence was verified

Beyond the CMS56 fix above, this ticket's acceptance criteria asked for the `Retrieve` -> `Query` restructuring and `asType`/`asTypeSpecifier` swap to be traced through `Cql.Compiler`'s `CodeBuilder`, not just assumed safe from tests passing. Two levels of verification:

1. **Code-level trace**: `CodeBuilderContext.Refs.cs`'s `Retrieve()` folds a non-default `codeProperty` natively into `RetrieveParameters`; at runtime `BundleDataSource.ExecuteFilter` resolves it via the same `ValueSets.IsCodeInValueSet(...)` lookup used for the default code path. The generic `Query`/`Where` path (`CodeBuilderContext.Query.cs`) compiles *any* boolean predicate — including an `X in "ValueSet"` expression — through the identical `ValueSets.IsCodeInValueSet(...)` call. Both ELM shapes converge on the same membership test. Likewise, `CodeBuilderContext.Operators.cs`'s `As()`/`Is()` already symmetrically resolve both the string form (`asType`/`isType`) and object form (`asTypeSpecifier`/`isTypeSpecifier`) to the identical .NET `Type` and feed the same downstream cast/type-check codegen — this dual support predates this PR.
2. **Empirical A/B test**: found a real in-corpus example (`CMS144FHIRHFBetaBlockerForLVSD` reorders the two operands of a `union` between a valueset-filtered retrieve and a where-filtered retrieve — safe, since `union`/`exists()` are order-independent). Temporarily swapped that file's compiled output back to its pre-upgrade (`develop`) content, rebuilt, and re-ran its full 48-case `IntegrationRunner` suite; then restored the post-upgrade file and re-ran the same suite. **All 48 test cases produced byte-identical pass/fail/skip outcomes in both versions.**

Full write-up in FirelyTeam/firely-cql-sdk#1334.

### Submodule

`submodules/Firely.Cql.Sdk.Integration.Runner`'s packaged Library resources (`IntegrationRunner/Resources/libraries`) were regenerated to match the 4.6.0 output and committed on a matching branch in that repo; the submodule pointer here is bumped to that commit. Companion PR: FirelyTeam/Firely.Cql.Sdk.Integration.Runner#106. Its own `CodeGen` project was re-run to regenerate `IntegrationRunner/Tests/*.g.cs` and test-case bundles at every step — these came back unchanged each time (they depend only on external test-case data, not the translator version), so no test-file changes were needed. The full `IntegrationRunner` suite (2700 tests, real patient-bundle measure evaluations) passed at every step, including 4.6.0.

**Bonus finding — 136 additional passing tests discovered.** `IntegrationRunner`'s `passed.jsonl` is a known-good ledger: when present, the suite only re-verifies the tests already in it (`ReadAndOnlyRunPreviousPasses`); when absent, it falls back to attempting every test case for real and writes out whichever ones now pass (`Write` mode). Running in that discovery mode (2836 passed / 578 failed / 443 skipped, vs. the usual 2700 / 0 / 1157) and diffing against the prior ledger showed **zero regressions** — every one of the original 2700 known-good cases still passes — and **136 newly-discovered passing cases**, concentrated in `CMS1173FHIRDiagnosticDelayVTE` (41), `CMS190FHIRVTEProphylaxisICU` (30), and `CMS108FHIRVTEProphylaxis` (25), among others. The 578 still-failing cases are pre-existing environment/test-data gaps (e.g. "No ValueSets for ..."), unrelated to this upgrade. `passed.jsonl` updated accordingly; the prior ledger archived as `passed.jsonl.2026-07-27` per its existing dated-snapshot convention.

Also fixed a stale `docs/build-pipeline.md` reference that still said the pinned version was "currently `4.0.0`".

Closes #1334.

---

## Auxiliary Work

None.


alexzautke (commented):
All "translatorVersion" should be raised to "5.0.0", no?

baseTwo:
> All "translatorVersion" should be raised to "5.0.0", no?

I only keep the latest version of files where there were actual output changes. Only some elm files were affected by 4.6. No need to update all elm files if only the version changed.

Same for the C# generated files.

alexzautke (approved):

PR #1431, opened 2026-07-27
Add MeasureGroupCodeSystem option to emit Measure.group.code

alexzautke:
## Summary

Adds support for emitting `Measure.group.code` during measure packaging. A new optional `MeasureGroupCodeSystem` configuration setting allows users to specify a code system URL; when set, each measure group's id is also emitted as a coding with that system and the group id as its code.

## Changes

- **Core packaging logic** (`FhirMeasureExtensions.cs`):
  - Added `measureGroupCodeSystem` parameter to `CreateMeasureResource()`, `TryCreateMeasure()`, `AnnotateMeasurePopulations()`, and `AnnotateMeasureStratifiers()`
  - Updated `GetOrCreateGroup()` to accept and apply the code system, setting `Measure.group.code` with a `Coding` containing the system and group id when the parameter is non-null
  - Removed commented-out code that was a placeholder for this feature

- **Configuration** (`PackagingToolkitConfig.cs`):
  - Added `MeasureGroupCodeSystem` property with documentation explaining its behavior

- **CLI commands** (`CqlToFhirCommand.cs`, `ElmToFhirCommand.cs`):
  - Added `--measure-group-code-system` option to both commands with help text
  - Wired the option to `PackagingOptions.MeasureGroupCodeSystem` config section

- **Packaging orchestration** (`PackagingToolkit.cs`, `ResourcePackager.cs`):
  - Threaded `measureGroupCodeSystem` parameter through the packaging pipeline from config to resource creation

- **Tests** (`FhirMeasureExtensionsTests.cs`):
  - Refactored `CreateMeasure()` helper to accept optional `measureGroupCodeSystem` parameter
  - Added three new test cases:
    - `MeasureGroupCodeSystem_SetsGroupCodeWithGroupIdAsCode`: verifies groups get the code system and group id as code
    - `MeasureGroupCodeSystem_AlsoAppliesToGroupsCreatedByStratifiers`: verifies the code system applies to groups created by stratifiers
    - `NoMeasureGroupCodeSystem_LeavesGroupCodeUnset`: verifies default behavior (null) leaves code unset

- **Documentation**:
  - Updated release notes with feature description
  - Updated CLI documentation with the new option

## Implementation Details

- The feature is opt-in: when `measureGroupCodeSystem` is null (the default), no `Measure.group.code` is emitted, preserving backward compatibility
- The code system is applied consistently to all groups, whether created directly from populations or indirectly from stratifiers
- The group id itself becomes the `code` value in the coding, with the provided system URL

https://claude.ai/code/session_01QqhsgX24cc7zwWo6jhPhYr

### DIFF
--- a/Cql/CoreTests/Packaging/FhirMeasureExtensionsTests.cs
+++ b/Cql/CoreTests/Packaging/FhirMeasureExtensionsTests.cs
@@ -316,4 +316,49 @@ public class FhirMeasureExtensionsTests
 
         measure.Group.Should().OnlyContain(g => g.Code == null);
     }
+
+    [TestMethod]
+    [DataRow("Rate\tA", DisplayName = "tab")]
+    [DataRow("Rate  A", DisplayName = "double space")]
+    [DataRow(" RateA", DisplayName = "leading space")]
+    [DataRow("RateA ", DisplayName = "trailing space")]
+    public void MeasureGroupCodeSystem_GroupIdViolatingFhirCodeConstraints_Throws(string groupId)
+    {
+        const string system = "https://example.org/fhir/CodeSystem/measure-group";
+
+        var act = () => CreateMeasure(system, BaseStatements(
+            Def("Region Stratifier",
+                CreateTag("group", groupId),
+                CreateTag("stratifier", "Region"))));
+
+        act.Should().Throw<InvalidOperationException>()
+           .WithMessage($"*'{groupId}'*FHIR code*");
+    }
+
+    [TestMethod]
+    public void MeasureGroupCodeSystem_GroupIdWithSingleInternalSpaces_IsAccepted()
+    {
+        const string system = "https://example.org/fhir/CodeSystem/measure-group";
+
+        var measure = CreateMeasure(system, BaseStatements(
+            Def("Region Stratifier",
+                CreateTag("group", "Rate C"),
+                CreateTag("stratifier", "Region"))));
+
+        var group = measure.Group.Single(g => g.ElementId == "Rate C");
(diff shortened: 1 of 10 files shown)


baseTwo:
@copilot, please fix these

1. GetOrCreateGroup treats a whitespace-only measureGroupCodeSystem the same as null (via string.IsNullOrWhiteSpace), but the PackagingToolkitConfig.MeasureGroupCodeSystem XML doc only documents the null case. A one-line doc addition ("empty/whitespace is also treated as unset") would make the contract fully explicit.
2. Coding.Code = groupId embeds the raw @group annotation value directly as a FHIR code, with no validation against the code datatype's whitespace constraints.

alexzautke:
@baseTwo Both points are now fully addressed in 6ecd5af (rebased on top of Copilot's 30872b8, since we both picked up your comment):

1. The `MeasureGroupCodeSystem` XML doc now states that an empty or whitespace-only value is treated the same as `null`.
2. Instead of silently trimming, group ids are now validated against the full FHIR `code` datatype constraints (non-empty, no leading/trailing whitespace, internal whitespace limited to single spaces) before being emitted as `Coding.code`, and packaging fails with a clear error pointing at the offending `@group` annotation value. I removed the `.Trim()` from 30872b8: the validation makes it a no-op, and trimming silently would have let `Coding.code` diverge from the group's `ElementId` (which stays untrimmed). Validation only applies when a code system is configured — without one no code is emitted, so the constraints don't apply. Covered by new tests in `FhirMeasureExtensionsTests`, and the vnext release note mentions the validation.

---

baseTwo (approved):

PR #1455, opened 2026-07-30
Range to interval conversions for Long and a dispatch helper

### LINKED ISSUE #1454: Range → interval conversions: inbound CqlInterval&lt;long?&gt; registration and a point-type-aware ConvertRangeToCqlInterval helper
## Problem

Two gaps in the `Range` side of the host conversion family introduced by #1408 / the FHIR-56226 work:

1. **Round-trip asymmetry for `Interval<Long>`.** Since 2.11.0 the engine serializes `CqlInterval<long?>` *out* to a unit-less FHIR `Range`, but `FhirTypeConverter` registers no conversion back *in* — while `Interval<Integer>` and `Interval<Decimal>` have both directions (`FhirTypeConverter.cs` registers `Range → CqlInterval<CqlQuantity>`, `→ CqlInterval<decimal?>`, `→ CqlInterval<int?>`). The engine can emit a shape it cannot read, and every host accepting Range-encoded `Long` interval parameters must hand-roll the bounds reading (Firely Server just did, in `FhirValueToCqlValueConverter`).

2. **No point-type-aware dispatch helper for `Range`.** 2.11.0 added `ConvertPeriodToCqlInterval(this TypeConverter, Period?, string? cqlPointTypeName)` because the CQL point type of a `Period` is knowledge only the host has (e.g. from the cqf-cqlType extension) while the dispatch belongs to the engine. A `Range` has the identical ambiguity with **four** outcomes instead of two: `Interval<Quantity>`, or an interval of `Integer`, `Decimal` or `Long` serialized as a Range of unit-less quantities. Without an engine-side helper, every host re-implements the dispatch and re-decides the accepted point-type spellings and the fallback.

## Proposal

- Register the inbound `Range → CqlInterval<long?>` conversion alongside the existing three, with defined semantics for fractional bound values consistent with the `CqlInterval<int?>` path, and closed bounds (FHIR Range bounds are always inclusive).
- Add `ConvertRangeToCqlInterval(this TypeConverter, Range?, string? cqlPointTypeName)` next to `ConvertPeriodToCqlInterval`, dispatching `Integer`/`Decimal`/`Long` to the typed interval conversions and anything else — including an absent hint — to the widest reading, `CqlInterval<CqlQuantity>`, mirroring the Period helper's naming convention (`nameof(CqlPrimitiveType.…)`) and null pass-through.

This lets hosts replace their hand-rolled `Range` dispatch with one call, centralizes the hint-spelling convention with the Period helper, and closes the emit-but-can't-read asymmetry. Firely Server will adopt it in a follow-up once released (VONK-10011 shipped the host-side dispatch in the meantime).

alexzautke:
## Description

The engine serializes a CQL `Interval<Long>` *out* to a unit-less FHIR `Range` (FHIR-56226) but registered no conversion back *in* — hosts accepting Range-encoded `Long` interval parameters had to read the bounds by hand. And unlike `Period` (which got `ConvertPeriodToCqlInterval` in 2.11.0), `Range` had no point-type-aware dispatch helper, even though its mapping is ambiguous four ways (`Interval<Quantity>` vs intervals of `Integer`/`Decimal`/`Long`).

This PR:

- **Registers the inbound `Range → CqlInterval<long?>` conversion** alongside its three neighbors, plus the previously missing `Quantity → long?` conversion it builds on — with semantics traced from and identical to the `int?` path: fractional bound values truncate towards zero (the C# `decimal → long` conversion), out-of-range values throw `OverflowException`, absent bounds and value-less quantities pass through as `null`, and bounds are closed (FHIR `Range` bounds are always inclusive).
- **Adds `ConvertRangeToCqlInterval(this TypeConverter, Range?, string? cqlPointTypeName)`** directly next to `ConvertPeriodToCqlInterval`, mirroring its shape: `Integer`/`Decimal`/`Long` (per `nameof(CqlPrimitiveType.…)`) yield the typed interval, anything else — including an absent hint — the widest reading, `Interval<Quantity>`. A null range yields null via the converter's existing null pass-through, same as the Period helper.

One public API addition (`ConvertRangeToCqlInterval`), declared in `PublicAPI.Unshipped.txt`; the registered conversions are internal wiring.

## Related issues

Closes #1454.

## Testing

- 20 new tests in `FhirTypeConverterHostConversionTests`: the inbound `long?` conversion (whole and beyond-int values, absent/value-less bounds, closed bounds, truncation, overflow), full round trips `Interval<Long> → Range → Interval<Long>` (including a `Range` carrying the `quantity-precision` extensions the outbound side emits, and open-bound → closed-equivalent), and all dispatch outcomes of the helper (three typed hints, null/empty/unrecognized hint → `Interval<Quantity>`, null range → null).
- Full `CoreTests` suite (net10.0): 998 passed, 0 failed, 8 pre-existing skips.
- `Cql.Firely` Release build clean — PublicAPI analyzer passes with the new entry.

## FirelyTeam Checklist
- [ ] Mark the PR with the label **breaking change** when this PR introduces breaking changes (not applicable — additive)

https://claude.ai/code/session_019LxWQZxuUUup1ZULebJnuK

---


baseTwo:
@copilot This PR currently has merge conflicts against develop (mergeStateStatus: DIRTY). Please rebase (or merge develop in, whichever keeps history cleanest per this repo's convention) and resolve the conflicts before I continue reviewing.

baseTwo:
@copilot Reviewed this thoroughly (correctness/spec-conformance and test-coverage/API lenses, both independently verified against the actual branch code). The core conversion logic is solid — decimal→long truncation/overflow semantics correctly mirror the Integer path (verified at both boundaries empirically), the FHIR `Range` resource's bounds really are always inclusive per the R4/R5 StructureDefinitions (so hardcoding `lowClosed: true, highClosed: true` is correct, not just asserted), and the "Quantity" fallback case is actually the spec-correct default per the FHIR-CQL conformance IG. Two required fixes before merge:

1. **`ConvertRangeToCqlInterval`'s dispatch never actually matches this SDK's own packager output.** `Cql/Cql.Packaging/FhirLibraryExtensions.cs`'s `AddParameterCqlTypeExtension` writes the `cqf-cqlType` extension as `"Interval<Long>"`/`"Interval<Integer>"`/`"Interval<Decimal>"` (confirmed format, and confirmed in real corpus data under the Integration Runner submodule that e.g. `Interval<DateTime>` really is stored that way) — never the bare `"Long"`/`"Integer"`/`"Decimal"` the switch in `ConvertRangeToCqlInterval` compares against via `nameof(CqlPrimitiveType.X)`. So a host that reads the extension as documented ("typically taken from the cqf-cqlType extension on a Library parameter") and passes it straight through gets silently routed to the `Interval<Quantity>` fallback every time, regardless of the parameter's real declared point type — no exception, just silently wrong. Please fix the dispatch to correctly handle the real `"Interval<X>"` wrapped format (e.g. strip the `Interval<...>` wrapper before matching), correct the docstring/release-note wording to describe the actual expected input, and add a regression test using the real wrapped format (e.g. `"Interval<Long>"`) to prove it dispatches correctly. Note the identical bug already exists in the pre-existing, shipped `ConvertPeriodToCqlInterval` (same root cause, same mismatch for `"Date"`/`"DateTime"`) — since it's the same one-line fix, please apply it there too if that's easy to do safely alongside this PR; otherwise let me know and I'll file a separate tracked issue for the `Period` side so it isn't lost.
2. **Missing negative-overflow test.** `Convert_Range_CqlLongInterval_ThrowsForBoundOutsideLongRange` only tests a value above `long.MaxValue` (`20000000000000000000m`). Please add a mirrored test for a value below `long.MinValue` (e.g. `-20000000000000000000m`), since the underlying `(long?)decimal` cast is documented to throw symmetrically in both directions and that's currently unverified.

Please push both as a follow-up commit.

baseTwo (approved):
Both required fixes verified against the actual branch code:

1. **Wrapped `cqf-cqlType` format now handled correctly.** `ConvertRangeToCqlInterval` and `ConvertPeriodToCqlInterval` both route through a new shared `GetCqlIntervalPointTypeName` helper that strips the `Interval<...>` wrapper this SDK's own packager actually emits, while still accepting the bare point-type name for backward compatibility. Covered by new `ConvertRangeToCqlInterval_WrappedLongHint_ReturnsIntervalOfLong` (`"Interval<Long>"`) and a matching Period-side test (`"Interval<Date>"`). Docstrings and the release-note fragment updated to describe the real input format. Traced the edge cases (empty string, unrecognized hint) through the new helper — all still fall through to the correct default.
2. **Negative-overflow test added**: `Convert_Range_CqlLongInterval_ThrowsForBoundBelowLongRange` mirrors the existing positive-overflow test.

Ran the full `FhirTypeConverter*` test suite in an isolated worktree on both net8.0 and net10.0: 96/96 pass on each (8 pre-existing, unrelated skips). Approving.

Issue #132, opened 2023-12-21
Implement uncertainty


EvanMachusak:
We don't implement uncertainty in the same way as the other two open source implementations, and this will cause testing differences that would be otherwise unresolvable.

We need to replace the use of `int?`, `long?`, and `decimal?` as the default bound types for the ELM `System.Integer`, `System.Long`, and `System.Decimal` types respectively to CQL types like `CqlInteger`, `CqlLong`, and `CqlDecimal`.  These new structures need to contain their primitive value, and must also contain a property that can contain the uncertainty range.  When the uncertainty range is needed and is populated, the result value is considered *uncertain*.  Uncertainty will need to be taken into account in a myriad of operators, mainly boolean comparison operators.

These types should be implicitly convertible `from` their respective primitive types.  It is not clear whether they should be implicitly convertible `to` their primitive types, since when they are uncertain, we would be collapsing a range down to a scalar value.  It may be reasonable to return `null` in such cases, or throw a runtime exception.  It may also be preferable to make the cast possible but explicit.

We will need comparers and type converters implemented for these new types.

The signatures of all of the `ICqlOperators` interface will need to have their primitive return types changed to use these structures so that uncertainty can be propagated.

This is a significant change and will result in a large pull request.

ewoutkramer:
BTW: this will be added to the language as an official Uncertain<T> type, so I propose we do not act until that's official.

alexzautke:
# Proposal: scoped uncertainty type (revised approach)

This proposal supersedes the original sketch above in one key respect: rather than replacing `int?`/`long?`/`decimal?` wholesale across `ICqlOperators`, it introduces a dedicated uncertainty type threaded **statically, only where uncertainty can actually flow**. The original proposal predates the typed-IR pipeline; now that every expression has a concrete static C# type end-to-end, scoped static propagation is both feasible and much cheaper than a wholesale primitive replacement.

## 1. Grounding

Per spec §5 (Language Semantics, "Uncertainty"): *"the concept of uncertainty is not exposed directly in CQL or in ELM, but is defined as an implementation detail."* An uncertainty is a closed interval with precisely specified comparison semantics — e.g. equality is defined as:

```
if A.low <= B.high and A.high >= B.low
  then if A.low = A.high and B.low = B.high then true else null
  else false
```

and relational comparison analogously (`A.high < B.low → true; A.low >= B.high → false; else null`), including the boundary case where overlapping-by-one still yields a decidable answer.

[...]

PR #1459, opened 2026-07-31
Fix interval Intersect with null boundaries

### LINKED ISSUE #1457: Interval intersect returns null when a boundary is null instead of applying CQL null-boundary semantics
## Summary

The runtime interval `intersect` operator (`CqlOperators.Intersect<T>`) returns `null` whenever **any** of the four boundary values of the two operands is `null`, regardless of whether the boundary is open or closed. Per the CQL spec, a `null` boundary is not a null-propagating value: a null **closed** boundary means the minimum/maximum of the point type, and a null **open** boundary is an *uncertainty* that only nulls out computations that actually depend on it.

## Impact

Any `intersect` over intervals with an unbounded/unknown side silently evaluates to `null` instead of the spec-mandated interval. This affects real-world measure logic, e.g. intersecting a measurement period with an ongoing (open-ended) medication or encounter period.

## Observed Behavior

```cql
Interval[1, 10] intersect Interval[5, null)   // returns null
Interval[1, 10] intersect Interval[5, null]   // returns null
```

## Expected Behavior

The CQL spec (Language Semantics, "Interval Operators") uses the first expression as its worked example:

> Note that open null boundaries of intervals are treated as uncertainties for the purposes of interval computation. For example: `Interval[1, 10] intersect Interval[5, null)` — This results in an interval that begins at 5, and ends at some value between 5 and 10.

(Local mirror: `spec/cql/condensed/05-languagesemantics.md:994-1006`; null-boundary semantics in `spec/cql/condensed/04-logicalspecification.md:2771-2775` and `spec/cql/condensed/02-authorsguide.md:1845-1859`.)

So:

- `Interval[1, 10] intersect Interval[5, null)` → `Interval[5, null)` (start is certain; end is unknown, representable only as a null open boundary)
- `Interval[1, 10] intersect Interval[5, null]` → `Interval[5, 10]` (a closed null high boundary *is* the maximum of the point type, so the result is fully determinate)

## Root Cause

`Cql/Cql.Runtime/Operators/CqlOperators.IntervalOperators.cs:1072-1080` — the opening guard of `Intersect<T>`:

```csharp
if (left == null
   || right == null
   || left.low == null
   || left.high == null
   || right.low == null
   || right.high == null)
    return null;
```

Only the two operand null checks are correct ("If either argument is null, the result is null"). The four boundary-value checks short-circuit before any intersection logic runs and ignore the open/closed flags entirely.

Supporting evidence that this guard is the defect rather than a design choice:

- Lines 1082-1085 fall back with `left.low ?? MinValue<T>()`, `left.high ?? MaxValue<T>()`, etc. — normalization code that is unreachable dead code because the guard already rejected every null boundary.
- `CoreTests/PrimitiveTests.cs:3394` (`TestIntersectNull`) asserts `null` for exactly the spec's worked example, enshrining the incorrect behavior.

## Repro Steps

1. Run this unit test (it currently passes, codifying the bug):

```csharp
var lhs = new CqlInterval<int?>(1, 10, true, true);
var rhs = new CqlInterval<int?>(5, null, true, false);
var result = GetNewContext().Operators.Intersect(lhs, rhs);
Assert.IsNull(result); // passes today; per spec the result should be Interval[5, null)
```

2. Or evaluate `define "X": Interval[1, 10] intersect Interval[5, null)` through the engine and observe `X = null`.

## Acceptance Criteria

- [ ] `Intersect<T>` no longer null-short-circuits on boundary values; only null *operands* return null.
- [ ] Null **closed** boundaries are treated as `MinValue<T>()`/`MaxValue<T>()` and flow through the max-of-lows/min-of-highs logic (`Interval[1, 10] intersect Interval[5, null]` → `Interval[5, 10]`).
- [ ] Null **open** boundaries are treated as uncertainties: a result boundary decided by comparison against an unknown boundary is emitted as a null open boundary (`Interval[1, 10] intersect Interval[5, null)` → `Interval[5, null)`), mirroring the `IsUnknownBoundary`/boundary-range handling already used by `IntervalIncludesInterval` (same file, lines 1043-1050).
- [ ] The overlap pre-check (line 1086) uses the uncertainty range when comparing against unknown boundaries instead of comparing raw values.
- [ ] `TestIntersectNull` is updated to the spec-correct expectation, and tests are added for the closed-null and open-null variants on both low and high boundaries.
- [ ] No `ICqlOperators` signature change, so no `GeneratorToolVersion` bump or `*.g.cs` regeneration is required.

---

> _**Acceptance criteria un-ticked on 2026-08-11.** They were ticked against PR #1459, which was **closed unmerged** ("Closing until #132 is done"), so none of them actually landed — the defect still reproduces on `develop`. Ticked boxes on an unmerged branch make the backlog read as done when it is not. See the triage comment below._

alexzautke:
## Description
`Intersect<T>` for intervals short-circuited to `null` whenever **any** of the four boundary values was `null`, ignoring the open/closed flag. Per the CQL spec, a null closed boundary is the minimum/maximum of the point type, and a null open boundary is an *uncertainty* that only nulls out comparisons that depend on it (spec 05, "Interval Operators" — which uses `Interval[1, 10] intersect Interval[5, null)` as its worked example).

`Intersect` now reuses the existing `IsUnknownBoundary`/boundary-range helpers already used by `IntervalIncludesInterval` and `OverlapsHelper`:

- The overlap pre-check compares uncertainty ranges when an unknown boundary is involved; the result is `null` when the intervals are disjoint **or** when the overlap cannot be established because of an unknown boundary.
- The max-of-lows/min-of-highs boundary selection likewise compares ranges; an indeterminate comparison emits a null **open** result boundary (unknown), while determinate comparisons pick the known boundary and its closed flag as before.
- Null **closed** boundaries flow through the pre-existing `?? MinValue<T>()`/`?? MaxValue<T>()` normalization (previously dead code behind the removed guard).

Behavior for intervals with all-known boundaries is unchanged, and there is no `ICqlOperators` signature change (no `GeneratorToolVersion` bump needed).

Examples:

| Expression | Before | After |
|---|---|---|
| `Interval[1, 10] intersect Interval[5, null)` | `null` | `Interval[5, null)` |
| `Interval[1, 10] intersect Interval[5, null]` | `null` | `Interval[5, 10]` |
| `Interval(null, 10] intersect Interval[5, 20]` | `null` | `Interval(null, 10]` |
| `Interval[1, null) intersect Interval[5, 10]` | `null` | `null` (overlap unknowable) |

## Related issues
Fixes #1457.

## Testing
- Replaced `TestIntersectNull` (which asserted the buggy null result for the spec's worked example) with seven tests in `CoreTests/PrimitiveTests.cs` covering: open-null high (spec example), closed-null high, open-null low, closed-null low, indeterminate overlap → null, disjoint with unknown boundary → null, and certain overlap with unknown high.
- Full `CoreTests` suite: 1061 passed, 0 failed (net10.0).
- Full integration test suite (`Firely.Cql.Sdk.Integration.Runner`, CMS/HEDIS corpus, 3857 cases): **no regressions** — every case in the committed `passed.jsonl` baseline (2851) still passes with this branch. The run showed 3 additional passes vs. that baseline (all `CMS156FHIRHighRiskMedsElderly`), but an A/B run against `develop` without this fix produces the identical pass set, so those flips come from changes already on `develop` since the baseline was last regenerated (2026-07-29), not from this PR.

## FirelyTeam Checklist

https://claude.ai/code/session_018jcpygZCSctt5u8CffsYxd

### DIFF
--- a/Cql/CoreTests/PrimitiveTests.cs
+++ b/Cql/CoreTests/PrimitiveTests.cs
@@ -3390,14 +3390,108 @@ namespace CoreTests
             var result = ops.IntervalIncludesInterval(lhs, rhs, null);
             Assert.IsNull(result);
         }
+        // Interval[1, 10] intersect Interval[5, null) returns Interval[5, null):
+        // the start is certain, the end is unknown (spec 05, Interval Operators).
         [TestMethod]
-        public void TestIntersectNull()
+        public void TestIntersectOpenNullHighBoundary()
         {
             var lhs = new CqlInterval<int?>(1, 10, true, true);
             var rhs = new CqlInterval<int?>(5, null, true, false);
             var ops = GetNewContext().Operators;
             var result = ops.Intersect(lhs, rhs);
+            Assert.IsNotNull(result);
+            Assert.AreEqual(5, result.low);
+            Assert.IsTrue(result.lowClosed ?? false);
+            Assert.IsNull(result.high);
+            Assert.IsFalse(result.highClosed ?? false);
+        }
+
+        // Interval[1, 10] intersect Interval[5, null] returns Interval[5, 10]:
+        // a null closed boundary is the maximum value of the point type.
+        [TestMethod]
+        public void TestIntersectClosedNullHighBoundary()
+        {
+            var lhs = new CqlInterval<int?>(1, 10, true, true);
+            var rhs = new CqlInterval<int?>(5, null, true, true);
+            var ops = GetNewContext().Operators;
+            var result = ops.Intersect(lhs, rhs);
+            Assert.IsNotNull(result);
+            Assert.AreEqual(5, result.low);
+            Assert.IsTrue(result.lowClosed ?? false);
+            Assert.AreEqual(10, result.high);
+            Assert.IsTrue(result.highClosed ?? false);
+        }
+
+        // Interval(null, 10] intersect Interval[5, 20] returns Interval(null, 10]:
+        // the end is certain, the start is unknown.
+        [TestMethod]
+        public void TestIntersectOpenNullLowBoundary()
+        {
+            var lhs = new CqlInterval<int?>(null, 10, false, true);
+            var rhs = new CqlInterval<int?>(5, 20, true, true);
+            var ops = GetNewContext().Operators;
+            var result = ops.Intersect(lhs, rhs);
+            Assert.IsNotNull(result);
+            Assert.IsNull(result.low);
+            Assert.IsFalse(result.lowClosed ?? false);
+            Assert.AreEqual(10, result.high);
+            Assert.IsTrue(result.highClosed ?? false);
+        }
+
+        // Interval[null, 10] intersect Interval[5, 20] returns Interval[5, 10]:
+        // a null closed boundary is the minimum value of the point type.
+        [TestMethod]
+        public void TestIntersectClosedNullLowBoundary()
+        {
+            var lhs = new CqlInterval<int?>(null, 10, true, true);
+            var rhs = new CqlInterval<int?>(5, 20, true, true);
+            var ops = GetNewContext().Operators;
+            var result = ops.Intersect(lhs, rhs);
+            Assert.IsNotNull(result);
+            Assert.AreEqual(5, result.low);
+            Assert.IsTrue(result.lowClosed ?? false);
+            Assert.AreEqual(10, result.high);
+            Assert.IsTrue(result.highClosed ?? false);
+        }
+
+        // Interval[1, null) intersect Interval[5, 10] returns null: whether the
+        // intervals overlap at all depends on the unknown boundary.
+        [TestMethod]
+        public void TestIntersectIndeterminateOverlapIsNull()
+        {
+            var lhs = new CqlInterval<int?>(1, null, true, false);
+            var rhs = new CqlInterval<int?>(5, 10, true, true);
+            var ops = GetNewContext().Operators;
+            var result = ops.Intersect(lhs, rhs);
             Assert.IsNull(result);
+        }
+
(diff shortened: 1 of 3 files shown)


baseTwo:
Reviewed this against the code and the local spec mirror — the diagnosis is correct and
well-grounded. Confirming the claims, then three things worth folding in.

**Verified:** the guard at `CqlOperators.IntervalOperators.cs:1074-1080` is as quoted; the
`?? MinValue<T>()`/`?? MaxValue<T>()` normalization at 1082-1085 is genuinely unreachable
dead code; `IntervalIncludesInterval` (1035-1050) really does already implement correct
`IsUnknownBoundary` uncertainty handling; `TestIntersectNull` (`PrimitiveTests.cs:3393-3400`)
is verbatim the spec's worked example asserting `IsNull`. Spec citations all check out.

**Extra supporting evidence:** sibling `Union<T>` guards on `left == null || right == null`
only — no boundary checks — so `Intersect` is the outlier, not house style. And
`IntervalUnionHelper` (2193-2231) handles null boundaries *deliberately* correctly, calling
`toClosed()` up front then normalizing `low ?? MinValue<T>()` / absent-high-as-maximum. That's
the spec's own recommended strategy (05-languagesemantics.md:1003-1004), so there's no
parallel bug there.

No `ICqlOperators` signature change, so the "no `GeneratorToolVersion` bump" claim holds.

baseTwo:
Posted 3 gaps that need to be addressed first https://github.com/FirelyTeam/firely-cql-sdk/issues/1457#issuecomment-5142937735

alexzautke:
Closing until https://github.com/FirelyTeam/firely-cql-sdk/issues/132 is done.

Issue #1460, opened 2026-07-31
Three process-lifetime retention leaks: ELM ObjectIDGenerator, undisposed toolkit ServiceProviders, stale weak-cache entries


baseTwo:
## Summary

Three independent process-lifetime retention leaks found while auditing assembly-load-context reclaimability for [#1440](https://github.com/FirelyTeam/firely-cql-sdk/issues/1440). None of them roots a collectible `AssemblyLoadContext` (that audit came back clean — see the comment on #1440), but each retains memory for the life of the process and they are cheap, independent fixes. Grouping them because they are all the same class of defect: something long-lived holding references it never releases.

## 1. `Cql.Elm.Initializers._idGenerator` retains every ELM element ever created

[`Cql/Cql.Elm/Initializers.cs:13-22`](../blob/develop/Cql/Elm/Initializers.cs)

```csharp
internal static class Initializers
{
#pragma warning disable SYSLIB0050 // ObjectIDGenerator is obsolete!
    private static readonly ObjectIDGenerator _idGenerator = new();
#pragma warning restore SYSLIB0050

    public static string NextId(object context) => _idGenerator.GetId(context, out _)
            .ToString(CultureInfo.InvariantCulture);

    public static string NextId() => NextId(Random.Shared.Next());

    public static T WithId<T>(this T t) where T : Element
    {
        t.localId = NextId(t);
        return t;
    }
```

`ObjectIDGenerator` keeps a **strong** reference to every object it is asked to identify — it is designed to live for the duration of one serialization graph, which is why it is obsoleted as `SYSLIB0050` (the warning is suppressed here). Because the instance is `static readonly`, every `Element` passed through `WithId()` / `WithLocator()` is retained for the life of the process, as is a boxed `int` for every parameterless `NextId()` call.

There are ~211 call sites, effectively all in `Cql.CqlToElm`, so the exposure is scoped to processes that translate CQL → ELM at runtime (the Packager CLI is short-lived and largely unaffected; a long-running host that translates is not). This is the largest raw-memory retention found.

The `localId` values only need to be unique within a library, so a process-wide identity map is more than the requirement needs.

## 2. No toolkit disposes its `ServiceProvider` — and one of them has an unused `Dispose()`

[...]

PR #1461, opened 2026-07-31
Pool LibrarySetInvoker instances to stop ALC accumulation

### LINKED ISSUE #1440: Pool/reuse LibrarySetInvoker instances instead of creating a fresh AssemblyLoadContext per CreateLibrarySetInvoker call
## Problem

`InvocationToolkit.CreateLibrarySetInvoker` (`Cql/Cql.Invocation/Toolkit/InvocationToolkit.cs`) unconditionally creates a new collectible `AssemblyLoadContext` and loads every `AssemblyBinary` from bytes on each call, and `LibrarySetInvoker.Dispose()` unconditionally calls `alc.Unload()`.

For a consumer that evaluates the same library set repeatedly — e.g. Firely Server executing `Measure/$evaluate-measure` over a patient population, which runs `Library/$evaluate` once per subject × group — this means the entire library set (for HEDIS: the large generated measure assembly plus its full dependency closure) is re-loaded and re-JIT-compiled for every single evaluation:

- Byte-array loads share nothing between contexts: no file-backed image mapping, no JIT code reuse across ALCs.
- Collectible ALC unloading is asynchronous (completes only after GC proves the context unreachable), so under sustained parallel evaluation live ALCs accumulate faster than they are collected.
- With W^X enabled (default on .NET 8 Linux), each JIT'd code region costs two memory mappings (RW + RX views). The process eventually hits the kernel's `vm.max_map_count` limit (default 65530); the next `mmap` fails and the runtime hard-aborts, presenting as a segfault.

Consumers can work around this today by caching `LibrarySetInvoker` instances themselves (the generated library classes are stateless singletons and `LibraryInvokers`/`Definitions` are immutable, so sharing one invoker across concurrent evaluations is safe as long as each evaluation gets its own `CqlContext`). But the SDK owns the ALC lifecycle, so it is the natural long-term home for deduplication.

## Proposed enhancement

Add invoker pooling inside the SDK:

1. **Pool keyed by binary content.** Inputs are already modeled as an `AssemblyBinaryHashSet`; key live ALCs by the content hash of that set. On a hit, `CreateLibrarySetInvoker` returns a new invoker handle sharing the already-loaded ALC instead of loading again.
2. **Decouple handle lifetime from ALC ownership.** `Dispose()` on a pooled handle decrements a ref-count rather than calling `alc.Unload()` directly; the ALC unloads when the last handle is released and the pool evicts the entry (bounded, e.g. LRU with a configurable size).
3. **Share the constructed invoker graph, not just the ALC.** Because the generated library classes are stateless singletons and the invoker dictionaries are immutable, the pool can hand out the fully-built `LibrarySetInvoker` — skipping the `GetTypes()` walk and delegate creation on every hit, in addition to the assembly load and JIT.
4. **Optional deeper optimization:** once the SDK owns pooling, it could persist assemblies to disk and use file-backed loads so the OS can share image mappings between contexts — something byte-array loads can never do.

Backwards compatibility: the current create-per-call behavior could remain the default, with pooling opt-in via the toolkit builder (or a new `LibrarySetInvokerPool` type), so existing consumers that rely on immediate unload semantics are unaffected.

## Context

Observed in production as segfaults of Firely Server under parallel HEDIS measure evaluation. Firely Server will ship a consumer-side cache built on the existing public API as the immediate fix; this issue tracks moving that capability into the SDK so all consumers benefit.

baseTwo:
## Primary Work

Adds an opt-in `LibrarySetInvokerPool` so repeated evaluations of the same library set reuse one loaded `AssemblyLoadContext` and its already-built invoker graph, instead of reloading and re-JIT-compiling the whole assembly closure per evaluation.

### Why

`InvocationToolkit.CreateLibrarySetInvoker` always creates a fresh collectible `AssemblyLoadContext` and reloads every `AssemblyBinary`. Collectible contexts never use ReadyToRun, so every generated method is JIT-compiled per context, and unloading is cooperative/asynchronous, so contexts accumulate faster than the GC reclaims them. With W^X (default on .NET 8 Linux) each JIT-compiled region costs two memory mappings, so a host evaluating one library set once per subject eventually exceeds `vm.max_map_count` and the next `mmap` hard-aborts the process.

### New public API

| Type | Purpose |
| --- | --- |
| `LibrarySetInvokerPool` | `GetOrCreate(InvocationToolkit, string librarySetName = "")` returns a pooled invoker; `IDisposable`; LRU eviction |
| `LibrarySetInvokerPoolOptions` | `Capacity` (soft target), `MaxPendingUnloads` (leak warning threshold) |
| `LibrarySetInvokerPoolStatistics` | `Hits`, `Misses`, `Entries`, `PendingUnloads` |

`CreateLibrarySetInvoker` is untouched and still creates per call, so nothing changes unless a consumer opts in. `GetOrCreate` returns the real `LibrarySetInvoker`, so all existing extension methods (`InvokeLibraryDefinition`, `SelectExpressionsForLibrary`, …) work verbatim.

### Content-based keying

Entries key on a sorted SHA-256 hash over each binary's assembly **and** debug-symbol bytes (length-prefixed so concatenation is unambiguous), plus the library set name and exception-continuation policy. Sorting makes it order-independent, which it must be since the input is a set; XOR would collide on permutations.

This is necessary because `AssemblyBinary` is a positional record over `byte[]`, so its synthesized equality falls through to reference identity — and `ElmToolkitInvocationExtensions.CreateInvocationToolkit()` allocates fresh arrays on every call, so two logically identical inputs never compare equal. The hash is derived at the cache boundary rather than by changing `AssemblyBinary`'s shipped public equality, following the approach `CompiledPropertyAccessor` already takes for `PropertyInfo`.

Debug symbols are deliberately included: excluding them would let two entries with identical assemblies but different symbols share one invoker and silently serve whichever symbols loaded first, observable in stack traces. Including them can only cost a redundant miss.

The key and the load both come from **one snapshot** of the toolkit's inputs. `GetOrCreate` reads `AssemblyBinaries` and `BatchProcessExceptionContinuation` once and passes the same values to `LibrarySetInvokerPoolKey.Create` and to `CreateLibrarySetInvoker`. The entry's `Lazy` is forced outside the pool lock, so re-reading the toolkit at that point would let a concurrent `AddAssemblyBinaries` file an entry under a key that no longer described what had been loaded. `AssemblyBinaryReadOnlyHashSet` is already immutable, so the snapshot costs nothing. The derived hash is memoized per set instance, so a pool hit does not re-hash the binaries.

### No reference counting — and why that is safe

Eviction calls `Unload()` without tracking active users. That is safe because unloading a collectible context is *cooperative*: `Unload()` only initiates it, and completion is gated on the GC proving no thread has a frame from those assemblies and no strong references remain. Forceful teardown is an explicit **non-goal** of the runtime feature. Verified against the [`Unload()` API reference](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext.unload), the [unloadability guidance](https://learn.microsoft.com/en-us/dotnet/standard/assembly/unloadability), the [runtime design doc](https://github.com/dotnet/runtime/blob/main/docs/design/features/unloadability.md), and runtime source (`_state` gates only the three `LoadFrom*` methods — nothing gates the JIT, type loader, delegate invocation or reflection).

A ref-counted lease was considered and rejected: without per-lease identity a consumer's accidental double-release would silently over-decrement and unload a context under a live evaluation, reintroducing the crash class being fixed.

### What eviction *does* have to do

`Unload()` reclaims nothing while the `LibrarySetInvoker` is reachable, and that object roots the context two ways: transitively through every generated type, singleton and bound delegate in its invoker graph, and directly through its own `AssemblyLoadContext` field. So `Unload()` now releases **both** — it clears the graph and nulls the field — and runs in a `[MethodImpl(MethodImplOptions.NoInlining)]` helper so JIT-introduced stack slots don't pin either. It returns the context it unloaded, so the pool can register its `WeakReference` without reading the property back and putting a second strong reference in that frame.

This does not make every retained object harmless: a consumer holding a `LibraryInvoker` or `DefinitionInvoker` still roots the context through the generated library instance and bound delegate it holds directly, which is why the docs say not to retain those past the library set.

Concurrency: entries hold a `Lazy<LibrarySetInvoker>` with `ExecutionAndPublication`, so concurrent misses on one key load once — a tolerate-and-discard race would leak the losing context, since nothing would ever unload it. Loading happens outside the pool lock. Entries still loading are never evicted, which is why `Capacity` is documented as a soft target.

Telemetry deliberately does **not** use `AssemblyLoadContext.All`: `InitiateUnload` removes a context from that collection, so precisely the leaked contexts would be invisible. The pool keeps its own `WeakReference` list instead, and prunes reclaimed entries on every eviction — not only when `MaxPendingUnloads` is exceeded, since that threshold defaults to 0 and the only other thing that prunes is the purely diagnostic `Statistics` property, which a consumer may never read.

### Behaviour changes to `LibrarySetInvoker` (recorded as breaking)

- **`LibraryInvokers` now throws `ObjectDisposedException` after unload.** Releasing the graph makes an unloaded invoker unusable; reporting an empty library set instead would turn that into silently wrong results. Code that uses an invoker only within its `using` scope is unaffected.
- **`Dispose()` is now idempotent.** Unloading twice was an access violation before .NET 8.0.0 ([dotnet/runtime#91283](https://github.com/dotnet/runtime/pull/91283)) and pointless after.
- **`Dispose()` and `SetBatchProcessExceptionContinuation` are inert on a pooled instance**, and log that they were ignored. A pooled instance is shared, and the continuation is the only mutable state in its graph, so honouring either would reconfigure — or unload the assemblies out from under — every other consumer of that entry. The continuation is part of the pool key instead, so callers wanting a different policy get their own instance.

### Fixed while reviewing the branch

- **`Dispose()` racing an in-flight `GetOrCreate` leaked a context permanently.** Loading runs outside the pool lock, and `CollectAll()` skips entries whose `Lazy` has not completed — so a dispose landing mid-load dropped the entry while the loading thread went on to receive a finished, pool-owned invoker. Nothing would ever unload it: the pool had forgotten it, and the invoker's own `Dispose()` is inert because `isPoolOwned` is `true`, so the consumer could not unload it either. That made the window strictly worse than not pooling at all. `GetOrCreate` now re-checks `_disposed` after the load, unloads the orphan and throws `ObjectDisposedException` — which is also the documented contract. The opposite interleaving needs no handling: a `Dispose` that takes the lock after the re-check sees `IsValueCreated` true and collects the entry itself.

  Its regression test is deterministic rather than timing-based. The toolkit's logger is called from inside `CreateLibrarySetInvoker` before it creates the context, so a callback `ILoggerFactory` parks the loading thread exactly in the window between entry registration and load completion. Verified it fails with the fix reverted — no exception is thrown, and the context stays listed in `AssemblyLoadContext.All`.
- **`CreateLibrarySetInvoker` leaked its `AssemblyLoadContext` when loading failed.** The context was created up front and only handed to `LibrarySetInvoker` at the very end, so with the default `Throw` continuation an unloadable binary — or the reflection walk over the loaded types — escaped before anything took ownership. Unloading is cooperative and never begins without an `Unload()` call, so the context and whatever had already loaded into it stayed resident for the life of the process. Pooling makes retries after a failure more likely, so it matters more now. The failure path unloads before rethrowing. Its regression test keys off `AssemblyLoadContext.All` rather than a `WeakReference`: initiating an unload removes a context from that collection, so a context still listed there is precisely one whose unload never started — deterministic, and needs no GC.
- **The `LibraryInvokers` getter could return the empty library set its own remarks promise never to serve.** Testing an `_unloadInitiated` flag and then reading the graph is check-then-act no matter how the two reads are ordered or fenced, and the reachable interleaving handed back a dictionary with no libraries — `SelectExpressions()` quietly yielding nothing — which is precisely the silently-wrong-results outcome the throw exists to prevent. The field is now itself the sentinel: it is nullable, `Unload()` sets it to `null`, and the getter is a single atomic reference read that yields either the live graph or the exception. `_unloadInitiated` survives purely as the double-unload guard. Since pool eviction on one thread while others evaluate is a supported scenario, this window was reachable in the intended usage, not just in pathological code.
- **`GetOrCreate_ConcurrentMissesOnOneKey_LoadsTheLibrarySetOnce` asserted equivalence, not identity.** `AllBeEquivalentTo` compares structurally, so it would have passed had the pool created two distinct invokers over the same binaries — the very leak the test exists to catch. It now asserts reference identity and that the observed instances collapse to one under `ReferenceEqualityComparer`.
- **Every `GetOrCreate` re-hashed the binaries, hits included.** SHA-256 over every assembly and pdb byte is the right trade against a load plus JIT, but the intended hot path is one call per subject over an unchanged toolkit, so a hit paid a full hash pass forever. Content hashes are now memoized in a `ConditionalWeakTable` keyed on the binary set, making a repeat call a lookup. Keying on reference identity is sound because `AddAssemblyBinaries` builds a new immutable set and swaps the field, so a given instance's content never changes; the weak table keeps a set the consumer has dropped from pinning its entry for the life of the process. The test pins both halves, since the risk memoizing introduces is regressing the key back to an identity comparison and defeating the point of hashing content.
- **The concurrent-miss test could not detect duplicate loads.** Switching the pool's `Lazy` to `PublicationOnly` — which runs the factory on every racing thread and publishes one winner, leaking an assembly load context per discarded load — left every assertion green, because reference identity, `Distinct()`, `Misses` and `Entries` all still hold. The pool's own `Misses` counter cannot serve as the signal either: it counts threads that found no entry, not assembly loads. The test now counts real loads via a callback `ILoggerFactory` on the toolkit, since `CreateLibrarySetInvoker` logs once per invocation; under `PublicationOnly` it reports 8 instead of 1.
- **The content-hash memo did not help the workload it was added for.** Keyed on the binary *set* instance, it never hits for a consumer that builds a fresh `InvocationToolkit` per call — which is the real hot path (~10,000 `Library/$evaluate` per HEDIS run), so every call still hashed megabytes. It is now keyed on the `byte[]` arrays, which are what a consumer holds stable behind a per-run artifact cache. The two arrays of a binary are hashed and memoized separately rather than streamed into one hash, so the memo stays correct if the same assembly array is ever paired with different symbols; length prefixes are no longer needed, since every component is fixed-length hex or an `absent` marker. The set-level memo remains as a short-circuit for consumers that do reuse a toolkit.

### Pre-implementation audit

Before building this I audited whether anything in the SDK roots a collectible context, since pooling only helps if evicted contexts actually unload. Full write-up in [this comment](https://github.com/FirelyTeam/firely-cql-sdk/issues/1440#issuecomment-5143364931). Verdict: nothing does. A probe running a real end-to-end evaluation showed the context collected in **1 GC round** on net8.0 and net10.0 across 10 scenarios including a 10× churn cycle, with a negative control confirming the probe detects deliberate rooting. Every `Type`-keyed cache candidate came back safe.

Three unrelated process-lifetime retention leaks found during that audit are filed separately as #1460 and are **not** in this PR.

### Testing

27 new tests (29 cases) in `Cql/CoreTests/LibrarySetInvokerPoolTests.cs`:

- content-based reuse across separately built toolkits over equal-but-distinct byte arrays (the case reference equality would fail), and distinct entries for differing content / name / continuation
- `GetOrCreate_ConcurrentMissesOnOneKey_LoadsTheLibrarySetOnce` — 8 barrier-released threads, asserts exactly one load by reference identity
- `SharedInvoker_EvaluatedConcurrentlyFromManyThreads_MatchesSequentialBaseline` — 8 threads × 50 iterations over one shared invoker, each with its own `CqlContext`. **Closes a repo-wide gap: no existing test evaluated a shared `LibrarySetInvoker` concurrently.**
- `Eviction_AfterAllReferencesReleased_UnloadsTheAssemblyLoadContext` — `WeakReference` + GC retry loop; this is the test that proves the leak is actually fixed, and no such test existed before
- `Eviction_WhileAConsumerStillHoldsTheInvoker_StillReleasesTheAssemblyLoadContext` — the only test that detects removal of the graph/field release, since it deliberately keeps the evicted invoker strongly referenced across the GC loop
- `Eviction_EvictsTheLeastRecentlyUsedEntry_NotTheMostRecentlyUsed` — at capacity 2 with an interleaved re-request. Capacity 1 cannot detect a recency bug: walking the LRU list from the wrong end and dropping the recency touch on the hit path look identical with one candidate.
- `Statistics_PendingUnloads_CountsEvictedContextsThatAreStillAlive` and `UnloadEvicted_PrunesReclaimedContexts_WithoutNeedingAStatisticsRead` — the latter asserts against the private list, because reading `Statistics` prunes as a side effect and would mask what is being tested
- `CreateLibrarySetInvoker_WhenLoadingFails_UnloadsTheAssemblyLoadContext`
- eviction-while-held, pooled-`Dispose`-is-inert, continuation-ignored-when-pooled, unpooled double-dispose, capacity validation

Full `CoreTests`: **1084 passed, 0 failed**, 8 pre-existing skips, on both net8.0 and net10.0. `Cql-Sdk.slnf` builds clean, so the PublicAPI analyzer is satisfied.

---

## Auxiliary Work

- **New `InvocationToolkit Services Dependency Diagram`** in [docs/dependency-diagrams.md](docs/dependency-diagrams.md) plus its `.mmd`/`.svg` pair. `InvocationToolkit` was the only toolkit with no diagram — it is also the only one without an internal DI container, so the graph shows construction and ownership rather than container registrations. The back-reference edges are drawn deliberately, since they are the one way a consumer can defeat the pool. Corrected the runtime diagram's "three toolkit diagrams above" wording to name the three DI-container ones explicitly, so it stays accurate with a fourth toolkit section inserted.
- **`Cql.Invocation/README.md`** — new "Reusing a library set across evaluations" section covering the opt-in, the ownership rules, and `PendingUnloads` as the leak signal — including the `Capacity` sizing rule (size it to at least the number of library sets in concurrent use, since eviction-while-held is the one failure mode an adopter can only avoid by configuration).
- **Release-note fragment** [docs/releases/vnext/1461-library-set-invoker-pool.md](docs/releases/vnext/1461-library-set-invoker-pool.md), named per the folder README convention, with Breaking Changes / Features / Fixes / Documentation sections.
- **`CoreTests` added as an `InternalsVisibleTo` friend of `Cql.Invocation`**, so the unload tests can weakly reference the internal `AssemblyLoadContext` property. `CoreTests` was already a friend of nearly every other Cql project. (The pool's private pending-unload list is reached by reflection instead, since it is not merely internal.)

Refs #1440


Cql/Cql.Invocation/Toolkit/InvocationToolkit.cs
```diff
@@ -136,6 +144,7 @@ public LibrarySetInvoker CreateLibrarySetInvoker(string librarySetName = "")
             alc,
             LoggerFactory,
             BatchProcessExceptionContinuation,
-            librarySetName);
+            librarySetName,
+            isPoolOwned);
```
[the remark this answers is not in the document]
baseTwo: Good catch, and confirmed as a real leak — fixed in 958fc137f.

With the default `BatchProcessExceptionContinuation.Throw`, a corrupt binary escapes `TryForEach` (as could the reflection walk in the `LibrarySetInvoker` constructor) leaving a context that no `LibrarySetInvoker` ever took ownership of. Since cooperative unloading never begins without an `Unload()` call, that context plus any assemblies already loaded into it stayed resident for the process lifetime. The failure path now unloads before rethrowing.

Added `CreateLibrarySetInvoker_WhenLoadingFails_UnloadsTheAssemblyLoadContext`. It keys off `AssemblyLoadContext.All` rather than a weak reference, because initiating an unload *removes* a context from that collection — so a context still listed there is precisely one whose unload never started. That is deterministic and needs no GC. I verified it genuinely detects the bug by reverting just the `alc.Unload()` line and watching it fail.

Cql/CoreTests/LibrarySetInvokerPoolTests.cs
```diff
@@ -0,0 +1,468 @@
...
+            threads[i].Start();
+        }
+
+        foreach (var thread in threads)
+            thread.Join();
+
+        // Assert - one load, and everybody got the same instance. A tolerate-and-discard race would
+        // have created a second assembly load context that nothing would ever unload.
+        failures.Should().BeEmpty();
+        observed.Should().AllBeEquivalentTo(observed[0]);
```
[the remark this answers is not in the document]
baseTwo: Agreed — fixed in 552644ea8.

Reference identity is exactly the point of that test: if the pool had created two invokers, nothing would ever unload the losing context, and two invokers over the same binaries could well look structurally equivalent, so `AllBeEquivalentTo` could have passed through the very bug being guarded. Now asserts `ReferenceEquals` against the first observed instance, plus that they collapse to a single entry under `ReferenceEqualityComparer.Instance`.

Cql/Cql.Invocation/Toolkit/LibrarySetInvokerPoolOptions.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// </summary>
+    /// <remarks>
+    /// This is a target rather than a hard ceiling. A library set whose assemblies are still being
+    /// loaded is never evicted, so a burst of concurrent misses on distinct library sets can briefly
+    /// leave more entries in the pool than this. Eviction also only <em>initiates</em> unloading:
+    /// reclaiming the memory needs the garbage collector to run and finalizers to complete, so
+    /// resident memory lags this number.
+    /// </remarks>
+    public int Capacity { get; init; } = Capacity;
+}
```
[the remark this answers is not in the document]
baseTwo: I don't think this one holds — explicitly declaring a property that matches a positional record parameter is legal C#, not a duplicate definition. When you declare the member yourself the compiler uses your declaration instead of synthesizing one, and the `= Capacity` initializer refers to the primary constructor parameter.

Evidence:
- It compiles. `Cql-Sdk.slnf` builds clean in Release on net8.0 and net10.0, which is also where the PublicAPI analyzer runs.
- The value genuinely flows through: `Constructor_CapacityBelowOne_Throws` passes `Capacity: 0 / -1 / int.MinValue` and the pool reads `options.Capacity` to validate, and those cases throw as expected.
- There is an existing precedent in this repo doing exactly the same thing — `ElmToolkitConfig` lines 72 and 80 (`public bool AllowScopeRedefinition { get; init; } = AllowScopeRedefinition;`).

The reason for redeclaring it is to attach the `<remarks>` block explaining that `Capacity` is a soft target rather than a hard ceiling (entries still loading are never evicted, and reclamation lags eviction because unloading is GC-gated). A positional `<param>` doc cannot carry that. Leaving as-is.

Cql/Cql.Invocation/Toolkit/LibrarySetInvokerPool.cs
```diff
@@ -0,0 +1,350 @@
...
+    {
+        var all = _entries.Values
+            .Where(entry => entry.Invoker.IsValueCreated)
+            .Select(entry => entry.Invoker.Value)
+            .ToList();
+
+        _entries.Clear();
+        _leastRecentlyUsed.Clear();
+        return all;
+    }
```
alexzautke, line 302: **`Dispose()` racing an in-flight `GetOrCreate` load leaks the context permanently.**

`CollectAll()` drops entries whose `Lazy` hasn't completed (`IsValueCreated == false`) and clears `_entries`, but nothing ever unloads those entries afterwards:

1. Thread A misses, registers the `Lazy`, releases `_gate`, and starts loading outside the lock.
2. Thread B calls `Dispose()`; `CollectAll()` skips A's still-loading entry and clears the dictionary.
3. A's load completes and `GetOrCreate` returns a pool-owned invoker.

Nothing will ever call `Unload()` on that invoker: the pool has forgotten it, and its own `Dispose()` is inert because it was created with `isPoolOwned: true`. That's a permanent ALC leak of exactly the class this PR fixes — and worse than the pre-pool behaviour, because the consumer holding it has no way to unload it themselves.

A cheap fix: after `lazyInvoker.Value` succeeds in `GetOrCreate`, re-take `_gate` and check `_disposed`; if the pool was disposed in the interim, unload the invoker and throw `ObjectDisposedException`, matching the existing `GetOrCreate_AfterDispose_Throws` contract. A test for `Dispose()` racing a concurrent first-time miss would pin this down.

Narrow window, admittedly — it needs `Dispose()` to overlap a first-time load — so it's harmless for a process-lifetime pool disposed at shutdown, but reachable for pool-per-scope usage.

---

Cql/Cql.Invocation/Toolkit/LibrarySetInvoker.cs
```diff
@@ -57,20 +77,88 @@ internal LibrarySetInvoker(
...
     public IReadOnlyDictionary<CqlVersionedLibraryIdentifier, LibraryInvoker> LibraryInvokers
     {
         [DebuggerStepThrough]
-        get;
+        get
+        {
+            ObjectDisposedException.ThrowIf(_unloadInitiated != 0, this);
+            return _libraryInvokers;
+        }
     }
```
alexzautke, line 199: **Check-then-read race can return the empty dictionary the remarks promise never to serve.**

The getter reads `_unloadInitiated` and then `_libraryInvokers` as two separate steps, while `Unload()` on another thread sets the flag and then swaps the field:

1. Reader checks `_unloadInitiated` → still 0, passes the throw.
2. Evicting thread runs `Unload()`: flag becomes 1, `_libraryInvokers` becomes `Empty`.
3. Reader returns `_libraryInvokers` → the empty dictionary.

The caller then sees a library set containing no libraries — `SelectExpressions()` quietly yields nothing — which is precisely the "silently wrong results" outcome the XML remarks here (and the release note) say this design rejects in favour of throwing. Since pool eviction on one thread while other threads evaluate is a supported scenario, this window is reachable in the intended usage, not just in pathological code.

A race-free shape is to make the field itself the sentinel: declare `_libraryInvokers` nullable, have `Unload()` set it to `null` (keeping `_unloadInitiated` purely as the double-unload guard for `_alc.Unload()`), and make the getter a single read:

```csharp
get => _libraryInvokers ?? throw new ObjectDisposedException(nameof(LibrarySetInvoker));
```

One atomic reference read, no check-then-act gap — a racing reader gets either the live graph or the exception, never the empty dictionary.

---

Cql/Cql.Invocation/Toolkit/LibrarySetInvokerPool.cs
```diff
@@ -0,0 +1,350 @@
...
+    /// <exception cref="ObjectDisposedException">The pool has been disposed.</exception>
+    public LibrarySetInvoker GetOrCreate(InvocationToolkit toolkit, string librarySetName = "")
+    {
+        ArgumentNullException.ThrowIfNull(toolkit);
+        ArgumentNullException.ThrowIfNull(librarySetName);
+
+        var key = LibrarySetInvokerPoolKey.Create(
+            toolkit.AssemblyBinaries,
+            librarySetName,
+            toolkit.BatchProcessExceptionContinuation);
```
alexzautke: **Two notes on key derivation — hit-path cost, and a small TOCTOU.**

1. *Hits still pay a full hash pass.* Every `GetOrCreate` — including hits — runs SHA-256 over every assembly and pdb byte. Against a load + JIT that's obviously the right trade, but the intended hot path is one call per subject, so multi-MB library sets pay a nontrivial per-evaluation cost forever. Since `AssemblyBinaries` is an immutable `AssemblyBinaryHashSet` swapped wholesale by `AddAssemblyBinaries`, the content hash could be memoized per set instance (e.g. a `ConditionalWeakTable<AssemblyBinaryReadOnlyHashSet, string>`), making repeat calls over an unchanged toolkit O(1). If you'd rather not add that now, a `<remarks>` on `GetOrCreate` noting that a hit still hashes the binaries would let consumers make the call.

2. *The key and the load read the toolkit at different times.* The key is computed from `toolkit.AssemblyBinaries` here, but the `Lazy` body re-reads the toolkit's *current* binaries via `toolkit.CreateLibrarySetInvoker(...)`. If another thread mutates the toolkit in between (`AddAssemblyBinaries` swaps the field atomically, so no torn state is needed), the entry is built from content that doesn't match its key — and every future hit on that key silently serves the wrong assemblies, forever. Capturing `var binaries = toolkit.AssemblyBinaries;` once and passing that snapshot through an internal `CreateLibrarySetInvoker` overload would close the gap. Concurrent toolkit mutation is a consumer error, but the failure mode lands in the silent-wrong-results class this PR is otherwise very careful to keep loud.

---

alexzautke (commented):
## Overview

This PR adds an opt-in `LibrarySetInvokerPool` that reuses a loaded `AssemblyLoadContext` and its invoker graph across evaluations of the same library set, fixing the ALC-accumulation / `vm.max_map_count` crash from #1440. Entries are keyed on a content hash (sorted SHA-256 over assembly + pdb bytes, plus name and continuation policy), eviction is LRU, and `LibrarySetInvoker` gains real unload semantics (graph release, idempotent `Dispose`, `ObjectDisposedException` after unload, inert `Dispose`/`SetBatchProcessExceptionContinuation` on pooled instances).

## Assessment

This is a very strong PR. The design reasoning is documented where it matters (why content hashing instead of `AssemblyBinary` equality, why sorted hashes instead of XOR, why debug symbols are part of the identity, why no ref counting, why `Lazy` with `ExecutionAndPublication`, why telemetry avoids `AssemblyLoadContext.All`), the failure-path unwind in `CreateLibrarySetInvoker` is a genuine pre-existing leak fix, and the test suite covers the cases that actually matter — content-based reuse across distinct byte arrays, concurrent single-load, concurrent shared evaluation, and a `WeakReference`-based proof that eviction actually unloads. Repo conventions are all met: copyright headers with the current year, per-file `#nullable enable`, `InternalsVisibleTo` in the `.csproj`, internal types kept internal, `PublicAPI.Unshipped.txt` updated, pre-rendered Mermaid SVG, release-note fragment under `docs/releases/vnext/` with the PR-number naming, and the breaking-change label.

## Findings (inline)

1. **Pool `Dispose()` racing an in-flight load leaks a context permanently** — `CollectAll()` skips still-loading entries but clears the dictionary, so the finished invoker is orphaned with an inert `Dispose()`. This reintroduces the leak class the PR fixes, and the consumer can't recover. (`LibrarySetInvokerPool.cs`)
2. **`LibraryInvokers` check-then-read race** — a reader racing `Unload()` can pass the `ObjectDisposedException` check and still return the cleared (empty) dictionary, producing exactly the silently-wrong-results outcome the design rejects. A nullable field as the sentinel with a single read fixes it. (`LibrarySetInvoker.cs`)
3. **Key derivation** — hits pay a full SHA-256 pass per call (memoizable per immutable binary set), and there's a small TOCTOU between hashing `toolkit.AssemblyBinaries` and the `Lazy` body re-reading the toolkit that can poison a cache entry under concurrent toolkit mutation. (`LibrarySetInvokerPool.cs`)

Items 1 and 2 are narrow-window concurrency bugs rather than mainline defects — the mainline design is sound — but both land in exactly the failure classes (permanent leak, silent wrong results) this PR sets out to eliminate, so they seem worth closing before merge.

## Minor / no action needed

- `Eviction_AfterAllReferencesReleased_UnloadsTheAssemblyLoadContext` depends on GC timing; the 20-round `CollectUntil` loop should be robust on CI, but if it ever flakes there, that's the knob.
- No `GeneratorToolVersion` bump is needed — nothing here changes emitted C# — which matches the PR.

---

baseTwo:
⚠️ **Please hold off merging.** A final adversarial review pass over this branch found **4 blocking issues**, three of which are my own tests being vacuous — verified by mutation testing, i.e. deleting the production line and confirming all 22 tests still pass. Recording them here so nobody approves this unaware; I'll fix them before asking for review again.

### 1. `_pendingUnloads` grows unbounded on the default path — in a PR about unbounded growth

`WarnIfTooManyPendingUnloads` returns at `MaxPendingUnloads <= 0` *before* reaching `CountPendingUnloads()`, and that is the only prune. Since `MaxPendingUnloads` defaults to `0`, the default configuration never prunes: every eviction adds a `WeakReference<AssemblyLoadContext>(…, trackResurrection: true)` that is never removed, so the long GC handle is never released either. Confirmed empirically — 12 evictions left 11 dead-but-retained entries after 20 GC rounds, dropping to 0 only once `Statistics` was read. A control run with `MaxPendingUnloads: 1000` held steady at 1.

Fix: prune unconditionally in the same lock as the `Add`, and compare against `MaxPendingUnloads` (when > 0) after releasing it.

### 2. No test pins the invoker-graph release — the headline leak fix

Removing `_libraryInvokers = …Empty;` from `Unload()` leaves **all 22 tests green**. The collectability test drops every reference to the evicted invoker, so the ALC is collectable regardless of the clearing; the two tests whose names claim graph release only assert that `LibraryInvokers` throws, which is the `_unloadInitiated` flag, not the clearing. A refactor keeping the flag but dropping the clear would silently reintroduce the exact leak this PR exists to fix.

### 3. The concurrent-miss test cannot detect duplicate loads

Switching the `Lazy` to `PublicationOnly` leaves all 22 tests green. `PublicationOnly` runs the factory on several threads and publishes one winner, so `ReferenceEquals`, `Distinct().HaveCount(1)`, `Misses == 1` and `Entries == 1` all still hold — while every discarded load is a permanently leaked ALC. Ironically the test's own comment names precisely the regression it cannot see. Needs to count actual loads.

### 4. LRU recency is never asserted, only capacity

Removing the recency touch from the hit path leaves all 22 tests green. Both eviction tests use `Capacity: 1`, where there is exactly one candidate, so neither recency nor the multi-node walk is exercised — evicting the MRU instead would go undetected. Concrete regression: with `Capacity: 2`, `A, B, A, C` must evict `B`; without the touch it evicts `A`, unloading the caller's hot library set.

### Also confirmed (non-blocking)

- **Memory-model read ordering** in the `LibraryInvokers` getter: `_unloadInitiated` and `_libraryInvokers` are two plain loads in the opposite order to the writes, so on arm64 a thread can see the stale flag with the fresh `Empty` sentinel and return an empty dictionary instead of throwing — the silent-wrong-answer the getter exists to prevent. Fix is to read the dictionary first with `Volatile.Read`, making the failure mode conservative.
- `Statistics.PendingUnloads` and `MaxPendingUnloads` are new public API with no test coverage.
- The pool key is a snapshot of `toolkit.AssemblyBinaries` while the `Lazy` factory re-reads the same mutable toolkit later, so an entry can be filed under a hash that no longer describes its contents.
- `using System.Runtime.CompilerServices;` in the test file duplicates a `global using`, which `CLAUDE.md` forbids.

For calibration: the same pass **refuted 10 other candidate findings**, so this was discriminating rather than a blanket flag.

Stacked PR #1465 is unaffected by these (it only adds an example and docs) but should still merge after this one.

alexzautke:
Feedback from the Firely Server side, where this pool would be adopted for `Measure/$evaluate-measure` (one `Library/$evaluate` per subject × group, i.e. ~10,000 `GetOrCreate` calls per HEDIS run over the *same* library set — the exact workload behind the `vm.max_map_count` aborts this fixes). The design reads as adoptable as-is: we don't use `SetBatchProcessExceptionContinuation`, all our downstream usage goes through the extension methods, and the inert-`Dispose` semantics just mean we drop our `using` on the pooled instance.

Two consumer observations:

1. **Per-call hashing cost.** `GetOrCreate` hashes every binary's assembly + pdb bytes on each call. HEDIS assembly closures run to megabytes, so at ~10k calls per measure run that's tens of seconds of pure SHA-256 per run for a key that never changes within the run. On our side the `byte[]` instances are stable for a whole run (they come from a per-run artifacts cache), so a reference-identity memo of the per-binary hash — e.g. a `ConditionalWeakTable<byte[], string>` inside the pool — would collapse that to one hash per distinct array without weakening the content-based contract (same array ⇒ same content, since the inputs are treated as immutable).

2. **Eviction-while-held surface.** `LibraryInvokers` throwing `ObjectDisposedException` after eviction is the right call vs. silently-empty results, and with capacity ≥ distinct hot library sets it should be unreachable for us. It would be worth one sentence in the README's ownership rules stating the consumer-facing sizing rule explicitly ("size `Capacity` to at least the number of library sets in concurrent use; an evicted-while-held instance throws on next use") — that's the one failure mode an adopter can only mitigate by configuration.

---

Cql/Cql.Invocation/Toolkit/LibrarySetInvokerPool.cs
baseTwo: Fixed in 25eef99c0 — good catch, this was the one real bug of the three.

`GetOrCreate` now re-checks `_disposed` after `lazyInvoker.Value` returns, and if the pool was disposed
in the interim it unloads the orphaned invoker and throws `ObjectDisposedException`, matching the
existing `GetOrCreate_AfterDispose_Throws` contract as you suggested. The opposite interleaving needs
nothing: a `Dispose` that takes the lock *after* the re-check sees `IsValueCreated` true and collects
the entry itself, so both orders are covered.

The regression test, `Dispose_RacingAnInFlightLoad_UnloadsTheOrphanedContextAndThrows`, is deterministic
rather than timing-based. `CreateLibrarySetInvoker` logs before it creates the context, so a callback
`ILoggerFactory` on the toolkit parks the loading thread exactly in the window between entry
registration and load completion — no sleeps, no retry loop. It asserts both halves: that the call
throws, and that the context is gone from `AssemblyLoadContext.All` (initiating an unload removes it,
so one still listed is precisely one whose unload never started). Verified it fails with the fix
reverted — no exception thrown, and the context stays listed.

Cql/Cql.Invocation/Toolkit/LibrarySetInvoker.cs
baseTwo: Fixed in 25eef99c0, and I took your suggested shape rather than the ordering fix I had first written.

You're right that this is check-then-act however the two reads are ordered or fenced. I had initially
reordered the getter to read the graph before the flag, both `Volatile`, which does close the specific
interleaving — but it only works by a release/acquire argument that a future reader has to reconstruct
to avoid breaking it. Making the field the sentinel needs no memory-model reasoning at all, so it is
the better fix.

`_libraryInvokers` is now nullable, `Unload()` sets it to `null`, `_unloadInitiated` survives purely as
the double-unload guard for the context, and the getter is exactly the single read you proposed:

```csharp
get => Volatile.Read(ref _libraryInvokers)
       ?? throw new ObjectDisposedException(nameof(LibrarySetInvoker));
```

(`Volatile.Read` only so the cleared field is observed promptly rather than from a cached value — the
correctness property is the single atomic read, as you described.) I also recorded the reasoning on the
field itself, so the next person does not "simplify" it back into a flag test.

Cql/Cql.Invocation/Toolkit/LibrarySetInvokerPool.cs
baseTwo: Both addressed — point 2 in 14b56bec4, point 1 in 25eef99c0.

**2. Key/load reading the toolkit at different times.** Fixed exactly as you suggested. `GetOrCreate`
snapshots `AssemblyBinaries` and `BatchProcessExceptionContinuation` once, and passes that same
snapshot both to `LibrarySetInvokerPoolKey.Create` and to a new internal
`CreateLibrarySetInvoker(name, isPoolOwned, assemblyBinaries, continuation)` overload, so the entry can
no longer be built from content that does not match its key. Agreed that concurrent toolkit mutation is
a consumer error — the reason to close it is that the failure mode landed in the silent-wrong-results
class, and it would have been permanent.

**1. Hit-path hash cost.** Took the memoization rather than the doc note. Content hashes now live in a
`ConditionalWeakTable<AssemblyBinaryReadOnlyHashSet, string>`, so a repeat call over an unchanged
toolkit is a lookup instead of a full SHA-256 pass. Keying on reference identity is sound for the
reason you gave — `AddAssemblyBinaries` builds a new immutable set and swaps the field, so a given
instance's content never changes — and the weak table keeps a set the consumer has dropped from pinning
its entry for the life of the process. `GetValue` may run the factory twice under contention and keep
one result, which is harmless since hashing is pure.

`PoolKey_ContentHash_IsMemoizedPerSetInstanceButStaysContentBased` pins both halves: stable across
repeated calls on one instance, and still equal across two distinct sets holding equal bytes. That
second half is the real risk memoizing introduces — regressing the key back to identity comparison
would defeat the whole point of hashing content.

baseTwo:
All four blocking issues are now fixed, plus the four non-blocking ones. Mutation-verified each, since
the original point of this pass was that green tests were proving nothing.

| # | Issue | Fixed in | Guarded by |
| --- | --- | --- | --- |
| 1 | `_pendingUnloads` never pruned on the default path | 14b56bec4 | `UnloadEvicted_PrunesReclaimedContexts_WithoutNeedingAStatisticsRead` |
| 2 | Nothing pinned the invoker-graph release | 14b56bec4 | `Eviction_WhileAConsumerStillHoldsTheInvoker_StillReleasesTheAssemblyLoadContext` |
| 3 | Concurrent-miss test could not detect duplicate loads | 5ecf5a7ca | `GetOrCreate_ConcurrentMissesOnOneKey_LoadsTheLibrarySetOnce` |
| 4 | LRU recency never asserted | 14b56bec4 | `Eviction_EvictsTheLeastRecentlyUsedEntry_NotTheMostRecentlyUsed` |

**1.** `WarnIfTooManyPendingUnloads` is now `PruneAndWarn`: it prunes unconditionally and only compares
against `MaxPendingUnloads` when that is `> 0`. The test reads the private list directly rather than
`Statistics`, because reading `Statistics` prunes as a side effect and would mask the very thing being
tested.

**2.** The new test keeps the evicted invoker strongly referenced across the GC loop — that is the whole
point, and it is what distinguishes graph release from the collectability test, which drops every
reference and so passes either way. `Unload()` also now releases its own `_alc` field, which the same
test covers: without it a retained invoker pins the context directly, independently of the graph.

**3.** This one was still open — 552644ea8 fixed the *identity* assertion, which is a different gap.
Re-confirmed the finding by mutation first: with `PublicationOnly` all 28 tests were green. The test now
counts loads that actually happened, using a callback `ILoggerFactory` on the toolkit, since
`CreateLibrarySetInvoker` logs once per invocation. Under `PublicationOnly` it now reports 8 loads
instead of 1. Worth noting the pool's own `Misses` counter cannot serve as the signal: it counts threads
that found no entry, not assembly loads, so it reads 1 either way.

**4.** Exactly the `A, B, A, C` case from the finding, at `Capacity: 2`.

### Non-blocking items

- **Memory-model read ordering** — fixed, then superseded. I first reordered the getter to read the graph
  before the flag, both `Volatile`. @alexzautke pointed out that this is still check-then-act however it
  is fenced, so the field is now itself the sentinel: `_libraryInvokers` is nullable, `Unload()` nulls it,
  and the getter is one atomic read. `_unloadInitiated` survives purely as the double-unload guard. No
  memory-model argument left to preserve.
- **`PendingUnloads` / `MaxPendingUnloads` untested** — `Statistics_PendingUnloads_CountsEvictedContextsThatAreStillAlive`.
- **Key/load reading the toolkit at different times** — fixed in 14b56bec4; both now derive from one snapshot.
- **Duplicate `using System.Runtime.CompilerServices;`** — removed.

Also fixed while here: a failed load leaked its context (958fc137f), and `Dispose()` racing an in-flight
load orphaned one permanently (25eef99c0).

Ready for review again.

baseTwo:
Thanks — this is exactly the feedback that was needed, and point 1 caught a fix that would not have
worked for you. Both addressed in 5ecf5a7ca.

**1. Per-call hashing cost.** You are right, and importantly right about the *key*. I had already added a
memo in 14b56bec4, but keyed on the `AssemblyBinaryHashSet` instance — which is useless for your access
pattern: a fresh `InvocationToolkit` per `Library/$evaluate` means a fresh set instance every call, so
that memo would never have hit and you would still have paid a full SHA-256 pass over megabytes ~10,000
times per run. Re-keyed on the arrays as you suggested.

One deviation from your sketch: rather than one `ConditionalWeakTable<byte[], string>` holding the
combined per-binary hash, each array is hashed and memoized *separately* and the two hashes are then
combined. A memo keyed on `AssemblyBytes` alone would be wrong if the same assembly array were ever
paired with different debug symbols — it would serve the first pairing's hash. Hashing them
independently keeps that correct at no extra cost, since each array is still hashed exactly once ever.
Length prefixes are no longer needed for unambiguous combination, because every component is either
fixed-length hex or an `absent` marker and neither can contain the separator.

Net effect for your run: one pass per distinct array, then each later call hashes only a handful of
64-character hex strings. The set-level memo stays as a short-circuit for consumers that do reuse a
toolkit.

`PoolKey_ContentHash_ReusesArrayHashesAcrossFreshToolkitsOverTheSameArrays` pins it, asserting membership
of the exact arrays in the memo. (The first version of that test asserted the memo "did not grow", which
was vacuous — bypassing the memo entirely also never grows it. It now fails when the memo is bypassed.)

**2. Eviction-while-held sizing rule.** Added as a third bullet to the ownership rules in
[Cql.Invocation/README.md](Cql/Cql.Invocation/README.md): size `Capacity` to at least the number of
library sets in concurrent use, since eviction unloads without waiting for users, so an
evicted-while-held instance throws `ObjectDisposedException` on next use — the one failure mode an
adopter can only avoid by configuration, and unreachable at or above that capacity.

Good to hear the rest reads as adoptable as-is, and that the inert-`Dispose` semantics land the right
way for you.

alexzautke:
Re-reviewed the branch (`5ecf5a7`) from the Firely Server adopter perspective — this is now sufficient for us to adopt as-is once it ships.

Both earlier observations are resolved in code, not just in prose:

1. **Hashing:** the `ConditionalWeakTable<byte[], string>` memo covers our exact access pattern (fresh toolkit per call, arrays stable behind a per-run artifact cache), and hashing the assembly and symbol arrays separately is strictly better than the single-table sketch I gave — it can't serve a stale hash if an assembly array is ever re-paired with different symbols. One pass per distinct array per run is exactly the cost profile we need at ~10k `GetOrCreate` calls.
2. **Sizing rule:** the README's third ownership rule states the capacity contract we'll configure against.

Beyond those, the fix that matters most to our usage is the `LibraryInvokers` getter becoming a single atomic read with the nulled field as sentinel. Pool eviction racing concurrent readers is a supported scenario in our measure path (parallel group evaluations share the pooled invoker), so the old check-then-act returning an *empty* library set would have been a silent-wrong-results hazard for us specifically — good catch making the field itself the sentinel.

Two notes we'll carry into our adoption rather than asking anything of this PR: `MaxPendingUnloads` defaulting to 0 means the leak warning is off by default, so we'll set a positive threshold and surface `Statistics` in our server log; and we'll size `Capacity` from configuration per the README rule.

No blockers from the Firely Server side.

---

alexzautke (approved):

Issue #1452, opened 2026-07-30
Opt-in evaluation cache profiles on FhirCqlContext, a concurrent profile, and a documented thread-safety contract for shared contexts


alexzautke:
## Background

`CqlContext.GetOrCompute` memoizes define results in an opt-in cache: `UseNewCache()` must be called, and today nothing in the production code paths does — the `FhirCqlContext` factories create contexts with the cache disabled, so every cross-define reference re-evaluates the referenced define. Consumers that do enable it get a `ConcurrentDictionary(concurrencyLevel: 1, …)` — correct under concurrency, but every cache write funnels through one lock, which matters for consumers evaluating many expressions concurrently over one context (e.g. a measure evaluator running population groups in parallel over one patient's data).

`GetOrCompute` itself is already designed for concurrency: lock-free reads, `Interlocked` counters, and a compute-then-`GetOrAdd` miss path that tolerates duplicate computation because CQL expression bodies are pure. What is missing is (a) a first-class way to opt in via the FHIR factories, (b) a cache profile suitable for concurrent evaluation, and (c) an explicit, tested contract for what "sharing a `CqlContext` across threads" requires.

## Proposal

**1. Concurrent cache profile (small).**
- `UseNewCache(int initialCapacity, int concurrencyLevel)` overload (existing single-parameter overload unchanged; default behavior identical). Validates `concurrencyLevel >= 1`.

**2. Opt-in via `FhirCqlContextOptions`.**
- A new option (e.g. an `EvaluationCache` profile with initial capacity + concurrency level, `null`/off by default) that makes `FhirCqlContext.CreateContext` call `UseNewCache(...)` on the created context. Default behavior unchanged — the cache stays off unless requested, preserving the semantics for consumers that reuse contexts across data changes.

**3. Thread-safety contract + stress test (the real work).**
- Document on `CqlContext` what concurrent evaluation over one shared context requires: `Definitions` and `Parameters` fully populated before fan-out and not mutated during evaluation (both are plain non-concurrent collections); `GetOrCompute` safe including bounded duplicate factory invocation; the data source and value-set implementations must tolerate concurrent reads (`BundleDataSource`/`IndexedBundle` and `ValueSetSource` already do).
- Add a stress test: evaluate all definitions of a generated test library from N threads over one shared context created with the concurrent profile; assert value-parity with a sequential baseline and consistent cache counters (`CacheHits + CacheMisses == CacheCallCount`, misses bounded).

[...]

Issue #1450, opened 2026-07-30
BundleDataSource: cache extracted codings per resource and avoid reflection/re-extraction on every retrieve


alexzautke:
## Problem

`BundleDataSource.Retrieve<T>` with a code or value-set filter re-does per-candidate work on **every** retrieve over the same bundle (`Cql/Cql.Firely/BundleDataSource.cs`, `Cql/Cql.Firely/IndexedBundle.cs`):

- `IndexedBundle.FilterByType<T>(filter)` calls `codedCandidate.ToCodings()` per candidate per retrieve, re-allocating the coding enumeration each time.
- When a `CodeProperty` is specified, `ExecuteFilter` reads it via `PropertyInfo.GetValue(instance)` (reflection) per candidate per retrieve, then converts with `ToCodings()`.
- `RetrieveByCodes` with an explicit code list checks each candidate coding with `allowedCodes.Any(...)` — `O(codes × codings)` per candidate.

A measure evaluation (e.g. HEDIS) issues many retrieves over the same immutable bundle — commonly dozens of value-set-filtered retrieves against the same resource types — so the same codings are extracted, allocated, and compared over and over.

## Proposal

The bundle inside `IndexedBundle` is immutable for the lifetime of the data source, so extraction work can be done once:

1. **Cache extracted codings per resource.** In `IndexedBundle`, memoize the coding list per candidate per code source — keyed by type for the `ICoded` path and by (type, code property) for the `CodeProperty` path — so each resource's codings are extracted at most once per bundle lifetime, regardless of how many retrieves hit that type.
2. **Compile code-property accessors.** Replace `PropertyInfo.GetValue` with cached compiled delegates (expression trees, cached statically per `PropertyInfo`), eliminating per-candidate reflection.
3. **Hash-set lookup for explicit code lists.** In `RetrieveByCodes`, when the default string comparers are in use, build a set keyed by (system, code) once per retrieve instead of the nested `Any(...)` scan. Custom comparers keep the existing fallback path so semantics are preserved exactly, including null-system handling.

Behavioral requirements:

- Bit-for-bit parity with current results, including: case-insensitive default comparison, null system/code handling, at-most-once yielding per resource, derived-type indexing, and the `ICoded` vs `CodeProperty` paths (both scalar `DataType` and `IEnumerable<DataType>` properties).
- Thread-safe lazy initialization of the caches, so a data source can be shared by concurrent evaluations.

## Context

[...]

PR #1465, opened 2026-07-31
Add pooling example and caching-strategies documentation

### LINKED ISSUE #1464: Add a CqlSdkExamples example for LibrarySetInvokerPool
## Summary

Add a runnable example for `LibrarySetInvokerPool` to `Examples/CqlSdkExamples`, alongside the existing invocation examples.

The pool is being added in [#1461](https://github.com/FirelyTeam/firely-cql-sdk/pull/1461) (for [#1440](https://github.com/FirelyTeam/firely-cql-sdk/issues/1440)) with XML docs and a `Cql.Invocation/README.md` usage section, but no entry in the examples project. Every other consumer-facing invocation capability has one — `310` hello world, `320` parameters, `330` function arguments, `340` caching — and pooling is the one a host most needs to see wired up correctly, because the ownership rules are easy to get wrong.

Filed separately rather than folded into #1461 because `CqlSdkExamples` is **not** in `Cql-Sdk.slnf`, so it sits outside that PR's build and test verification, and because adding an example is not a drop-in file (see Scope below).

## Impact

Without an example, the two rules that keep pooling safe are documented but never demonstrated:

- **Do not dispose a pooled invoker.** It is shared with every other caller for the same library set. `Dispose()` on a pooled instance is a deliberate no-op for that reason, but a consumer copying the `using var invoker = …` idiom from examples `310`–`340` will assume otherwise.
- **Do not retain a `LibraryInvoker` or `DefinitionInvoker` past the library set.** Both reach their library set through a back-reference, so holding one keeps the whole `AssemblyLoadContext` alive even after eviction — the one way a consumer can defeat the pool. `LibrarySetInvokerPoolStatistics.PendingUnloads` is the signal that this is happening.

## Scope

Suggested number `350`, so it sits with the other invocation examples and directly after `340 Caching Example`, which is its closest sibling.

- New `Examples/CqlSdkExamples/350 <name>/Program.cs`, following the established shape: `partial class Program` with a single `void` method, 2026 copyright header.
- The `PickExample` switch in `Examples/CqlSdkExamples/Program.cs` is **code-generated** between `<codegen-switch>` markers by `Examples/CqlSdkExamplesCodeGeneration.targets` — regenerate it via the build rather than hand-editing.
- Update the example-range description in [docs/getting-started.md](../blob/develop/docs/getting-started.md) line ~111, which currently reads "**310–340**: Invoking CQL definitions, parameters, functions, and caching".
- Consider adding the new code to the `codes` array in `Examples/CqlSdkExamples/000 Run All/Program.cs`. Note that array is already out of date — it omits `222` and `340` — so worth deciding whether to fix that at the same time or leave it alone.
- `CqlSdkExamples` is not part of `Cql-Sdk.slnf`, so verify by building/running that project directly.

Worth demonstrating:

1. A pool created once and reused across many evaluations, showing the second request is a cache hit via `Statistics.Hits`/`Misses`.
2. That content keying works: two independently built `InvocationToolkit` instances over equal-but-distinct assembly bytes hit the same pool entry.
3. One pooled invoker shared across concurrent evaluations, each with its own `CqlContext` — mirroring the parallel section of `340 Caching Example`.
4. The ownership rules called out explicitly in the console output and comments, in the same "Key Points" style `340` ends with.

## Acceptance Criteria

- [ ] A new numbered example under `Examples/CqlSdkExamples/` demonstrates creating a `LibrarySetInvokerPool`, reusing it across evaluations, and reading `Statistics`.
- [ ] The example shows a pooled invoker shared across concurrent evaluations with a per-evaluation `CqlContext`.
- [ ] The example states the two ownership rules (do not dispose a pooled invoker; do not retain `LibraryInvoker`/`DefinitionInvoker` past the library set) and mentions `PendingUnloads` as the signal that the second one is being violated.
- [ ] The `PickExample` switch is regenerated via `CqlSdkExamplesCodeGeneration.targets`, not hand-edited.
- [ ] The example-range description in `docs/getting-started.md` is updated.
- [ ] `Examples/CqlSdkExamples` builds, and the new example runs to completion.

Depends on #1461 being merged, since the API it demonstrates lands there.

---

## Added to scope: end-user documentation on caching strategies

The SDK now has several independent caches, each covering a different cost and living for a different length of time:

| Cost | Cache | Scope | Default |
| --- | --- | --- | --- |
| Load assemblies + JIT-compile | `LibrarySetInvokerPool` (#1440) | Process | off |
| Evaluate a definition | `CqlContext` evaluation cache (#1452) | One `CqlContext` | off |
| Convert `FhirDateTime` → `CqlDateTime` | FHIR type-converter LRU | Process | on, 10,000 entries |

<sub>(Corrected 2026-08-04: this row originally read "Convert FHIR ↔ CQL values … on, unbounded". Both halves were wrong — the default is bounded at `FhirTypeConverter.DefaultCacheSize` = 10,000, and the only conversion cached is `FhirDateTime` → `CqlDateTime` plus its `.DateOnly` projection. Note also that setting the size to `0` *disables* the cache rather than unbounding it. Corrected here because the claim was copied verbatim into `docs/caching-strategies.md` in #1465 and would otherwise seed it again.)</sub>
| Extract codings for retrieves | Automatic in `BundleDataSource` (#1450) | One bundle | on |

Nothing explains how they relate, so a consumer cannot tell which one they want — or that pooling and evaluation caching solve *different* problems and compose. Three of the four landed in the last few days, which is how the gap appeared.

### Additional Acceptance Criteria

- [ ] A consumer-facing doc explains each cache: what it saves, its scope/lifetime, its default, and when to reach for it.
- [ ] It states pooling's two ownership rules (do not dispose a pooled invoker; do not retain `LibraryInvoker`/`DefinitionInvoker` past the library set) and names `PendingUnloads` as the signal the second is being violated.
- [ ] It covers the contract for sharing one `CqlContext` across threads, and recommends a context per evaluation over a shared pooled invoker as the simpler default.
- [ ] It documents the lazy-operator interaction: a long-lived cached context can hold references into the generated assemblies and so keep a pooled library set's load context alive.
- [ ] It is linked from `docs/getting-started.md` and from the `Cql.Invocation` and `Cql.Runtime` READMEs, and all relative links resolve.

baseTwo:
## Primary Work

Two consumer-facing additions for the `LibrarySetInvokerPool` introduced in #1461.

### 1. Example `350 Pooling Library Set Invokers`

Every other consumer-facing invocation capability has a runnable example — `310` hello world, `320` parameters, `330` function arguments, `340` caching. Pooling is the one a host most needs to see wired up, because its ownership rules are easy to get wrong when copying the `using var invoker = …` idiom from `310`–`340`.

The example, in order:

1. Unpooled create-per-call, for contrast.
2. Pooling, printing `Hits`/`Misses` so only the first request is a miss.
3. **Proves content keying** — hands the pool separately-built `InvocationToolkit`s over *cloned* byte arrays and shows `ReferenceEquals` is `true`, which reference-equality keying could never achieve.
4. One pooled invoker shared across concurrent evaluations, a `CqlContext` each.
5. `Dispose()` on a pooled invoker being deliberately ignored, then still invoking successfully.

Ends with the ownership rules in the "Key Points" style `340` uses.

**On the code generation:** the `PickExample` switch and `launchSettings.json` are normally generated by `Examples/CqlSdkExamplesCodeGeneration.targets`, which runs only under full MSBuild (`MSBuildRuntimeType != 'Core'`) — explicitly *not* under `dotnet build`. Full MSBuild wasn't available here, so both were written to match what that codegen emits, and I verified the switch-case order is identical to the directory-enumeration order the generator uses:

```
dirs:   000 210 220 221 222 230 240 250 251 310 320 330 340 350 400 410 500
cases:  000 210 220 221 222 230 240 250 251 310 320 330 340 350 400 410 500
```

Also added `350` to the `000 Run All` codes array. Note that array **still omits `222` and `340`** — pre-existing gaps I deliberately left alone rather than silently fixing.

### 2. `docs/caching-strategies.md`

Added to scope during review. The SDK now has several independent caches, three of which landed in the last few days, and nothing explained how they relate:

| Cost | Cache | Scope | Default |
| --- | --- | --- | --- |
| Load assemblies + JIT-compile | `LibrarySetInvokerPool` (#1440) | Process | off |
| Evaluate a definition | `CqlContext` evaluation cache (#1452) | One `CqlContext` | off |
| Convert FHIR ↔ CQL values | FHIR type-converter cache | Process | on, unbounded |
| Extract codings for retrieves | automatic in `BundleDataSource` (#1450) | One bundle | on |

A consumer had no way to tell which one they wanted, or that pooling and evaluation caching solve *different* problems and compose. The doc covers each cache's cost/scope/default, pooling's two ownership rules with `PendingUnloads` as the leak signal, the shared-context contract (recommending a context per evaluation over a shared pooled invoker as the simpler default), and the **lazy-operator interaction** — some list operators evaluate lazily, so a long-lived cached context can hold references into the generated assemblies and keep a pooled library set's load context alive.

Linked from `getting-started.md` and both the `Cql.Invocation` and `Cql.Runtime` READMEs. All relative links and self-anchors verified to resolve.

## Auxiliary Work

- Updated the example-range line in `docs/getting-started.md` from "310–340 … and caching" to "310–350 … and pooling library sets".
- **Running an example no longer breaks the next build.** The project excluded generated C# an example *reads* (`**\input\**\*.g.cs`) but not what it *writes*: 220 and 221 save `.g.cs` into their own `output\cs`, the default glob compiled it, and the build then failed with `CS0579` plus a run of `CS0111`/`CS0102` — duplicate types and members, since the same libraries are already compiled from `input`. Anyone who follows the new caching-strategies doc and actually runs the examples hits this, so it is fixed here: `**\output\**\*.g.cs` is excluded too. Verified by running 221 and rebuilding, which previously failed.
- **Aligned `caching-strategies.md` with the pool as merged.** It said "Two rules" where the `Cql.Invocation` README now has three, and never stated the consequence behind the third: an invoker evicted while still held throws `ObjectDisposedException` on next use. The doc only warned that holding one leaks the load context, which understates it — `Capacity` is a correctness setting, not just a tuning one.
- **Added example 340 to `000 Run All`.** Its absence was an oversight rather than a deliberate exclusion: the commit that added it (5a6781511) never touched Run All. It passes, and this PR's own doc cross-references it beside 350.

## Verification

- `Examples/CqlSdkExamples` builds clean, and still builds clean *after* the examples have been run (the point of the csproj fix above).
- Re-verified every example against #1461 as merged, invoked individually on net10.0: **210, 220, 221, 240, 250, 251, 310, 320, 330, 340, 350, 400, 410 and 500 all pass.**
- Ran `350` end to end; all five sections behave as described, including `Same invoker instance: True` and `hits/misses` reaching `2/1`.
- Confirmed no example uses an invoker after disposing it except 350, which does so deliberately to demonstrate that `Dispose()` is inert while the pool owns the instance.
- Every API the new doc references exists and is public: `CqlContext.CacheInitialCapacity`, `MinimumCacheInitialCapacity`, `UseNewCache`, `DontUseCaching`, `FhirCqlContextOptions.EvaluationCache`, `EvaluationCacheProfile`, `OverrideFhirTypeConverterCacheSize`, `ElmToolkitConfig.LRUCacheSize`, `LibrarySetInvokerPoolStatistics.PendingUnloads`.
- Link-checked the new doc programmatically: no broken relative links, no broken self-anchors.

### Known pre-existing failure, not addressed here

Example **230** fails in any fresh checkout: its `input\dll\*.dll` are gitignored (`.gitignore:45`) and no such files are tracked, so `DirectoryNotFoundException`. `000 Run All` aborts there, which means the chained run never reaches 240 onward — including 350. Every affected example works when invoked directly. Out of scope for this PR and worth its own issue.

## Note on stacking

Was stacked on #1461 (`1440-pool-library-set-invokers`), because the API it demonstrates only existed there. **#1461 has now merged** (`88f5988fa`), GitHub retargeted this to `develop`, and the branch has been rebased onto `develop` — so this is no longer blocked and its diff contains only its own work.

Refs #1464


Examples/CqlSdkExamples/350 Pooling Library Set Invokers/Program.cs
```diff
@@ -0,0 +1,124 @@
...
+using Hl7.Cql.Invocation.Toolkit.Extensions;
+using Microsoft.Extensions.Logging;
+
+namespace CqlSdkExamples;
+
+partial class Program
+{
+    void PoolingLibrarySetInvokers()
+    {
+        // Only warnings and above, so the pool's own behaviour is easy to see. Drop this to
```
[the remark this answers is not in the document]
baseTwo: Fixed in cfc3b705c — you're right, and I checked the convention rather than taking it on trust: 173
occurrences of "behavior" against 3 of "behaviour" across `Cql/`, and the only other instance in
`Examples/` is 340's "default behavior" as you noted. Line 25 now reads "the pool's own behavior".

Also fixed the same class of thing in this PR's other new file, which you couldn't see because this review
only covered the `.cs`: `docs/caching-strategies.md` had "materialise" in section 5. Now "materialize".

Examples/CqlSdkExamples/350 Pooling Library Set Invokers/Program.cs
```diff
@@ -0,0 +1,124 @@
...
+                .ToList();
+
+        // A fresh toolkit over *copies* of the bytes, to show the pool keys on their content rather
+        // than on object identity.
+        InvocationToolkit CreateToolkit() =>
+            new InvocationToolkit(loggerFactory)
+                .AddAssemblyBinaries(
+                    assemblyBinaries.Select(binary => new AssemblyBinary(
+                        (byte[])binary.AssemblyBytes!.Clone(),
+                        (byte[]?)binary.DebugSymbolsBytes?.Clone())));
```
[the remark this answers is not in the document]
baseTwo: Accepted in cfc3b705c, but implemented differently from the suggestion, because an explicit guard would
have been dead code.

`GetElmToAssemblyResults` returns a tuple whose `assemblyBinary` member is **non-nullable** —
`byte[] assemblyBinary`, not `byte[]?` ([ElmToolkitExtensions.cs:59](Cql/CodeGeneration.NET/Toolkit/Extensions/ElmToolkitExtensions.cs#L59)) —
because its pattern match (`AssemblyBinary: {} assembly`) filters null results out before they are yielded.
So the value genuinely cannot be null at this point, and a `?? throw` or an `is null` check would be a
branch that can never be taken.

The `!` was only there because the example round-tripped those bytes *through* `AssemblyBinary`, whose
`AssemblyBytes` property **is** typed nullable — so reading them back required an assertion that the
compiler had already proved two lines earlier. The fix removes the round-trip rather than defending it: the
example now holds the non-nullable tuples from `GetElmToAssemblyResults` and clones from those.

```csharp
var compiled = new CqlToolkit(loggerFactory)
    .AddCqlLibraries(cql)
    .CompileToAssemblies()
    .GetElmToAssemblyResults()
    .Select(result => (result.assemblyBinary, result.debugSymbolsBinary))
    .ToList();

InvocationToolkit CreateToolkit() =>
    new InvocationToolkit(loggerFactory)
        .AddAssemblyBinaries(
            compiled.Select(binary => new AssemblyBinary(
                (byte[])binary.assemblyBinary.Clone(),
                (byte[]?)binary.debugSymbolsBinary?.Clone())));
```

No null-forgiving operator and no guard — non-nullness comes from the type system. That seemed the better
lesson for an example than a defensive throw that cannot fire. Verified: builds with no warnings, and 350
re-run end to end behaves identically (hits/misses reach 2/1, `Same invoker instance: True`, pooled
`Dispose()` still inert).

Examples/CqlSdkExamples/CqlSdkExamples.csproj
```diff
@@ -29,6 +29,13 @@
 		<None Remove="bin\**\input\**;obj\**\input\**" />
 
 		<Compile Remove="**\input\**\*.g.cs" />
+		<!--
+		  Generated C# an example WROTE must be excluded too, not just what it reads. Several examples
+		  save .g.cs into their own output\cs directory, and the default glob would then compile it into
+		  this project on the next build - producing duplicate types and members, since the same library
+		  is already compiled from input. Without this, running the examples breaks the following build.
```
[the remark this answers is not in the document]
baseTwo: Reworded in cfc3b705c. The ALL-CAPS "WROTE" was doing emphasis work that the sentence structure should
carry, and "Generated C# an example WROTE" scans badly even though the reduced relative clause is
technically valid. Now:

> Exclude generated C# that an example writes, not only what it reads. Several examples save `.g.cs` into
> their own `output\cs` directory, and without this the default glob compiles that output into this project
> on the next build. It then fails with duplicate types and members, because the same libraries are already
> compiled from `input` — so running the examples would break the following build.

Same rationale, in the active voice and without the shouting. The exclusion itself is unchanged and still
verified: running example 221 and then rebuilding used to fail with `CS0579` plus a run of
`CS0111`/`CS0102`, and now builds clean.

baseTwo:
Posted a follow up for further optimization, so that we don't have to using hashing on megabytes of assembly bytes to use as a key.  https://github.com/FirelyTeam/firely-cql-sdk/issues/1498

docs/caching-strategies.md
```diff
@@ -0,0 +1,107 @@
...
+Runnable example: **340 Caching Example**.
+
+## 4. FHIR type-conversion caching
+
+Converting between FHIR POCOs and CQL values is cached process-wide. This is on by default and unbounded. To bound it, set a size:
+
+- at context creation, via `FhirCqlContextOptions.OverrideFhirTypeConverterCacheSize`;
+- at compile time, via `ElmToolkitConfig.LRUCacheSize` (`0`, the default, means unbounded).
+
+You normally do not need to touch either. Bound it if you are memory-constrained and converting a very wide variety of values.
```
alexzautke: This section is wrong on every claim it makes, and the summary row on line 16 (`on, unbounded`) repeats it. I checked each against the code rather than assuming, because the direction of the error is the dangerous one: it tells a memory-constrained reader to do something that has the opposite effect.

**1. The default is bounded, not unbounded.** A context built through `FhirCqlContext` passes `options.OverrideFhirTypeConverterCacheSize ?? FhirTypeConverter.DefaultCacheSize` (`Cql/Cql.Firely/FhirCqlContext.cs:55`), and `DefaultCacheSize` is `10_000` (`Cql/Cql.Firely/FhirTypeConverter.cs:25`). `LRUCache<T>` is a real capacity-bounded LRU with eviction, not an unbounded memo.

**2. `0` means *disabled*, not unbounded — so "to bound it, set a size" is backwards.** `FhirTypeConverter.CreateImpl` maps the size `0 => null` for the dateTimes cache (`FhirTypeConverter.cs:81-87`), and `Create` maps `0` on the default model to the cache-*less* `Default` converter (`FhirTypeConverter.cs:55-61`). A reader who follows this page and sets `OverrideFhirTypeConverterCacheSize = 0` to "cap" the cache silently turns it off; one who leaves the default believing it is unbounded is already bounded at 10,000.

**3. `ElmToolkitConfig.LRUCacheSize` is a different converter instance, on a different path.** It feeds `AddCqlCompilerServices(services, config.LRUCacheSize, …)` (`Cql/CodeGeneration.NET/Toolkit/Internal/ElmToolkitServices.cs:46`), which registers a `TypeConverter` in the *compiler's* DI container for ELM → C# code generation. It has no effect on the converter a runtime `FhirCqlContext` builds — that one is constructed independently in `CreateOperators`. Listing it as the "at compile time" knob for the same cache implies it carries over into evaluation. It doesn't.

**4. The cache is much narrower than "Convert FHIR ↔ CQL values".** The only thing `cacheSize` configures is an `LRUCache<CqlDateTime>` handed to `ConvertFhirToCqlPrimitives(dateTimesCache)`, consumed solely by `FhirDateTimeToCqlDateTimeViaCaching` — i.e. `FhirDateTime` → `CqlDateTime` and its `.DateOnly` projection (`FhirTypeConverter.cs:153-155`). Every other conversion registered on that converter is uncached, so the "very wide variety of values" advice on line 88 doesn't describe what is actually being held.

Suggested replacement for the section:

> ## 4. FHIR date/time conversion caching
>
> Converting `FhirDateTime` values to `CqlDateTime` is memoized in a process-wide LRU cache. It is on by default, bounded at 10,000 entries, and shared by every context that uses the default model and cache size. No other FHIR ↔ CQL conversion is cached.
>
> To change the bound, set `FhirCqlContextOptions.OverrideFhirTypeConverterCacheSize` at context creation. Note `0` **disables** the cache rather than unbounding it, and any other value gets its own converter and cache instance rather than the shared default pair.
>
> You normally do not need to touch this. Lower it only if you are memory-constrained; raise it if you evaluate over data with far more than 10,000 distinct date/time values and see the conversion cost.

…and line 16's row becomes `| Convert FhirDateTime → CqlDateTime | FHIR type-converter LRU | Process | on, 10,000 entries |`.

For what it's worth, the error isn't yours originally — the `on, unbounded` row is copied verbatim from the added-to-scope table in #1464, so that issue is worth correcting too so it doesn't seed the same claim again.

---

docs/caching-strategies.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+…or, preferably, when the context is created:
+
+```csharp
+var options = new FhirCqlContextOptions { EvaluationCache = EvaluationCacheProfile.Concurrent };
+var context = FhirCqlContext.ForBundle(bundle, options: options);
+```
+
+`EvaluationCacheProfile.Sequential` is the default shape; `EvaluationCacheProfile.Concurrent` sizes the cache for one writer per processor. `ConcurrencyLevel` only affects how many internal locks cache *writes* spread over — reads are lock-free at any level — so raise it when several threads evaluate over one context, and leave it alone otherwise.
```
alexzautke: The headline sample recommends `EvaluationCacheProfile.Concurrent`, which contradicts the rest of the page in two places: line 73 immediately below says to raise `ConcurrencyLevel` "when several threads evaluate over one context, and leave it alone otherwise", and line 98 says the pattern to prefer "unless you have measured a reason not to" is one `CqlContext` per evaluation over a shared pooled invoker. Under that recommended pattern only one thread ever writes to a given context's cache, so `Sequential` is the correct profile and `Concurrent` is the exception.

It isn't only a consistency nit. `Concurrent` is `ConcurrencyLevel = Environment.ProcessorCount` (`EvaluationCacheProfile.cs:51`), and `UseNewCache` passes that straight into `new ConcurrentDictionary<long, object?>(concurrencyLevel, capacity: initialCapacity)` (`CqlContext.cs:199`), which allocates a lock array sized to the concurrency level. On a 64-core host, a reader who copies this sample into a per-subject loop allocates 64 locks per context per evaluation for a cache exactly one thread will ever write.

Suggest making the "preferably" sample `EvaluationCacheProfile.Sequential`, and adding one line after it along the lines of: "Use `EvaluationCacheProfile.Concurrent` instead only when several threads evaluate over the *same* context — see [5](#5-sharing-one-context-across-threads)."

---

docs/caching-strategies.md
```diff
@@ -0,0 +1,107 @@
...
+
+// Once, for the lifetime of the host.
+using var pool = new LibrarySetInvokerPool(new LibrarySetInvokerPoolOptions(Capacity: 4));
+
+// Per request: loads on the first call, reused after that.
+var invoker = pool.GetOrCreate(invocationToolkit, "HEDIS");
+var result = invoker.InvokeLibraryDefinition(context, libraryIdentifier, "Numerator");
+```
+
+Entries are keyed on the **content** of the assembly binaries, so rebuilding an equivalent `InvocationToolkit` from freshly read bytes still hits the pool.
```
alexzautke: Content is only two thirds of the key. `LibrarySetInvokerPoolKey` is a record struct over `(ContentHash, LibrarySetName, Continuation)` (`Cql/Cql.Invocation/Toolkit/Internal/LibrarySetInvokerPoolKey.cs:34-37`), so the `librarySetName` argument to `GetOrCreate` participates in equality.

That matters more than it sounds. A host that passes a name derived from the request — `pool.GetOrCreate(toolkit, $"HEDIS-{subjectId}")`, which reads perfectly naturally given the parameter also names the load context for diagnostics — misses on every single call. With `Capacity: 4` it then evicts on every call too, so it loads, unloads and re-JITs per subject while accumulating `PendingUnloads`: strictly worse than not pooling, and it looks like the pool is broken rather than misused. It's the easiest way to get this wrong, and this page is where someone will look.

The `GetOrCreate` XML doc does say it ("used for the assembly load context's name and as part of the pool key"), but neither this line, the corresponding line in `Cql/Cql.Invocation/README.md`, nor example 350's "Key Points" mentions it. Suggest something like:

> Entries are keyed on the **content** of the assembly binaries together with the library set name you pass to `GetOrCreate`, so rebuilding an equivalent `InvocationToolkit` from freshly read bytes still hits the pool — but the name must be stable for a given library set. A name that varies per request misses every time.

Worth mirroring into the README and into 350's Key Points line "Entries are keyed on assembly-binary content, not object identity", which has the same gap.

---

Examples/CqlSdkExamples/350 Pooling Library Set Invokers/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+        // 4. One pooled invoker is safe to share across concurrent evaluations, as long as each
+        //    evaluation brings its own CqlContext.
+        Console.WriteLine("4. Concurrent evaluation over one shared invoker:");
+        var shared = pool.GetOrCreate(CreateToolkit(), "PoolingExample");
+        Parallel.For(0, 4, i =>
+        {
+            var context = FhirCqlContext.WithDataSource(); // one context per evaluation
+            var result = shared.InvokeLibraryDefinition(context, cql.LibraryIdentifier, "Answer");
+            Console.WriteLine($"   Thread {i}: Answer = {result}");
+        });
```
alexzautke, line 109: `i` is the `Parallel.For` iteration index, not a thread. With four iterations the partitioner may well run several — or all — of them on the invoking thread, so the output can print `Thread 0` … `Thread 3` when there was only ever one thread. This section's whole purpose is to show that concurrent evaluation over a shared pooled invoker is safe, and it's the one place in the example where the printed claim isn't self-evidently what happened.

Cheap fix that makes the output true and actually demonstrates the point:

```csharp
Console.WriteLine($"   Evaluation {i} on thread {Environment.CurrentManagedThreadId}: Answer = {result}");
```

---

Examples/CqlSdkExamples/CqlSdkExamples.csproj
```diff
@@ -29,6 +29,14 @@
...
 
 		<Compile Remove="**\input\**\*.g.cs" />
+		<!--
+		  Exclude generated C# that an example writes, not only what it reads. Several examples save
+		  .g.cs into their own output\cs directory, and without this the default glob compiles that
+		  output into this project on the next build. It then fails with duplicate types and members,
+		  because the same libraries are already compiled from input - so running the examples would
+		  break the following build.
+		-->
+		<Compile Remove="**\output\**\*.g.cs" />
```
alexzautke: Good catch and the diagnosis is right — `Examples/**/output/` is gitignored (`.gitignore:432-433`), so the generated `.g.cs` never shows up in `git status` and the failure looks like it came from nowhere. Confirmed the trigger too: 220, 221 and 222 all write to `output/cs`, and `SaveCSharpFilesToDirectory` emits `.g.cs`.

One suggestion though: consider collapsing both rules into a single `<Compile Remove="**\*.g.cs" />`. The project never compiles a hand-written `.g.cs` — the only tracked ones are the two under `230 …/input/cs/`, and those are deliberately `None`-included on line 42 for copy-to-output, not compiled. As written the exclusion is keyed on the literal directory name `output`, so the invariant "this project compiles no generated C#" is enforced by two rules that both have to be kept in step with wherever examples happen to write. The next example that saves generated C# to, say, `generated/` or `out/` reintroduces exactly this break, with the same confusing `CS0579`/`CS0111`/`CS0102` signature and the same invisibility in `git status`.

Not blocking — the current fix is correct for every example that exists today.

---

docs/getting-started.md
```diff
@@ -108,10 +108,14 @@ var results = invoker
...
 - **210–251**: Packaging CQL/ELM into FHIR Library resources
-- **310–340**: Invoking CQL definitions, parameters, functions, and caching
+- **310–350**: Invoking CQL definitions, parameters, functions, caching, and pooling library sets
 - **400–410**: Loading and invoking from FHIR resource directories
 - **500**: Extending SDK functionality
 
+If you are running the same CQL repeatedly — evaluating a measure over a population, for instance — read
+**[Caching strategies](caching-strategies.md)** before you tune anything. It covers which of the SDK's
+caches to reach for, what each one costs, and how they interact.
+
```
alexzautke, line 118: Nit: this places the link mid-document, but the "Further Reading" list at the bottom (line 148 onward) enumerates every other `docs/*.md` — engine architecture, dependency diagrams, demo projects, build pipeline, packager, technical README. Caching strategies is now the only one missing, and that list is where a reader scanning for "what else is documented" will look. Worth adding a bullet there as well:

```markdown
* [Caching strategies](caching-strategies.md) — which cache to reach for, what each one costs, and how they interact
```

---

alexzautke (commented):
Reviewed at `93eb278`. Strong PR overall — the example is well-shaped and the doc fills a real gap. One substantive correctness problem in the new doc, plus four smaller items.

No .NET SDK in this environment, so I verified by reading source rather than building or running. Everything below is checked against the code on this branch, not inferred.

## Must fix before merge

**`docs/caching-strategies.md` §4 (FHIR type-conversion caching) is factually wrong, in the dangerous direction.** Details inline. In short: the cache is not unbounded (it's an LRU bounded at `FhirTypeConverter.DefaultCacheSize = 10_000` on the `FhirCqlContext` path); `0` **disables** it rather than unbounding it, so the page's "to bound it, set a size" advice inverts the actual effect; `ElmToolkitConfig.LRUCacheSize` configures the *compiler's* converter and has no bearing on runtime evaluation; and the cache only covers `FhirDateTime` → `CqlDateTime`, not FHIR ↔ CQL conversion generally. A memory-constrained reader following this section would set `0` and silently lose the cache while believing they had capped it.

The `on, unbounded` row is copied verbatim from the added-to-scope table in #1464, so that issue wants the same correction — otherwise it seeds the claim again.

## Worth addressing

- **Pool keying omits the library set name** (`docs/caching-strategies.md:38`, mirrored in `Cql/Cql.Invocation/README.md` and 350's Key Points). The key is `(ContentHash, LibrarySetName, Continuation)`. A name that varies per request — which reads entirely naturally, since the parameter also names the load context for diagnostics — misses 100% of the time and, at `Capacity: 4`, evicts on every call. That's worse than not pooling and presents as the pool being broken.
- **§3's headline sample recommends `EvaluationCacheProfile.Concurrent`**, contradicting both the sentence directly below it and §5's recommendation of one context per evaluation. Under the recommended pattern only one thread writes, and `Concurrent` allocates `Environment.ProcessorCount` locks per context.
- **`Parallel.For` output labels iteration indices as threads** in example 350 §4.
- **csproj `.g.cs` exclusion** is correct; suggested a simplification that makes it robust to future output paths. Not blocking.
- **`docs/getting-started.md` Further Reading list** doesn't include the new doc.

## What I verified as correct

- Every API the doc and example reference exists and is public: `CqlContext.CacheInitialCapacity` / `MinimumCacheInitialCapacity` / `UseNewCache` (both overloads) / `DontUseCaching`, `FhirCqlContextOptions.EvaluationCache`, `EvaluationCacheProfile.Sequential`/`.Concurrent`, `OverrideFhirTypeConverterCacheSize`, `ElmToolkitConfig.LRUCacheSize`, `LibrarySetInvokerPoolOptions(Capacity:)`, `LibrarySetInvokerPoolStatistics.{Hits,Misses,Entries,PendingUnloads}`.
- Every behavioural claim about the pool matches `LibrarySetInvokerPool`/`LibrarySetInvoker` as merged: `Dispose()` inert when `_isPoolOwned`; `SetBatchProcessExceptionContinuation` ignored and warned on a pooled instance; `LibraryInvokers` throwing `ObjectDisposedException` after unload rather than returning an empty set; `Capacity` a target not a ceiling (`CollectEvictions` skips entries still loading); `PendingUnloads` as the leak signal.
- §3 is accurate throughout, including the non-obvious "reads are lock-free at any level" — that matches `ConcurrentDictionary` semantics and the `UseNewCache` remarks.
- §1's `BundleDataSource` row is accurate; the per-bundle coding cache is real (`IndexedBundle._codedByType`/`_codedByProperty`, extracted once per bundle and reused).
- §5's lazy-operator interaction matches the remarks on `LibrarySetInvokerPool`.
- All relative links resolve, and all three self-anchors (`#2-…`, `#3-…`, `#4-…`) match GitHub's slug for their headings.
- The example's non-nullable-tuple approach from `GetElmToAssemblyResults` is right — `assemblyBinary` really is `byte[]`, not `byte[]?`, so the earlier `!` genuinely was removable rather than needing a guard.

## On the hand-written codegen

Acceptance criterion "The `PickExample` switch is regenerated via `CqlSdkExamplesCodeGeneration.targets`, not hand-edited" is not met as literally written, and I've left it unticked — but I traced the generator and the hand-written output is what it would emit. `UpdateLaunchSettingsInSource` derives the case from `tokens[0]` of the directory name and the method from the regex `partial\s+class\s+Program\s+{\s+void\s+(?<method>\w+)\s*\(`, which resolves to `PoolingLibrarySetInvokers` for this file; the profile key is the directory name verbatim with `commandName: "Project"` and `commandLineArgs: "350"`. Both match what's committed. The next full-MSBuild build should be a no-op here rather than producing a diff. Worth a maintainer with Visual Studio confirming that, since it's the one thing neither of us can check from a `dotnet`-only environment.

## Nit

`CLAUDE.md` states docs use hierarchical heading numbers (`# 1.`, `## 1.1.`). The new doc uses `# Title` + `## 1.`, `## 2.`; no sibling in `docs/` numbers headings at all. Three conventions in play, so this is really a note that the stated convention has drifted from practice — not something to fix in this PR.

---

docs/caching-strategies.md
baseTwo: Fixed in c186e74f3. I checked all four claims against the code rather than taking them, and all four hold —
the section was wrong on every point, and you're right that the direction of the error was the dangerous one.

- **Bounded, not unbounded.** `DefaultCacheSize = 10_000` (`FhirTypeConverter.cs:25`), and `FhirCqlContext`
  passes `OverrideFhirTypeConverterCacheSize ?? DefaultCacheSize`. Confirmed.
- **`0` disables.** `CreateImpl` maps `0 => null` for the dateTimes cache, and `Create` maps `0` on the
  default model to the cache-less `Default` converter. So "to bound it, set a size" was actively harmful
  advice: a memory-conscious reader setting `0` to cap it would have silently switched it off.
- **`ElmToolkitConfig.LRUCacheSize` is a different converter.** Confirmed it feeds
  `AddCqlCompilerServices(services, config.LRUCacheSize, …)` (`ElmToolkitServices.cs:46`) for ELM → C#
  codegen and never reaches the runtime converter. Removed from that list, and the page now says
  explicitly that it is **not** this cache, so the next reader does not re-derive the same wrong link.
- **Much narrower than "FHIR ↔ CQL values".** Confirmed the `LRUCache<CqlDateTime>` is passed only to
  `ConvertFhirToCqlPrimitives` and consumed only by `FhirDateTimeToCqlDateTimeViaCaching` — the
  `FhirDateTime` → `CqlDateTime` conversion and its `.DateOnly` projection.

Took your replacement text with one addition (that a non-default size gets its own converter *and* cache
instance rather than the shared pair), retitled the section to **FHIR date/time conversion caching**, and
updated the summary row to `| Convert FhirDateTime → CqlDateTime | FHIR type-converter LRU | Process | on,
10,000 entries |`. Section anchors re-checked against the new heading.

And thanks for tracing it to the source — **#1464 is corrected too**, with a dated note explaining what was
wrong, so the row cannot seed the same claim into the next doc.

docs/caching-strategies.md
baseTwo: Fixed in c186e74f3 — and it wasn't just a consistency nit, so thanks for pricing it.

Confirmed the cost: `EvaluationCacheProfile.Concurrent.ConcurrencyLevel` is `Environment.ProcessorCount`,
and `UseNewCache` passes it into `new ConcurrentDictionary<long, object?>(concurrencyLevel, capacity:
initialCapacity)` (`CqlContext.cs:199`), which allocates a lock array of that size. So the sample as written
told a reader to allocate one lock per core, per context, per evaluation, for a cache exactly one thread
would ever write — under the very pattern the same page recommends two sections later.

The "preferably" sample is now `EvaluationCacheProfile.Sequential`, followed by your suggested line pointing
at [5](#5-sharing-one-context-across-threads) for when `Concurrent` is actually right. I also added the
allocation cost to the `ConcurrencyLevel` paragraph, since "raise it when several threads evaluate over one
context" reads like free tuning otherwise.

docs/caching-strategies.md
baseTwo: Fixed in c186e74f3, in all three places you identified.

You're right that this is the easiest way to get the pool wrong, and the failure mode is nastier than a
plain miss: a per-request name misses *and*, at a small `Capacity`, evicts on every call — so it loads,
unloads and re-JIT-compiles per subject while `PendingUnloads` climbs. That is strictly worse than not
pooling, and it presents as the pool being broken rather than misused. `$"HEDIS-{subjectId}"` is a very
natural thing to write given the parameter also names the load context for diagnostics.

Used your wording in `docs/caching-strategies.md`, with the eviction consequence spelled out, and mirrored
it into `Cql.Invocation/README.md` and example 350's Key Points — the latter is now three lines rather than
one, because "keyed on content, not object identity" was exactly the half-truth that invites the mistake.

Also mentioned the **exception-continuation policy** in both docs, since it is the third key component and
was equally undocumented outside the XML doc.

Examples/CqlSdkExamples/350 Pooling Library Set Invokers/Program.cs
baseTwo: Fixed in c186e74f3. You're right, and it was the one place in the example where the printed claim wasn't
what had been demonstrated — in a section whose entire purpose is to show that concurrent evaluation over a
shared pooled invoker is safe.

Took your line, plus a comment saying why, so nobody "simplifies" it back to the index:

```csharp
// The real thread id, not the loop index: Parallel.For may run several iterations on the
// invoking thread, so printing `i` as a thread would claim concurrency that did not happen.
Console.WriteLine($"   Evaluation {i} on thread {Environment.CurrentManagedThreadId}: Answer = {result}");
```

The run now genuinely evidences the point — four distinct thread ids:

```
   Evaluation 0 on thread 2: Answer = 42
   Evaluation 1 on thread 5: Answer = 42
   Evaluation 3 on thread 11: Answer = 42
   Evaluation 2 on thread 7: Answer = 42
```

Examples/CqlSdkExamples/CqlSdkExamples.csproj
baseTwo: Taken in c186e74f3 — collapsed to a single `<Compile Remove="**\*.g.cs" />`.

You've got the better framing: the invariant is "this project compiles no generated C#", and expressing it
as two directory-keyed rules means it has to be maintained in step with wherever examples happen to write
next. The example that saves to `generated\` would reintroduce the identical break, with the same
`CS0579`/`CS0111`/`CS0102` signature and the same invisibility in `git status`. One suffix-keyed rule cannot
drift that way.

Verified your reasoning about the tracked files before collapsing it: `git ls-files "Examples/**/*.g.cs"`
returns exactly the two under `230 …/input/cs/`, and those are `None`-included for copy-to-output, so
nothing needs them on the `Compile` item.

Re-verified the whole loop after the change: build clean, run 221 (which writes `.g.cs`), rebuild clean,
350 still runs end to end.

docs/getting-started.md
baseTwo: Fixed in c186e74f3 — added with your wording, right after the engine-architecture bullet:

```markdown
* [Caching strategies](caching-strategies.md) — which cache to reach for, what each one costs, and how they interact
```

Agreed on the reasoning: that list enumerates every other `docs/*.md`, so an omission there is where a
reader scanning for "what else is documented" would miss it entirely. The mid-document link stays as well,
since it is in the section where pooling is introduced.

alexzautke (approved):

PR #1467, opened 2026-07-31
Allow sharing a prebuilt bundle index across data sources

### LINKED ISSUE #1466: Allow sharing a prebuilt bundle index across BundleDataSource instances
## Context

In Firely Server, `Measure/$evaluate-measure` evaluates each subject's `$everything` bundle once per measure group (~10 groups for HEDIS measures), each evaluation via `FhirCqlContext.ForBundle(bundle, parameters, valueSets)`. Every call constructs a fresh `BundleDataSource`, whose constructor builds a new `IndexedBundle` over the **same immutable `Bundle` instance**: an O(entries) pass that walks every entry and registers it under its full base-type chain. For a 2,000-resource compartment × 10 groups × 1,000 subjects that is ~20M entry visits per measure run where ~2M would do.

With #1450/#1451 this gets worse in a second dimension: the per-`IndexedBundle` coding caches are discarded and rebuilt with each of the ~10 data sources per subject, so the coding-extraction win only pays off fully within a single group.

## Why the host cannot simply cache the `BundleDataSource`

`BundleDataSource` permanently captures the `IValueSetDictionary` passed to its constructor, and `RetrieveByValueSet` consults it on every retrieve. In a host like Firely Server that dictionary is a `ValueSetSource`, which is *not* a passive snapshot: negative membership checks call `ITerminologyService.ValueSetValidateCode` and canonical misses hit the resolver. Both are DI-scoped, non-thread-safe services (EF Core-backed), and each group evaluation runs in its own DI scope — sometimes concurrently. Caching the whole `BundleDataSource` across scopes therefore either drives one scoped `DbContext` from multiple threads or dereferences services from an already-disposed scope.

The *index*, by contrast, is inherently shareable: `IndexedBundle._byType` is built eagerly in the constructor and only read afterwards, and the #1451 coding caches are `ConcurrentDictionary` (a `GetOrAdd` race at worst duplicates one extraction). So the expensive part is immutable and thread-safe; only the cheap wrapper (valueSets binding, comparers, profile filter) is scope-bound.

## Proposal

Expose a way to construct a `BundleDataSource` that **shares the index of an existing one** while binding fresh terminology/comparer state — without making `IndexedBundle` public. For example:

```csharp
// on BundleDataSource (internal type reached via FhirCqlContext / CqlOperators wiring):
public BundleDataSource WithValueSets(IValueSetDictionary valueSets) =>
    new BundleDataSource(this._indexedBundle, valueSets, _codeComparer, _systemComparer, _profileFilter);
```

plus a `FhirCqlContext`/`FhirCqlContextFactory` surface so a host can (1) obtain an opaque handle to a data source built for a bundle and (2) create later contexts from that handle with per-scope `valueSets` — e.g. an overload of the existing `WithDataSource(...)` path, or a `ForBundle` overload accepting a previously returned data source to clone the index from.

This mirrors the precedent set in 2.11 by the `InMemoryValueSet(InMemoryValueSet)` constructor that shares the lazily-built code hash sets: cache the immutable data, rebuild the scope-bound facade cheaply.

It would also be worth documenting the concurrency contract explicitly ("an `IndexedBundle`/shared index supports concurrent readers after construction"), since hosts sharing the index across parallel evaluations will rely on it.

## Expected impact for hosts

- Bundle index construction drops from once per evaluation to once per bundle (~10× for the measure-evaluation pattern above).
- The #1451 coding caches amortize across all evaluations over the same bundle instead of being rebuilt per evaluation.

Related: #1440 (invoker pooling), #1450 (bundle code indexing). Happy to contribute a PR along these lines if the approach sounds right.

alexzautke:
Closes #1466.

## Problem

Hosts that evaluate the same immutable bundle repeatedly — e.g. Firely Server evaluating a subject's `$everything` bundle once per measure group, ~10× per subject — pay a full O(entries) `IndexedBundle` construction for every `FhirCqlContext.ForBundle` call, and the caches derived from the index are discarded with each data source. Hosts cannot simply cache the `BundleDataSource`: it permanently captures the `IValueSetDictionary`, which in a server is typically scope-bound (negative membership checks resolve through scoped, non-thread-safe terminology services), so a cached source must not cross DI scopes.

The index itself *is* shareable — it is built eagerly in the constructor and read-only afterwards — so this PR separates the expensive immutable index from the cheap scope-bound binding, following the precedent of the `InMemoryValueSet(InMemoryValueSet)` sharing constructor from 2.11.

## Changes

- **`FhirCqlContext.DataSourceForBundle(Bundle, FhirCqlContextOptions?)`** (new public API): builds a reusable, scope-free `IDataSource` over a bundle. `FhirCqlContextOptions.OverrideRetrieveProfileFilter` applies; no value sets are captured.
- **`WithDataSource`**: when handed a `BundleDataSource` together with `valueSets`, the context now uses a lightweight view sharing the source's index but resolving value-set membership through the passed value sets. `WithDataSource(DataSourceForBundle(b), …, vs, …)` is therefore behaviorally identical to `ForBundle(b, …, vs, …)` — without this rebind, a cached source would silently answer retrieve-by-valueset from a stale dictionary. No existing caller can be affected: `BundleDataSource` is internal and was previously unobtainable outside the assembly, and `ForBundle` does not route through this branch.
- **`BundleDataSource.WithValueSets(IValueSetDictionary)`** (internal mechanics): clones sharing the `IndexedBundle`, comparers, and profile filter; returns `this` when the value sets are reference-equal.
- **Concurrency contract documented** on `BundleDataSource` and `IndexedBundle`: the index is complete at constructor return and read-only afterwards, supporting any number of concurrent readers; an individual source is only as thread-safe as the value sets it binds, so a scope-bound source must not leave its scope.
- Extension scenario added to `docs/cql-engine-architecture.md`; release-note fragment added.

Host usage:
```csharp
// once per bundle (cacheable, scope-free):
var source = FhirCqlContext.DataSourceForBundle(bundle);
// per evaluation / per DI scope:
var context = FhirCqlContext.WithDataSource(source, parameters, scopedValueSets);
```

## Testing

New `CoreTests.Fhir.SharedBundleIndexTests` (9 tests): index reference-shared and not rebuilt; same-value-sets returns the same instance; null-argument guards; retrieve-by-type equivalence between original and clone; two clones with different `IValueSetDictionary` stubs resolve independently while the original keeps its own; concurrent smoke test (8 tasks × 50 retrieves across two clones of one source); public-API-level tests through `DataSourceForBundle`/`WithDataSource`.

- `CoreTests` full run (net10.0): 1064 passed, 0 failed, 8 skipped (pre-existing).
- `Cql-Sdk.slnf` build: 0 warnings, 0 errors (PublicAPI analyzer clean; one new `PublicAPI.Unshipped.txt` entry).
- net8.0 target compiles; the net8.0 runtime is not available in the development environment, so that run is left to CI.

Note: the release-note fragment is named `1467-shared-bundle-index.md` as a PR-number placeholder — will rename to match this PR's actual number.

https://claude.ai/code/session_019LxWQZxuUUup1ZULebJnuK

---


alexzautke:
Supporting evidence for this PR from recent profiling of the CMS integration corpus (steady-state thread-time trace, dop1, bundle-parsing removed from the loop): `IndexedBundle` construction accounts for ~3.6% of active CPU per evaluation (as `Array.Copy` under the constructor), and populating its per-type cache for another ~3.6% (`ConcurrentDictionary.TryAddInternal`, 98.6% from `FilterByType`). Since the index is a pure function of the bundle, sharing a prebuilt index across data sources eliminates exactly this — and the benefit multiplies by measures-per-patient in hosts that evaluate a patient's bundle against many measures (the index is rebuilt per context today). ~7% of evaluation CPU is a solid win for that host shape.

---

baseTwo:
@copilot This PR is `CONFLICTING` against `develop` and needs the conflict resolved before it can merge. Please do that first — I'm reviewing the diff in parallel and will post findings separately.

**Merge `develop` into `claude/shared-bundle-index-1466`** (a merge commit, not a rebase — that is this repo's convention for integrating `develop`, and this branch is not part of a stacked chain).

### The only conflict, and its resolution

`Cql/Cql.Firely/PublicAPI.Unshipped.txt` — the merge is otherwise clean (I verified with `git merge-tree`; it is the single conflicting path).

After the merge the file must contain **exactly** this, i.e. only this PR's own new entry:

```
#nullable enable
static Hl7.Cql.Fhir.FhirCqlContext.DataSourceForBundle(Hl7.Fhir.Model.Bundle! bundle, Hl7.Cql.Fhir.FhirCqlContextOptions? options = null) -> Hl7.Cql.Operators.IDataSource!
```

**Do not keep the other three lines** currently on this branch:

```
Hl7.Cql.Fhir.FhirCqlContextOptions.EvaluationCache.get -> Hl7.Cql.Runtime.EvaluationCacheProfile?
Hl7.Cql.Fhir.FhirCqlContextOptions.EvaluationCache.init -> void
static Hl7.Cql.Fhir.FhirTypeConverter.ConvertRangeToCqlInterval(this Hl7.Cql.Conversion.TypeConverter! converter, Hl7.Fhir.Model.Range? range, string? cqlPointTypeName) -> object?
```

They are not yours — they were inherited from the merge base and the 2.12.0 release (#1504, merge commit `cd055ce2a`) has since **promoted all three into `Cql/Cql.Firely/PublicAPI.Shipped.txt`** (lines 65-67 on `develop`), reducing `Unshipped.txt` to `#nullable enable`. Carrying them forward would list the same members as both shipped and unshipped, which the public-API analyzer rejects — and that failure only shows up in the `Build, Sign & Package` job, not in a diff read. `Shipped.txt` itself does not conflict; leave it exactly as `develop` has it.

### Two things that need no change (please don't "fix" them)

- **The release-note fragment is already correctly named.** `docs/releases/vnext/1467-shared-bundle-index.md` matches this PR's number, so the placeholder caveat in the PR description is already satisfied — please just drop that trailing "will rename to match this PR's actual number" note from the description while you're in there. The 2.12.0 cut deleted every fragment, so `docs/releases/vnext/` on `develop` holds only `README.md` and your fragment lands cleanly as the first entry for the next release.
- **The fragment stays under `docs/releases/vnext/`** — do not move its content into `docs/releases/release-notes-2.12.0.md`. 2.12.0 is already tagged and published; this change ships in the next release.

### After the merge

`develop` has moved a long way since this branch was cut at `261501c42` on 2026-07-31 — the 2.12.0 release, the Measure-stratifier fix (#1499) with its regenerated demo `*.g.cs`, and the Integration Runner submodule pointer bump are all in between. So please:

1. Build `Cql-Sdk.slnf` and confirm 0 warnings / 0 errors, including a clean public-API analyzer run.
2. Re-run `CoreTests` on both TFMs and report the actual counts. The PR description's "1064 passed, 0 failed, 8 skipped" predates all of the above and is now stale — update that section to the post-merge numbers rather than leaving the old figure.
3. Confirm all 8 Azure checks are green on the merge commit, and report the **actual** SHA in your reply (a previous reply on #1504 pasted an unexpanded `$(git rev-parse --short HEAD)` — please expand it).

baseTwo (changes requested):
## Review of `50b693e2a` — the sharing mechanism is right; five required fixes, four of them about what happens when a caller gets the two-call workflow slightly wrong

The core design is the right one and matches what #1466 proposed: the expensive, immutable part (`IndexedBundle`) is shared, the cheap scope-bound part (value sets, comparers, profile filter) is rebuilt per evaluation. `WithValueSets` copies every field of the original, the reference-equality shortcut is a nice touch, and the concurrency documentation is unusually thorough.

My findings are almost all in the same place: **splitting `ForBundle` into "build a source" + "create a context" creates two call sites where there was one, and every piece of state that the caller can now supply to the *wrong* one fails silently rather than loudly.** That is worth fixing before this ships, because the failure mode is an under- or over-populated measure result, not an exception.

### Verified independently (not taken from the description)

- **No existing caller is affected — confirmed by tracing, not by trusting the claim.** `ForBundle` builds its source through `CreateDataSource(bundle, valueSets ?? new HashValueSetDictionary(), …)` and then calls `WithDataSource(source, parameters, valueSets, …)`. When `valueSets` is non-null it is the *same instance* the source captured, so `WithValueSets` hits `ReferenceEquals` and returns `this`; when it is null the new branch is skipped entirely because it tests `valueSets is not null`. Either way `ForBundle` allocates nothing extra and behaves exactly as before.
- **`IndexedBundle`'s concurrency claim holds.** `_byType` is a plain `Dictionary` but is fully built in the constructor and only ever read afterwards; `_codedByType` and `_codedByProperty` are both `ConcurrentDictionary`, so the "two threads may both extract, only one result is kept" wording is accurate. I also followed the one indirection the claim depends on that the doc comment does not mention: `CompiledPropertyAccessor.Accessors` is a `ConcurrentDictionary` keyed by getter identity, so the compiled-getter path is safe too. The contract as documented is true.
- **`docs/releases/vnext/1467-shared-bundle-index.md` is correctly named and correctly shaped** — `<PR-number>-<slug>.md` per `docs/releases/vnext/README.md`, and `## Features` is the heading its three closest precedents used (`1451-indexed-bundle-coding-cache.md`, `1453-…`, `1455-…`), all of which 2.12.0 folded up cleanly. Nothing to change here.
- New test file carries the current copyright header and its own `#nullable enable`; `Bundle` was widened to `internal`, not `public`, which is what `CLAUDE.md` asks for; no codegen is touched so no `GeneratorToolVersion` bump is owed; the new architecture-doc section matches that file's unnumbered FAQ style.

---

### 1. `OverrideRetrieveProfileFilter` passed to `WithDataSource` is silently ignored — retrieves then return resources the filter was meant to exclude

`WithValueSets` copies `_profileFilter` from the source, and `WithDataSource` never reads `options?.OverrideRetrieveProfileFilter` (it didn't before either — `CreateOperators` doesn't use it). Only `DataSourceForBundle`/`CreateDataSource` do. So:

```csharp
var source  = FhirCqlContext.DataSourceForBundle(bundle);                 // no options -> QICoreRetrieveProfileFilter.Default
var context = FhirCqlContext.WithDataSource(source, parameters, valueSets,
                  options: new FhirCqlContextOptions { OverrideRetrieveProfileFilter = myFilter });   // silently dropped
```

`myFilter` never runs; retrieves are filtered by the QICore default instead. The equivalent `ForBundle(bundle, parameters, valueSets, options: …)` call *does* apply it, so the description's claim that `WithDataSource(DataSourceForBundle(b), …, vs, …)` is "behaviorally identical to `ForBundle(b, …, vs, …)`" is **not true** in the one dimension `DataSourceForBundle`'s own `options` parameter exists for. Both parameters are optional and trailing, and a caller reasonably reads `options` as per-evaluation — which is exactly when it gets dropped.

Please pick one and make it explicit; silent divergence is the only unacceptable outcome:

- **preferred** — give `WithValueSets` an optional profile-filter argument and have `WithDataSource` pass `options?.OverrideRetrieveProfileFilter` when rebinding, so per-call options win and fall back to the source's; or
- throw an `InvalidOperationException` from `WithDataSource` when it is handed a `BundleDataSource` together with an `OverrideRetrieveProfileFilter` that is not reference-equal to the one the source holds, naming both.

Either way, document the precedence on both methods and pin it with a test (see 4a).

### 2. The unbound source's empty `HashValueSetDictionary` makes a forgotten rebind look like "no data matched"

`DataSourceForBundle` binds `new HashValueSetDictionary()`. That instance is only ever reachable **by mistake** — the whole point of the API is that `WithDataSource` rebinds before use — and when it is reached, `RetrieveByValueSet` resolves every value set as empty and returns nothing. No exception, no log: a measure population is simply empty.

Two ways to reach it, both plausible:

1. `WithDataSource(cachedSource, parameters)` with `valueSets` omitted on one of ~10 call sites. With `ForBundle` the bundle and the value sets were arguments to the same call, so they were hard to separate; now they are in different calls and different lifetimes.
2. **Any wrapping defeats the rebind's type test.** `WithDataSource` matches `source is BundleDataSource`, and `CompositeDataSource` is *public* and takes `params IDataSource[]`, so a host combining a cached bundle source with a second source — the pattern the section right above the new docs describes — silently keeps the empty dictionary:
   ```csharp
   var composite = new CompositeDataSource(cachedBundleSource, myOtherSource);
   var context   = FhirCqlContext.WithDataSource(composite, parameters, scopedValueSets);  // no rebind, value sets ignored
   ```

Please make the unbound state fail loudly instead of quietly: bind a private sentinel `IValueSetDictionary` whose members throw with a message that says the source came from `DataSourceForBundle` and must be handed to `WithDataSource` together with value sets. That covers both routes at once, and it is the same "validate and fail explicitly rather than fall back silently" rule `CLAUDE.md` states for `Library.Name`/`Library.Id`. Additionally, have the rebind recurse into `CompositeDataSource.DataSources` (rebinding each `BundleDataSource` it finds and leaving the rest untouched) so case 2 works rather than merely failing — `DataSources` is already public, so this needs no new surface. Tests for both.

### 3. The memory contract of a cached source is undocumented, and it is the whole point of the change

The XML docs cover the concurrency contract thoroughly and say nothing about lifetime. But the caches that now survive across evaluations are `_codedByType` and `_codedByProperty`, which hold a `Coding[]` **per resource, per retrieved type, and per (type, code-property) pair**. For the motivating case — a 2,000-resource `$everything` bundle retrieved under many types across ~10 groups — a cached source retains the bundle's entire extracted coding surface for as long as the host holds it, where previously it died with the evaluation. There is no way to release it short of dropping the source, and nothing bounds the number of distinct retrieve shapes.

That is an acceptable trade, but the host has to know it is making it. Please state on `DataSourceForBundle` that the returned source retains the index *and* the coding caches derived from it for its entire lifetime, that the caches grow with the number of distinct retrieve shapes, and that a host caching sources must bound that cache (per subject/request, not a static dictionary keyed by bundle). #1460 is open precisely about process-lifetime retention in this area, so please cross-reference the trade-off rather than leaving the next reader to discover it. If you have a measured figure for the motivating bundle, put it in the release-note fragment too — the fragment currently sells only the win.

### 4. Test gaps — all four required

The 9 tests cover index sharing, value-set independence, the null guards and the public-API path well. What they do not cover is everything the clone constructor copies *other than* the index, so a future refactor that drops a field from that constructor stays green:

- **a. Nothing pins `_profileFilter` surviving the clone**, nor which filter wins once finding 1 is resolved. Add a source built with a custom `IRetrieveProfileFilter`, retrieve through a clone with a `RetrieveParameters.TemplateId` that the filter acts on, and assert the filtered result — plus a test for the precedence rule you choose in finding 1.
- **b. `RetrieveByCodes` is never exercised on a clone at all** — every test uses `Retrieve<T>(null)` or a `ValueSet` parameter, so `_codeComparer`, `_systemComparer` and `_usesDefaultComparers` are copied but unverified. Dropping `_usesDefaultComparers` from the clone constructor would flip retrieves from the `BuildSetFilter` hash-set path to the linear `listFilter` path with identical results — a silent loss of the #1451 win that no test would notice. Add a retrieve-by-codes test through a clone, and one with non-default comparers to pin that they are carried over.
- **c. The `_codedByProperty` cache is never touched by this class.** `ByValueSet()` passes `codeProperty: null`, so only `_codedByType` is exercised. `_codedByProperty` is the harder of the two #1451 caches — keyed by getter identity, with a value factory that closes over a delegate created fresh on every call — and it is the one whose behavior under a *shared* index is least obvious. Add a retrieve that supplies `RetrieveParameters.CodeProperty` through two clones of one source, and include it in the parallel test.
- **d. The parallel test has no start gate**, so 8 `Task.Run` bodies may well run mostly sequentially on a cold thread pool and never race `GetOrAdd` for the same key. Gate them on a `Barrier`/`ManualResetEventSlim` so all tasks hit the first retrieve together — that is what makes the racing-value-factory behavior the `IndexedBundle` remarks describe actually exercised. (Assertions inside the tasks do surface correctly through `Task.WaitAll`, so that part is fine.)

### 5. `docs/cql-engine-architecture.md:115` — the snippet directly above your new section does not compile

```csharp
var ctx = FhirCqlContext.ForBundle(bundle, dataSource: new MyDataSource());
```

`ForBundle` has no `dataSource` parameter (its signature is `bundle, parameters, valueSets, now, options`), and there is no other overload. The correct call is `FhirCqlContext.WithDataSource(new MyDataSource())` — the very method your new section documents, which is why this belongs in this PR rather than a separate one: you are adding a `WithDataSource` section immediately below a snippet that tells readers to reach the same goal through a parameter that does not exist. Please fix that line while you are in the file.

---

@copilot please apply findings 1-5 (including all four test gaps in 4 — each is a required fix, not optional polish), on top of the `develop` merge requested in my previous comment. When you reply, cite the actual commit SHA and say what changed for each numbered item; where a fix is a design choice (finding 1's precedence rule, finding 2's sentinel vs. recursion), say which you picked and why.

baseTwo (changes requested):
## Verification of `32c12cae6` + `8b037b4ed` — three of five fully fixed, and one of your two self-initiated fixes was on my list before I could post it. Four items still open

Checked each fix against the code rather than against the summary. Findings 1, 3, 4a, 4c, 4d and 5 are genuinely closed; finding 2 is closed for the case it was written about but its documented guarantee overreaches the implementation; finding 4b is **not** closed, and the reply says it is.

Also worth noting: `8b037b4ed`'s two changes — tightening the `WithValueSets` reference-equality shortcut and swapping the gate for a `Barrier` — were both items I had already drafted for this round. The shortcut one mattered: as written in `32c12cae6` it silently reintroduced a per-call allocation on the `ForBundle` hot path, and your fix is exactly right. I re-derived it to be sure the property is restored: the public constructor stores `profileFilter ?? QICoreRetrieveProfileFilter.Default`, so when a caller passes options carrying a filter, `ReferenceEquals(profileFilter, _profileFilter)` holds and `WithValueSets` returns `this` — `ForBundle` allocates nothing extra again, with or without a custom filter.

### Confirmed fixed

- **Finding 1** — precedence is implemented as asked (per-call filter wins, `?? source._profileFilter` falls back) and pinned from both directions by `WithValueSets_ProfileFilterOverrideTakesPrecedenceOverSourceFilter` and the public-API-level `WithDataSource_ProfileFilterOptionOverridesSourceFilter`.
- **Finding 2 (sentinel half)** — all four `IValueSetDictionary` members throw, and `InvalidOperationException` is the **correct** exception type here, which is worth stating explicitly because this repo has a rule that points the other way: `CLAUDE.md` requires `CqlException<TError>` for anything the CQL spec says should signal an error to the calling environment, and reserves `InvalidOperationException` for programming errors. A host forgetting to rebind is host misuse, not a spec-mandated evaluation error, so this is on the right side of that line. `DataSourceForBundle_DirectRetrieveByType_Succeeds` correctly pins that the sentinel does not block type-only retrieves.
- **Finding 3** — the memory-contract paragraph says what it needed to (retention for the source's lifetime, growth per retrieve shape, never trimmed, scope the host cache per subject/request), and the fragment carries the caveat too.
- **Finding 4a** — mutation-checked by reading: drop `_profileFilter` from the cloning constructor and the field is left `null`, so `ApplyProfileFilter` throws an NRE and both tests fail. Real coverage.
- **Finding 4c** — better than what I asked for. Because `ByCodeProperty()` is called separately for each retrieve, `fromA` and `fromB` pass **two distinct `FhirModelPropertyInfo` instances**, so the test actually exercises the getter-identity keying that `_codedByProperty`'s comment is about, not just the shared-index reference.
- **Finding 4d** — `Barrier(TaskCount)` with `SignalAndWait()` is strictly better than the gate: threads rendezvous rather than merely being released together.
- **Finding 5** — fixed, and `WithDataSource(new MyDataSource())` does bind to the `source` parameter.
- Test semantics verified against the real APIs rather than assumed: `RetrieveParameters(CodeProperty, ValueSet, Codes, TemplateId)` matches the positional order every helper uses; `StringCqlComparer` normalizes and then defers to the supplied `StringComparer`, so `Ordinal` genuinely makes `"X" != "x"` and `RetrieveByCodes_CloneInheritsNonDefaultComparers` tests what it claims; `Cql.Runtime` grants `InternalsVisibleTo` to `CoreTests` in both Debug and Release, so reaching the internal `StringCqlComparer` from the test compiles; 17 `[TestMethod]`s counted, matching your figure.

---

### 6. `RebindValueSets` does **not** recurse — one level only, and both the reply and the XML doc say otherwise

Your reply says it "recurses into `CompositeDataSource.DataSources`", and the XML doc on `WithDataSource` says "each bundle-backed component within it is rebound the same way". The code does a single `.Select` over the direct children:

```csharp
var rebound = composite.DataSources
    .Select(s => s is BundleDataSource bs ? bs.WithValueSets(valueSets, profileFilter) : s)
    .ToArray();
```

A `CompositeDataSource` nested inside a `CompositeDataSource` — legal, since the constructor takes `params IDataSource[]` — is passed through untouched, so the `BundleDataSource` inside it keeps the sentinel and now throws on the first value-set retrieve. The failure is loud, which is the important part, but a documented guarantee that the implementation does not provide is worse than no guarantee. The fix is one line:

```csharp
.Select(s => RebindValueSets(s, valueSets, profileFilter) ?? s)
```

Please make it actually recursive and add a nested-composite test, or narrow both the doc and the reply to "direct components". Recursive is preferable — it costs nothing and matches what the doc already promises.

### 7. The composite branch allocates a new `CompositeDataSource` unconditionally, which under `#if VNEXT` leaks an event subscription per evaluation

`CompositeDataSource`'s constructor does `foreach (var r in sources) r.DataChanged += triggerDataChanged;` under `#if VNEXT`, and the class has **no unsubscribe and no `Dispose`** (I checked: only `+=`, never `-=`). `RebindValueSets` now builds a fresh composite on *every* `WithDataSource` call, so every evaluation permanently attaches another handler to each non-bundle component — which are the host's own long-lived sources — and each of those handlers roots the per-evaluation composite, and through it the bundle clones. Over the motivating workload (10 groups × 1,000 subjects) that is 10,000 handlers on a source that is supposed to outlive them all.

`VNEXT` is not defined in any `*.props`, `*.csproj` or `*.targets` in the repo, so this is latent today — but this PR is what introduces per-evaluation composite construction, and unbounded retention rooted in a long-lived object is precisely what #1460 is open about, which finding 3 now points readers at. Please return the original instance when nothing actually changed:

```csharp
var rebound = composite.DataSources.Select(...).ToArray();
return rebound.SequenceEqual(composite.DataSources) ? composite : new CompositeDataSource(rebound);
```

That removes the allocation entirely for the common "no bundle source inside" case, and confines the VNEXT subscription to calls that genuinely had something to rebind. If the VNEXT path is meant to survive long-term, a comment noting that a composite's subscriptions live as long as its components would help the next reader.

### 8. Finding 4b is still open: `_usesDefaultComparers` is not pinned, and neither new test can detect its loss

The reply states both new tests pin "`_codeComparer`/`_systemComparer`/`_usesDefaultComparers`". The first two are pinned. The flag is not, and it is the one whose loss is silent:

- **`RetrieveByCodes_OnAClone_UsesDefaultComparersFromTheSource`** — drop `_usesDefaultComparers` from the cloning constructor and it defaults to `false`, so `RetrieveByCodes` takes the `_ => listFilter` branch instead of `BuildSetFilter`. `listFilter` uses the comparers that *were* copied, which are the default case-insensitive ones, so `"X"` still matches `"x"` and the assertion still passes.
- **`RetrieveByCodes_CloneInheritsNonDefaultComparers`** — that scenario has `_usesDefaultComparers == false` by construction, so it cannot observe the flag being lost either.

So the net effect of dropping one line from the clone constructor is that **every clone silently degrades from the O(1) hash-set lookup to an O(n) linear scan per coding** — losing exactly the #1451 win this PR exists to amortize — with all 17 tests green. The two filter paths are deliberately behaviorally identical (`BuildSetFilter`'s remarks spell that out), so this cannot be pinned behaviorally; it needs a white-box assertion. Please add `internal bool UsesDefaultComparers => _usesDefaultComparers;` and assert it is `true` on a clone of a default-comparer source and `false` on a clone of a custom-comparer source — or assert the private field by reflection if you would rather not widen the type at all. Either is fine; leaving the hole is not.

### 9. Nit — the shipped XML doc points at a bare `#1460`

`DataSourceForBundle`'s new memory-contract paragraph ends with "See GitHub issue #1460 for background". XML doc comments ship inside the NuGet package and are read in IntelliSense by consumers with no GitHub context, where `#1460` resolves to nothing. Please use the full URL (`https://github.com/FirelyTeam/firely-cql-sdk/issues/1460`) — same reasoning as fully qualifying a cross-repo reference.

### 10. The reply cites a commit that does not exist

The comment attributes the source and doc changes to `a30db36`. That object is not in this repository (`git cat-file -e a30db36` fails); the actual commits are **`32c12cae6`** (five fixes) and **`8b037b4ed`** (shortcut + `Barrier`). This is the second time in two PRs that a reply has not cited a usable SHA — on #1504 it was an unexpanded `$(git rev-parse --short HEAD)`. The point of citing the commit is that someone reading the resolved thread later can go look at it, so please quote SHAs that exist on the branch.

---

@copilot please apply items 6-9 (item 10 just needs correcting in your next reply, no code). **Approval is held on CI, not on these items**: `Build, Sign & Package` is still pending on `8b037b4ed`, and that job is what runs the public-API analyzer over the new `DataSourceForBundle` entry and executes the 17 new tests on both TFMs — neither of which a diff read can establish. Once it is green and 6-9 are in, I expect to approve.

baseTwo:
**CI hold discharged — all 8 checks are green at `8b037b4ed`**, verified against that commit's own check-runs endpoint rather than the PR-level rollup, so the result is pinned to this exact SHA:

- **`Build, Sign & Package`** (8m36s) — this was the load-bearing one: it runs the public-API analyzer over the new `PublicAPI.Unshipped.txt` entry for `DataSourceForBundle`, which no amount of diff reading can establish.
- **`Test on .NET 8`** (9m45s) and **`Test on .NET 10`** (10m9s) — so the 17 `SharedBundleIndexTests` pass on both TFMs, and the test's access to the internal `StringCqlComparer` and `BundleDataSource` compiles under both the keyed (Release) and unkeyed (Debug) `InternalsVisibleTo` configurations.
- `HEDIS 2025 Golden Parity`, `Integration Tests`, `Compare Test Results Across Frameworks`, `Check for skippable changes` — all pass.

So the only thing between this PR and an approval is items **6-9** from [my last review](https://github.com/FirelyTeam/firely-cql-sdk/pull/1467#pullrequestreview-4872283982), none of which CI can catch by construction:

- **6** — make `RebindValueSets` actually recursive (`.Select(s => RebindValueSets(s, valueSets, profileFilter) ?? s)`) plus a nested-composite test, or narrow the XML doc to "direct components".
- **7** — return the existing `composite` when no child changed, so the `#if VNEXT` `DataChanged` subscription is not attached once per evaluation to the caller's long-lived sources.
- **8** — pin `_usesDefaultComparers` white-box; it is the one field whose loss both current tests pass through silently, and losing it costs exactly the #1451 hash-set win.
- **9** — full URL instead of a bare `#1460` in the shipped XML doc.

@copilot over to you; ping me when those are in and I will re-verify and approve.

baseTwo (changes requested):
## Third pass on `e9ec0485b` — items 6-9 are all correctly fixed, one of them better than I asked for. One cosmetic item left, and approval waits on CI at this SHA

Traced each fix through the code rather than reading the summary.

- **Item 6 — recursion is real now.** `RebindValueSets` calls itself per child, so `composite(composite(bundleSource))` reaches the bundle source; I traced the nested case and the inner composite is rebuilt only because its own child changed, which is what makes the outer one rebuild too. The doc no longer overclaims, and `WithDataSource_NestedCompositeContainingBundleDataSource_RebindsValueSetsRecursively` pins it two levels deep.
- **Item 7 — better than my suggestion.** I proposed `rebound.SequenceEqual(composite.DataSources)`, which walks the children twice. The lazy `rebound ??= (IDataSource[])sources.Clone()` does it in one pass and allocates nothing at all in the common case, and cloning rather than writing into `composite.DataSources` matters — that array is a public property, so mutating it in place would have corrupted the caller's composite. `WithDataSource_CompositeWithNoBundleSource_ReturnsSameCompositeInstance` pins the identity, which is the assertion that actually keeps the `#if VNEXT` `DataChanged` subscription from being attached per evaluation.
- **Item 8 — pinned, and it does detect the mutation.** Drop `_usesDefaultComparers` from the cloning constructor and the clone's field is `false` while the original's is `true`, so the first assertion fails. Reflection over widening the type is the right call of the two options.
- **Item 9 — `<see href="…">` is the correct XML idiom** for an external link, so consumers get a real hyperlink in IntelliSense rather than a dangling `#1460`.
- Test count checks out: 17 + 3 = 20 in `SharedBundleIndexTests`.

One thing I checked that is worth recording, because it would be a defect if one detail were different: making the rebind recursive means a `BundleDataSource` nested anywhere inside a composite now gets its value sets **replaced**, where before this PR it kept its own. That is only safe because `BundleDataSource` is `internal` and the sole way an external host can hold one is `DataSourceForBundle`, which binds the throwing sentinel — so an externally-held bundle source always wants rebinding, and the only other construction site (`CreateDataSource`, used by `ForBundle`) never builds composites. If `BundleDataSource` is ever made public, this recursion becomes a silent override of a deliberate binding and will need an opt-out.

### 11. Redundant namespace qualification — `System.Reflection` and `System.Threading` are already global usings in CoreTests

`Cql/CoreTests/GlobalUsings.cs` declares `global using System.Reflection;` and `global using System.Threading;`, so these qualify types whose namespaces are already imported:

- `System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance` (new in this commit) → `BindingFlags.NonPublic | BindingFlags.Instance`
- `using var barrier = new System.Threading.Barrier(TaskCount);` (from `8b037b4ed`) → `new Barrier(TaskCount)`

It is also internally inconsistent: `ObservationCodeProperty()` two methods above uses `PropertyInfo` unqualified, from that same global `System.Reflection`. `CLAUDE.md` asks contributors to check `GlobalUsings.cs` before adding a `using` so the global set is not restated locally, and fully qualifying is the same restatement in a different form. Please drop both qualifications.

---

@copilot please apply item 11 — it is the last one. **Approval is held only on CI at `e9ec0485b`**: the previous run was green at `8b037b4ed`, but `e9ec0485b` changed production code in `RebindValueSets` and added three tests, so the green from the earlier SHA does not carry over. `Build, Sign & Package` (public-API analyzer) and `Test on .NET 8`/`.NET 10` are the two that matter, the latter because the new reflection-based test reads a private field by name and would fail at runtime rather than at compile time if that field were ever renamed. Once item 11 is in and CI is green on the resulting commit, I will approve.

baseTwo:
@copilot One change is still outstanding — it is the only thing between this PR and my approval. Restating it here in full rather than by reference, since my previous review said "apply item 11" and your session closed without a commit.

**File:** `Cql/CoreTests/Fhir/SharedBundleIndexTests.cs`

**Change:** remove the redundant namespace qualification in two places. Both namespaces are already `global using`s in `Cql/CoreTests/GlobalUsings.cs` (`global using System.Reflection;` on line 18 and `global using System.Threading;` on line 23), so the qualification restates what is already imported project-wide.

1. In `WithValueSets_CloneInheritsUsesDefaultComparersFlag` (added in `e9ec0485b`):

```csharp
// before
var usesDefaultComparersField = typeof(BundleDataSource)
    .GetField("_usesDefaultComparers", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance)!;

// after
var usesDefaultComparersField = typeof(BundleDataSource)
    .GetField("_usesDefaultComparers", BindingFlags.NonPublic | BindingFlags.Instance)!;
```

2. In `Retrieve_OnSourcesSharingAnIndexFromParallelTasks_ReturnsCorrectResults` (added in `8b037b4ed`):

```csharp
// before
using var barrier = new System.Threading.Barrier(TaskCount);

// after
using var barrier = new Barrier(TaskCount);
```

**Why:** `CLAUDE.md` asks contributors to check `GlobalUsings.cs` before adding a `using` so the global set is not restated locally, and fully qualifying is that same restatement in another form. It is also internally inconsistent with the rest of this very file — `ObservationCodeProperty()` uses `PropertyInfo` unqualified, from that same global `System.Reflection`, and the file aliases `Task` rather than qualifying it.

Please do **not** add local `using System.Reflection;` / `using System.Threading;` directives to the file as part of this — the globals already cover it, and adding them would be the duplication `CLAUDE.md` warns about.

Nothing else is required. Note that this will invalidate the current green CI run: all 8 checks passed at `e9ec0485b` (verified against that commit's check-runs endpoint), but a new commit means a new run, and I will re-verify on the resulting SHA before approving. Please cite the real commit SHA in your reply — the earlier reply cited `a30db36`, which is not an object in this repository.

baseTwo (approved):
## Approving `2d00f474e`

All eleven items from the three review rounds are fixed and verified, and the last hold — CI on this SHA — is discharged. Verified against `2d00f474e`'s own check-runs endpoint rather than the PR-level rollup: **all 8 checks green**, including `Build, Sign & Package` (the public-API analyzer over the new `DataSourceForBundle` entry) and `Test on .NET 8` + `.NET 10` (the 20 `SharedBundleIndexTests` on both TFMs, and the reflection-based test that reads `_usesDefaultComparers` by name — a runtime failure, not a compile one, if that field is ever renamed). All 4 inline threads resolved, 0 unresolved.

### What this PR ends up being

The mechanism was right from the first read: share the expensive immutable `IndexedBundle`, rebuild the cheap scope-bound facade per evaluation. What the review changed is everything around the seam that the two-call workflow opens up — because splitting `ForBundle` into "build a source" and "create a context" creates two places to put per-evaluation state, and in the original every misplacement failed silently:

- A profile filter handed to `WithDataSource` was **dropped**, so retrieves silently returned resources the filter existed to exclude. Now the per-call filter wins with a documented precedence, pinned from both directions.
- A forgotten rebind resolved **every value set as empty** with no error. Now the unbound source carries a sentinel dictionary that throws with a message naming the fix — and `InvalidOperationException` is the right call, since host misuse is a programming error rather than the spec-mandated evaluation error this repo reserves `CqlException<TError>` for.
- Wrapping the source in a `CompositeDataSource` defeated the rebind's type test entirely. Now rebinding recurses through nested composites, and returns the original instance untouched when nothing changed — which also keeps the `#if VNEXT` `DataChanged` subscription from being attached once per evaluation to the caller's long-lived sources.
- The memory contract is now stated where a host will read it: the source retains the index *and* its coding caches for its whole lifetime, the caches grow per distinct retrieve shape and are never trimmed, so a host cache must be scoped per subject or request. That paragraph is the part I would most want a consumer to see, given #1460.

Test coverage went from 9 to 20, and the additions are the ones that matter: the clone's profile filter and comparers are pinned, `RetrieveByCodes` is exercised on a clone at all, `_codedByProperty` is exercised through two clones with two distinct `FhirModelPropertyInfo` instances (so the getter-identity keying is genuinely covered), the parallel test rendezvouses on a `Barrier` so it actually races `GetOrAdd`, and `_usesDefaultComparers` is pinned white-box because both filter paths are deliberately behaviorally identical and no black-box test can see the difference.

Two of the fixes came back better than what I asked for: the lazy `rebound ??= (IDataSource[])sources.Clone()` beats my two-pass `SequenceEqual` suggestion and avoids mutating the public `DataSources` array, and the `Barrier` rendezvous beats the release-gate I proposed.

### One thing to carry forward

Recursive rebinding **replaces** the value sets of a `BundleDataSource` nested anywhere inside a composite, where before it kept its own. That is safe today only because the type is `internal` and the sole externally reachable instance comes from `DataSourceForBundle` bound to the throwing sentinel, so it always wants rebinding — and the only other construction site never builds composites. **If `BundleDataSource` is ever made public, this silently overrides a deliberate binding and will need an opt-out.**

Nothing outstanding. Merging with a merge commit per this repo's `develop`-integration convention.

PR #1469, opened 2026-08-01
Resolve age operators' patient lookup once

### LINKED ISSUE #1468: Age operators re-resolve the patient mapping, birth-date property and retrieve on every call
## Summary

The age operators (`AgeInYears`, `AgeInYearsAt`, `CalculateAge` and their siblings, reached through
`ICqlOperators.Age` / `ICqlOperators.AgeAt`) redo their entire setup on every call: they ask the model inspector for
its patient mapping, re-resolve `IPatient.BirthDate` by reflection, and close `IDataSource.Retrieve<T>` over the
patient type with `MakeGenericMethod` before invoking it reflectively.

None of that depends on the arguments. It is fixed for the lifetime of the operators instance, but it is paid per
call — and measure logic calls these operators per element of a query, not once per patient.

## Impact

Measured over 200 calls against a one-patient bundle, on .NET 10:

| | Per `AgeAt` call |
| --- | --- |
| Time | 13.4 µs |
| Allocated | 6.5 KB |

Of that, a single access to `ModelInspector.PatientMapping` accounts for **15.8 µs and 5.8 KB** in isolation — it
searches the inspector's class mappings and allocates while doing so. `FhirTypeResolver.PatientType` is a plain
property with no memoization in front of it, so every call pays that search.

For comparison, the work the operator actually has to do — retrieve the patient, read the birth date, parse it and
count the calendar years — is under 1.5 µs. The overhead is roughly an order of magnitude larger than the
computation.

A measure that evaluates `AgeInYearsAt(start of X)` inside a `where` clause over a few hundred encounters spends
several milliseconds per patient on this alone.

## Observed Behavior

`CqlOperators.Age(string)` and `CqlOperators.AgeAt(CqlDate?, string)` each do, on every invocation:

1. `TypeResolver.PatientType` → `Inspector.PatientMapping?.NativeType` (searches the class mappings, allocates).
2. `TypeResolver.PatientBirthDateProperty` → `typeof(IPatient).GetProperty(...)` (reflection lookup).
3. `typeof(IDataSource).GetMethod(nameof(IDataSource.Retrieve)).MakeGenericMethod(patientType)` followed by
   `method.Invoke(...)`.
4. `patients.ToArray()` purely to test whether exactly one patient came back.

The two operators are also near-identical copies of each other.

## Root Cause

`FhirTypeResolver.PatientType` and `FhirTypeResolver.PatientBirthDateProperty` are computed properties that resolve
from scratch on each read, and the age operators resolve their retrieve by reflection inline rather than binding it
once. Neither the inspector, its mappings, nor the patient type change over the lifetime of a resolver or of a
`CqlOperators` instance.

## Repro Steps

```csharp
var bundle = new Bundle { Type = Bundle.BundleType.Collection };
bundle.Entry.Add(new Bundle.EntryComponent
{
    Resource = new Patient { Id = "p", BirthDate = "1980-05-17" }
});

var operators = FhirCqlContext.ForBundle(bundle).Operators;
var asOf = new CqlDate(2024, 6, 1);

// Time this loop, or run it under an allocation profiler.
for (var i = 0; i < 200; i++)
    _ = operators.AgeAt(asOf, "year");
```

## Expected Behavior

Everything that does not depend on the operator's arguments is resolved once and reused, so that the cost of an age
operator is the cost of retrieving the patient, reading the birth date and doing the calendar arithmetic.

## Acceptance Criteria

- [x] `FhirTypeResolver.PatientType` resolves the inspector's patient mapping once per resolver instead of on every
      read.
- [x] `FhirTypeResolver.PatientBirthDateProperty` resolves its `PropertyInfo` once instead of on every read.
- [x] The age operators bind the patient retrieve once instead of closing `IDataSource.Retrieve<T>` over the
      patient type and invoking it by reflection on every call.
- [x] `Age` and `AgeAt` share their patient-lookup logic rather than duplicating it.
- [x] The single-patient test no longer buffers the retrieve into an array just to read its length.
- [x] Behaviour is unchanged: the same result for one patient, `null` for zero or more than one, and the same
      `InvalidOperationException` when the type resolver supplies no patient type or birth-date property.
- [x] A benchmark covers the age operators so the cost is visible in future changes.

alexzautke:
### Primary Work

The age operators (`Age`, `AgeAt`, and the `AgeInYears`/`AgeInYearsAt` CQL operators that reach them) redid their
entire setup on every invocation:

1. `TypeResolver.PatientType` → `Inspector.PatientMapping?.NativeType`, which searches the inspector's class
   mappings and allocates while doing so.
2. `TypeResolver.PatientBirthDateProperty` → `typeof(IPatient).GetProperty(...)`.
3. `typeof(IDataSource).GetMethod(nameof(IDataSource.Retrieve)).MakeGenericMethod(patientType)`, then
   `method.Invoke(...)`.
4. `patients.ToArray()`, purely to test whether exactly one patient came back.

None of that depends on the operator's arguments — it is fixed for the lifetime of the operators instance — but
measure logic evaluates these operators per element of a query (`AgeInYearsAt(start of Encounter.period)` inside a
`where` clause), not once per patient. Step 1 alone costs 15.8 µs and 5.8 KB, against under 1.5 µs for the work the
operator actually has to do.

This PR resolves each of those once:

- `FhirTypeResolver.PatientType` resolves the inspector's patient mapping once per resolver instance.
- `FhirTypeResolver.PatientBirthDateProperty` resolves its `PropertyInfo` once.
- `Age` and `AgeAt` — previously near-identical copies of each other — share a single `PatientBirthDate` helper. It
  binds the patient retrieve once through a compiled expression instead of closing `IDataSource.Retrieve<T>` over
  the patient type and invoking it reflectively per call, and counts as it walks instead of buffering the retrieve
  into an array to read its length.

Measured over 200 calls against a one-patient bundle, on .NET 10, with the benchmark added in this PR:

| | Before | After |
| --- | --- | --- |
| `AgeAt` | 2.441 ms / 1.25 MB | 65.4 µs / 100 KB |
| `Age` | 2.297 ms / 1.25 MB | 64.1 µs / 100 KB |

That is ~37× less time and ~13× less allocation, or 12.2 µs → 0.33 µs per call.

No public API changes, no change to the generated C#, so `GeneratorToolVersion` is untouched.

### Auxiliary Work

- `AgeOperatorBenchmarks` — a BenchmarkDotNet benchmark for the two operators, so this cost stays visible.
- `AgeOperatorTests` — characterization tests for the behaviour the refactor has to preserve.
- A release-note fragment under `docs/releases/vnext/`.

---

## Related issues

Closes #1468

Related: #1471 addresses the same shape of problem — per-call setup resolving something that never varies — in
`TypeConverter`. The two are independent and reviewable separately. `claude/perf-combined-testing` carries both, so
the integration corpus can be measured against them together.

## Testing

- `AgeOperatorTests` (12 tests) covers one patient, no patient, more than one patient, a patient without a birth
  date, a null as-of date, repeated calls returning the same age, `Age` agreeing with `AgeAt` as of today, and both
  `InvalidOperationException` paths (a type resolver supplying no patient type, and one supplying no birth-date
  property).
  **All twelve were run against develop's unmodified operators and pass there too** — they characterize existing
  behaviour rather than describing the new implementation, which is what makes them evidence that nothing changed.
- Full `CoreTests` suite green on both target frameworks: 1067 passed / 8 skipped on `net10.0`, and green on
  `net8.0`.
- `Cql-Sdk.slnf` builds with 0 errors.
- Before/after benchmark numbers above were produced by running `AgeOperatorBenchmarks` on this branch and then
  with only the two source files reverted to their develop versions, so the comparison isolates this change.

## FirelyTeam Checklist


alexzautke:
## Integration runner: full CMS corpus

Ran the whole passing CMS corpus — **2851 cases** — against a build carrying this PR *and* #1471
(`claude/perf-combined-testing`), versus `develop`. Both builds green: `Failed: 0, Passed: 2851, Skipped: 1006`.

**Result: no measurable change at corpus level.**

| | CPU (user+sys), median of 10 | sd |
| --- | --- | --- |
| develop | 23.54 s | 1.59 |
| both changes | 23.43 s | 1.37 |

Median delta −0.5%; paired per-round deltas spread from −8.9% to +11.3% (mean +0.6%, sd 6.7). That is noise, not a
result in either direction.

### Why, specifically

**This corpus never reaches the operators this PR changes.** It calls `CalculateAgeAt(birthDate, asOf, precision)` —
98 call sites — which takes the birth date as an argument and performs no patient lookup. There are **zero** calls
to `Age`/`AgeAt`, the operators that did the per-call patient-mapping resolution. So this PR's change is simply not
exercised here. It applies to CQL that uses `AgeInYears()`/`AgeInYearsAt()`, which bind to the retrieve-based
operators.

#1471's change *is* exercised (1145 `Convert` call sites), but its per-call saving is small against ~8.4 ms of CPU
per case, which is dominated by retrieves, bundle deserialization and the query operators.

For context, what this corpus actually spends its operator calls on:

| Operator | Call sites |
| --- | --- |
| `Where` | 2199 |
| `Retrieve` | 1906 |
| `And` | 1655 |
| `Union` | 1190 |
| `Convert` | 1145 |
| `In` | 1024 |
| `Exists` | 927 |

### What this does and doesn't tell us

- It does **not** invalidate the micro-benchmarks. Those measure the operators directly and the improvements are
  real; this corpus just doesn't lean on them.
- It **does** give a strong correctness signal: 2851 real CMS test cases pass unchanged on both builds.
- It suggests the remaining sweep work — materializing retrieves and the list/sort operators — is aimed at where
  this corpus actually spends its time.

### Method

`dotnet test` over the passing set (`ReadAndOnlyRunPreviousPasses`), both variants prebuilt and run alternately
with the order flipped each round to cancel drift. CPU time rather than wall clock: wall was far too noisy here
(per-round swings of −57%..+25%). An always-`before`-first ordering initially showed a spurious +4%; balancing the
order removed it.

Not the skill's BenchmarkDotNet harness: that spawns a process per case, ~7 s each on this 4-core box, so the full
corpus would be ~6 h per side. No new baseline file was added to the submodule, since this run did not use that
harness.


---

alexzautke:
## Correction to my earlier corpus comment

My [earlier comment](https://github.com/FirelyTeam/firely-cql-sdk/pull/1469#issuecomment-5152042137) needs two
corrections. I'm leaving it in place for the audit trail; this comment supersedes it.

**1. The measurement behind "no measurable change at corpus level" was invalid.** The A/B ran inside the xunit
test host, and attributing that run's allocations showed the instrument measuring the test framework rather than
the engine — 554 MB of `Xunit.Sdk` attribute reflection against **zero** attributable to `Hl7.Cql.*`, with
per-round wall-clock swinging −57%..+25%. An engine-level change of several percent was invisible by
construction. The corrected instrument (in-process harness over the same generated cases, 900 cases, sequential,
median of 12 reps, deterministic allocation) resolves branch differences cleanly.

**2. The earlier comment's aside about #1471 was factually wrong.** It said #1471's saving "is small against
~8.4 ms of CPU per case." On the corrected instrument, #1471 alone is **−8.6% CPU** on this corpus (1420 →
1298 ms), corroborated by a profile showing the attribute reflection it removes at ~16% of self CPU inside
measure execution. I've posted the same correction
[on #1471](https://github.com/FirelyTeam/firely-cql-sdk/pull/1471#issuecomment-5153075193).

**What still stands for *this* PR:** the call-site analysis, which does not depend on the broken timing. This
corpus contains **zero** calls to `Age`/`AgeAt` — its measures use `CalculateAgeAt(birthDate, …)` (98 sites),
which takes the birth date as an argument and never performs the patient lookup this PR optimizes. So this PR's
gain is real but shows in the added `AgeOperatorBenchmarks` (2.44 ms → 65 µs per 200 calls) and in CQL that uses
`AgeInYears()`/`AgeInYearsAt()`, not in this corpus. The 2851/2851 correctness signal also stands. On the
corrected instrument, this PR combined with #1471 measures −10.1% vs −8.6% for #1471 alone — consistent with a
small-but-real contribution here, though that 1.5-point difference is near the instrument's resolution.


---

baseTwo (changes requested):
## Review of `15b26cde6` — the diagnosis is right and most of the win is already banked, but the delegate is cached at the wrong scope

All seven acceptance criteria on #1468 are met, the behaviour-preservation work is careful, and the test coverage is the best of this batch so far. One finding is substantive: the part of the change that compiles an expression is cached per `CqlOperators` instance, and on the call pattern this SDK's own hosts use, that is a regression rather than a win.

### Verified independently

- **The null-birth-date short-circuit is behaviour-preserving**, which is the least obvious part of the refactor. `Age`/`AgeAt` used to call `CalculateAge(null, …)` when the patient existed but had no birth date; they now return `null` without calling it. Those agree only because `CalculateAgeAt` returns `null` for a null `birthDate` — checked, it does, at the top of the method.
- **The counting loop is equivalent to the old `ToArray().Length == 1`**, and returns as soon as a second patient appears, which is a strict improvement for a lazy retrieve rather than just an allocation saving.
- **`Expression.Convert(call, typeof(IEnumerable<object>))` is sound**, and specifically because `IDataSource.Retrieve<T>` constrains `T : class`: covariance gives the reference conversion for free. It would be invalid for a value type, which the constraint forbids — worth keeping in mind if that constraint ever loosens.
- **The exception contract is preserved**, same order (`PatientType`, then `PatientBirthDateProperty`) and same type, and both paths have tests asserting the parameter name appears in the message.
- **`Lazy<Type?>` is thread-safe in its default mode**, and the fact that it caches a `null` result forever is consistent with what the resolver already does — `AddTypesFromInspector()` runs in the constructor, so this resolver has never been able to see mappings imported later.
- **The `_retrievePatients ??=` race is benign**: two threads may both build, the delegates are equivalent, and the reference write is atomic. Worth noting explicitly because #1453 shipped a concurrent evaluation-cache profile, so a shared context is a supported configuration.
- 12 tests covering one patient, zero, more than one, no birth date, null `asOf`, `Age` versus `AgeAt` agreement, repeated calls, non-interference, and both exception cases.

### 1. The compiled delegate is cached per operators instance, and on the one-call-per-context pattern that is ~120× slower than what it replaced

`_retrievePatients` is an instance field, so `Expression.Compile()` runs once per `CqlOperators` instance — that is, once per `CqlContext`, once per `FhirCqlContext.ForBundle(...)`.

I measured the two sides on this exact lambda shape (a generic `Retrieve` closed by `MakeGenericMethod`, wrapped in a `Convert`):

| | Cost |
| --- | --- |
| `Expression.Lambda(…).Compile()` | **42.6 µs**, once per instance |
| `MakeGenericMethod` + `Invoke` (what it replaces) | **0.36 µs**, per call |

Break-even is around **120 calls per operators instance**. Note what this implies about where the win in #1468 actually came from: the issue measured 13.4 µs per call, and its own breakdown attributes about 13 µs of that to `ModelInspector.PatientMapping`. **The `Lazy` on `PatientType` banks essentially all of that**, independently of this delegate. The reflective dispatch this finding is about was only 0.36 µs of the 13.4.

The problem is the call pattern. #1466 describes this SDK's own host as evaluating "each subject's `$everything` bundle once per measure group (~10 groups for HEDIS measures), each evaluation via `FhirCqlContext.ForBundle(...)`" — a fresh context, and therefore a fresh `CqlOperators`, per group per subject. A measure that uses `AgeInYears()` in a patient-level define calls the operator **once per context**. For 1,000 subjects × 10 groups that is 10,000 compiles ≈ **426 ms of pure `Expression.Compile`**, against ≈ 3.6 ms of reflective dispatch before this change.

The fix is small and has precedent in this same review batch: **cache the delegate statically, keyed by patient type.**

```csharp
private static readonly ConcurrentDictionary<Type, Func<IDataSource, IEnumerable<object>?>> PatientRetrievers = new();
...
var patients = PatientRetrievers.GetOrAdd(patientType, static t => BuildPatientRetriever(t))(DataSource);
```

That is trivially safe here precisely because of a decision already made in this PR: the lambda takes `IDataSource` as a **parameter** rather than closing over it, so nothing context-specific is captured and only the patient type varies. It makes the 42.6 µs a once-per-process cost, and then both call patterns win — the per-element case keeps everything it gained, and the once-per-context case stops paying for a compile it cannot amortise. #1458 and #1467 both landed static caches keyed on exactly the things that vary, for the same reason.

### 2. The benchmark cannot show that trade-off, by construction

`[GlobalSetup]` builds the context once, outside the measured region, and `[Params(200)]` is the only call count. So the compile happens on the first benchmark invocation and is amortised across every iteration — the number that would have exposed finding 1 is the one the benchmark is shaped to hide, and 200 calls is comfortably past break-even.

Please add `Calls = 1` (and something like 10) to the `[Params]`, and a benchmark whose measured region includes constructing the context, so per-context setup cost is visible. As it stands the benchmark answers "is the per-element path fast" but not "did we make the per-context path slower", which is the question this change actually raises.

### 3. The release fragment's `## Performance` is not a section that exists anywhere

`docs/releases/vnext/1469-age-operator-patient-lookup.md` uses `## Performance`. The template (`docs/releases/release-notes-template.md`) defines `New Public API`, `Improvements`, `Dependency Updates`, `Potentially Breaking`, `Breaking` and `Fixes`; no released notes document has ever had a `Performance` section; and the two fragments currently sitting in `vnext/` use `## Features` and `## Breaking changes`. At cut time the fragments are folded in by their `##` heading, so an unrecognised one has no home to fold into.

Please use **`## Improvements`**, which is where 2.12.0 put work of exactly this kind. Same class of defect as the missing `#### Dependency Updates` section on #1504 — a document that does not match its own template.

### 4. Restate the fragment's numbers once finding 1 is settled

"`AgeAt` went from 2.441 ms and 1.25 MB to 65.4 µs and 100 KB" is measured over 200 calls in one context and excludes the one-off compile. Once the delegate is cached statically that caveat disappears, so the simplest fix is to re-measure after finding 1 rather than to annotate it.

baseTwo:
@alexzautke @copilot — the three changes from my review, spelled out so they can be applied directly.

**1. Cache the retriever statically, keyed by patient type** (`Cql/Cql.Runtime/Operators/CqlOperators.ClinicalOperators2.cs`).

Replace the instance field with a static map, and make the builder take the type as a parameter so it can be a `static` lambda:

```csharp
// before
private Func<IDataSource, IEnumerable<object>?>? _retrievePatients;
...
var patients = (_retrievePatients ??= BuildPatientRetriever())(DataSource);

// after
private static readonly ConcurrentDictionary<Type, Func<IDataSource, IEnumerable<object>?>> PatientRetrievers = new();
...
var patients = PatientRetrievers.GetOrAdd(patientType, static t => BuildPatientRetriever(t))(DataSource);
```

`BuildPatientRetriever` then takes `Type patientType` instead of reading `TypeResolver.PatientType` itself; keep the existing null-check on `TypeResolver.PatientType` in `PatientBirthDate()` so the `InvalidOperationException` contract and its ordering are untouched, and pass the non-null value in.

**Why:** `Expression.Compile()` measures at **42.6 µs** on this lambda shape, against **0.36 µs** per call for the `MakeGenericMethod` + `Invoke` it replaces — break-even is about 120 calls *per operators instance*. Hosts build a context per subject per measure group (#1466 describes ~10 per subject), so a patient-level `AgeInYears()` pays the compile once per context and never reaches break-even. This is safe to make static precisely because your lambda takes `IDataSource` as a parameter rather than capturing it, so only the type varies.

**2. Make the benchmark able to show it** (`Cql/Benchmarks/AgeOperatorBenchmarks.cs`). `[GlobalSetup]` builds the context outside the measured region and `[Params(200)]` is the only call count, so the compile is amortised into invisibility. Add `Calls = 1` and `10` to the `[Params]`, and add one benchmark whose measured region constructs the context, so per-context setup cost shows up.

**3. Fix the fragment's section heading** (`docs/releases/vnext/1469-age-operator-patient-lookup.md`): `## Performance` → `## Improvements`. The template defines `New Public API`, `Improvements`, `Dependency Updates`, `Potentially Breaking`, `Breaking` and `Fixes`; no released notes document has a `Performance` section, and the other two fragments in `vnext/` use `## Features` and `## Breaking changes`. Fragments fold in by their `##` heading at cut time, so an unrecognised one has nowhere to go.

Once 1 is in, please re-measure and update the fragment's figures rather than annotating them — the "2.441 ms → 65.4 µs" numbers currently exclude the one-off compile.

Nothing else is outstanding: the null-birth-date equivalence, the exception contract, the covariance in `Expression.Convert`, the `Lazy` semantics and the benign `??=` race all check out, and the 12 tests cover every acceptance criterion on #1468.

baseTwo:
@copilot Findings 1-3 are all verified in `190456463` + `2a7241595` — the retriever is now a `static readonly ConcurrentDictionary<Type, …>` with `GetOrAdd` and a static builder, the benchmark gained `[Params(1, 10, 200)]` plus a "context setup + query elements" case that puts construction inside the measured region, and the fragment heading is `## Improvements`. That is exactly what was asked; nothing outstanding from them.

**One more required change**, from a convention just recorded in #1511: **this repo's reflection goes through `ReflectionUtility`, not string-based lookups.** Both of this PR's reflection sites qualify.

**1. `Cql/Cql.Runtime/Operators/CqlOperators.ClinicalOperators2.cs`** — in `BuildPatientRetriever`:

```csharp
// before
var retrieve = typeof(IDataSource)
    .GetMethod(nameof(IDataSource.Retrieve))!
    .MakeGenericMethod(patientType);

// after
private static readonly MethodInfo RetrieveDefinition =
    ReflectionUtility.GenericMethodDefinitionOf(() => default(IDataSource)!.Retrieve<object>(default));
...
var retrieve = RetrieveDefinition.MakeGenericMethod(patientType);
```

**2. `Cql/Cql.Firely/FhirTypeResolver.cs`** — the birth-date property:

```csharp
// before
private static readonly PropertyInfo? BirthDateProperty = typeof(IPatient).GetProperty(nameof(IPatient.BirthDate));

// after
private static readonly PropertyInfo BirthDateProperty =
    ReflectionUtility.PropertyOf(() => default(IPatient)!.BirthDate);
```

`PropertyOf` returns non-null or throws, so the field stops being nullable. **Keep `PatientBirthDateProperty`'s declared type as `PropertyInfo?`** — the base member is nullable and other resolvers legitimately return null, and the age operators' `InvalidOperationException` contract depends on that. Only this resolver's own field becomes non-null.

Two things already checked so you do not have to:

- **`using Hl7.Cql.Compiler.Infrastructure;` is required in both files.** That namespace is not a global using in `Cql.Runtime` or `Cql.Firely` — I checked both `GlobalUsings.cs`.
- **`GenericMethodDefinitionOf` takes an `Expression<Action>`, and a value-returning call binds to it.** I verified by compiling `Expression<Action> e = () => default(IDS).Retrieve<object>(default);`, so `Retrieve<T>` returning `IEnumerable<T>` is not a problem. `object` satisfies its `where T : class` constraint. Follow the existing shape at `Cql/CoreTests/Abstractions/CSharpFormatterTests.cs:130`.

**Why this and not `nameof`:** the current code already uses `nameof`, which survives a rename but still resolves by string at run time and still cannot notice an added overload or a changed signature — and it forces the `!` that the expression form makes unnecessary.

Nothing else is outstanding on this PR. Once this is in and the eight checks are green, I will re-review and approve.

baseTwo:
@copilot One more small change in `Cql/Cql.Firely/FhirTypeResolver.cs`, on top of the `ReflectionUtility` work — both of these members are `internal` but nothing outside the class uses them.

**1. `Inspector` → `private`** (line 128):

```csharp
// before
internal ModelInspector Inspector { get; }

// after
private ModelInspector Inspector { get; }
```

**2. `TypeSpecifiers` → `private`** (line 130):

```csharp
// before
internal IDictionary<Type, string> TypeSpecifiers { get; } = new Dictionary<Type, string>();

// after
private IDictionary<Type, string> TypeSpecifiers { get; } = new Dictionary<Type, string>();
```

**Evidence that this is safe:** I searched the whole repository for both names with word boundaries, including `CoreTests` and the other assemblies that `Cql.Firely` grants `InternalsVisibleTo` to, and the Integration Runner submodule. There are **no** references outside this file. The only hits for `TypeSpecifiers` elsewhere are `resultTypeSpecifiers`, an unrelated ELM concept. `Inspector`'s five in-file uses (the `_patientType` lambda, `ModelAssemblies`, `FindClassMapping` in `adjust()`, and twice in `AddTypesFromInspector`) are unaffected by the narrower accessibility.

---

*Edited: this comment also asked whether `TypeSpecifiers` — which is written twice and never read — should be deleted rather than hidden. That question is now **#1515**, @alexzautke has confirmed the deletion, and it is being done separately against `develop` so this PR stays scoped to the age operators. Narrowing it to `private` here remains correct either way.*

baseTwo:
> * FhirTypeResolver

@copilot , FhirTypeResolver.Inspector and FhirTypeResolver.TypeSpecifiers can be changed to fields

baseTwo:
## Final review of `e776cd918` — everything raised is applied; approval held only on the run in flight

Six commits since my review, all verified against the code rather than the summaries. The diff against `develop` is still exactly the five intended files, so the `develop` merge in `300010906` brought nothing along with it.

### The three review findings

- **Static retriever cache** (`190456463`). `PatientRetrievers.GetOrAdd(patientType, static type => BuildPatientRetriever(type))` — a `static readonly ConcurrentDictionary<Type, …>`, so the `Expression.Compile()` I measured at **42.6 µs** is now paid once per process rather than once per `CqlOperators` instance. That was the finding that mattered: against the **0.36 µs** per-call `MakeGenericMethod` + `Invoke` it replaces, break-even was ~120 calls per instance, and a host building a context per subject per measure group never reached it. Both call patterns now win. The comment naming the low-cardinality assumption is a good addition — it is the thing that makes an unbounded static dictionary safe, and it is now stated rather than implied.
- **Benchmark** (`190456463`). `[Params(1, 10, 200)]` plus a third case whose measured region includes constructing the context. `Calls = 1` is the one that would have shown the regression, and it exists now.
- **Fragment** (`190456463`). `## Improvements`, which the template defines.

### `ReflectionUtility` (`a38897c33`, `5905d3fbd`)

Both sites converted, and the generic definition is hoisted to a `static readonly MethodInfo` rather than resolved per build — a step past what I asked for:

```csharp
private static readonly MethodInfo RetrieveDefinition =
    ReflectionUtility.GenericMethodDefinitionOf(() => default(IDataSource)!.Retrieve<object>(default));
```

`BirthDateProperty` is now `PropertyOf(() => default(IPatient)!.BirthDate)` and its field type dropped from `PropertyInfo?` to `PropertyInfo`, with a comment recording that the nullable *override* is inherited from the base — exactly the distinction I asked to preserve, since other resolvers legitimately return null and the operators' `InvalidOperationException` contract depends on it. `using Hl7.Cql.Compiler.Infrastructure;` added to both files.

One consequence worth noting as an improvement rather than a defect: `BuildPatientRetriever` is now `static` and takes the patient type, so the duplicated null-check it used to carry is gone — while `PatientBirthDate()` still checks `PatientType` first and `PatientBirthDateProperty` second, preserving the original exception order and messages. The two tests asserting on those names still pass unchanged.

### Privacy (`e2458dbe0`, `e776cd918`)

`Inspector` and `TypeSpecifiers` are now `private readonly` fields, `_inspector` and `_typeSpecifiers`, matching the file's existing `_patientType` convention — a cleaner result than the property-visibility change I proposed. All five `_inspector` uses are intact.

The follow-up question about `_typeSpecifiers` being written and never read is now **#1515**, confirmed by @alexzautke and being done separately against `develop`, so it does not hold this PR up.

### Verdict

Nothing outstanding. All seven acceptance criteria on #1468 remain satisfied, and the two design points raised during review — hoisting the patient-lookup memoization into the abstraction, and narrowing `CqlOperators` onto it — are **#1512**, deliberately stacked on this PR rather than folded into it, since they reach into four assemblies and #1468 promised only the age operators.

`Build, Sign & Package` is still running on `e776cd918`. I will approve as soon as the eight checks are green — the merge from `develop` means this branch has not been fully exercised in CI at this SHA yet, and given it now carries #1458's date/time changes as well, that run is worth waiting for rather than approving on the strength of the diff.

baseTwo (approved):
## Approving

All checks green at this SHA, which is the run that mattered: the branch took a `develop` merge in `300010906` that brought #1458's date/time work with it, so this is the first time the whole set — `Build, Sign & Package` with the public-API analyzer, `Test on .NET 8` and `.NET 10`, HEDIS Golden Parity, and Integration Tests — has exercised the code at this commit.

Everything from the review is applied and verified against the code:

- **The retriever cache is static and keyed by patient type**, so the `Expression.Compile()` I measured at 42.6 µs is paid once per process rather than once per `CqlOperators` instance. That was the finding that mattered — against the 0.36 µs per-call reflection it replaced, break-even was roughly 120 calls per instance, and a host building a context per subject per measure group never got there.
- **The benchmark can now show that trade-off** (`[Params(1, 10, 200)]` plus a case whose measured region includes constructing the context), where before it was shaped to hide it.
- **Both reflection sites go through `ReflectionUtility`**, with the generic definition hoisted to a `static readonly MethodInfo` so it resolves once — a step past what was asked.
- **`Inspector` and `TypeSpecifiers` are private fields** matching the file's existing convention.
- The exception contract is intact: same order, same messages, and the two tests asserting on those member names still pass unchanged.

Two follow-ups are tracked rather than folded in, deliberately: **#1515** for the write-only `_typeSpecifiers` map (confirmed for deletion, being done separately against `develop`), and **#1512** for extracting `PatientTypeInfo` and narrowing the age operators onto it, stacked on this PR since it reaches into four assemblies and #1468 promised only the age operators.

Merging is yours to call.

PR #1475, opened 2026-08-01
Linear merge walk in Collapse

### LINKED ISSUE #1473: Collapse: quadratic RemoveAt(0) merge loop and a dead count fast-path in CollapseHelper
## Summary

`CollapseHelper<T>` in [`CqlOperators.IntervalOperators.cs`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/Cql/Cql.Runtime/Operators/CqlOperators.IntervalOperators.cs) backs all nine `Collapse` overloads (`Int32`, `Int64`, `Decimal`, `CqlQuantity`, `CqlDate`, `CqlDateTime`, `CqlTime`, …). It produces correct results, but the merge loop does quadratically more work than the operation needs, and the emptiness pre-check contains a dead fast-path. This is a behavior-preserving refactor — no semantic change.

## Impact

`Collapse` runs over interval lists built from patient data (encounter periods, medication periods), so its cost scales with the size of a patient's history. The merge loop is quadratic in the number of intervals: measured on disjoint intervals (Release build), develop takes 229 ms at n=25k and 4,794 ms at n=200k (superlinear), where a linear walk takes 190 ms and 536 ms respectively (~8.9× at n=200k).

## Observed Behavior

Two problems, both in `CollapseHelper<T>`:

1. **Quadratic drain of the sorted list.** The merge loop takes intervals off the *front* of a `List<>`:

   ```csharp
   while (queue.Count > 0)
   {
       var firstInQueue = queue[0];
       queue.RemoveAt(0);   // shifts every remaining element down one slot
       ...
   }
   ```

   `List<T>.RemoveAt(0)` copies all remaining elements down one position, so collapsing *n* intervals costs O(n²) element copies — for a list that is only ever read front to back.

2. **Dead count fast-path.** The emptiness pre-check tries to special-case materialized inputs:

   ```csharp
   if (intervals is IList<T> list)        // T is the POINT type (e.g. CqlDateTime?),
       count = list.Count;                // but intervals holds CqlInterval<T?>? elements
   else if (intervals is T[] array)
       count = array.Length;
   else count = intervals.Count();
   ```

   The type tests ask about `IList<T>`/`T[]` — the interval *point* type — while the sequence element type is `CqlInterval<T?>?`. They can never match, so every call falls through to `Count()`. Note this is a **dead-code/clarity problem, not a measurable performance problem**: `Enumerable.Count()` has the same `ICollection<T>` fast path, so materialized lists and arrays were never actually walked — only genuinely lazy sources are enumerated, exactly as they would be by any counting approach. (An earlier version of this issue claimed a full walk for lists and arrays; instrumented enumeration counting disproved that.)

## Root Cause

- The drain loop models the sorted input as a consumable queue when an index walk over the same list gives identical visit order without any element shifting.
- The count fast-path type-tests the wrong type argument, making it unreachable — misleading, though functionally harmless.

## Expected Behavior

- Identical results for every input (same sort, same `Meets`/`Overlaps`/`Includes`-driven merge logic, same `null`/empty returns).
- The merge phase is O(n) after the sort: walk the sorted list front to back, merging into the last element of the result list.
- The count probe uses `Enumerable.TryGetNonEnumeratedCount` (or equivalent), removing the dead type tests.

## Acceptance Criteria

- [x] The merge loop no longer uses `queue.RemoveAt(0)`; it walks the sorted list without shifting elements, and the result list is pre-sized to the input count.
- [x] The emptiness pre-check no longer type-tests `IList<T>`/`T[]` against the point type; it uses `TryGetNonEnumeratedCount` with `Count()` only as the fallback for lazy sources.
- [x] Behavior is unchanged: unit tests cover the merge semantics (overlapping intervals merge, meeting intervals merge, a chain of 3+ intervals merges into one — the spec's own example `collapse { Interval[1,4], Interval[4,8], Interval[7,9] }` → `{ Interval[1,9] }` — two separate merge groups in one result, an interval contained in the previous one is absorbed, disjoint intervals stay separate, unsorted input is sorted first, null elements and open boundaries, empty input, null input) for at least one numeric and one date/time overload — and these tests pass on `develop` too, locking the behavior rather than proving a bug.
- [x] `ICqlOperators` is untouched (no signature changes, no `GeneratorToolVersion` bump needed).
- [x] A release-note fragment is added under `docs/releases/vnext/` that does not claim a performance win for the count-probe cleanup.

alexzautke:
## Description
Behavior-preserving refactor of `CollapseHelper<T>` in `Cql/Cql.Runtime/Operators/CqlOperators.IntervalOperators.cs`, which backs all nine `Collapse` overloads:

- **Merge loop, quadratic → linear.** The merge phase drained the sorted list with `queue.RemoveAt(0)`, shifting every remaining element down one slot per interval — O(n²) element copies for a list only ever read front to back. It now walks the sorted list front to back, merging into the interval most recently added to the result (`result[^1]`), with the result list pre-sized. Same visit order, same `Meets`/`Overlaps`/`Includes` merge logic, same null/empty returns. Measured on disjoint intervals (Release): develop goes 229 ms → 4,794 ms from 25k → 200k intervals (superlinear); the new code goes 190 ms → 536 ms (essentially linear), ~8.9× faster at 200k.
- **Dead count fast-path removed.** The emptiness pre-check type-tested `intervals is IList<T>`/`T[]` — `T` being the interval *point* type (e.g. `CqlDateTime?`), not the `CqlInterval<T?>?` element type the sequence holds — so the tests could never match. Replaced with `Enumerable.TryGetNonEnumeratedCount`. This is a readability fix with **no performance effect**: the `Count()` fallback the dead tests always reached already short-circuits on `ICollection<T>`, so materialized inputs were never walked (verified by instrumented enumeration counting).

Zero behavior change was verified three ways: reasoning over both loops (the containment case where `TryCombine` returns its first argument is a self-assignment in both versions), a 424-case differential harness (hand-picked structural cases plus 400 seeded-random sets) producing byte-identical output under both implementations, and the behavior-lock test runs below.

No `ICqlOperators` signature changes — no `GeneratorToolVersion` bump, no `*.g.cs` regeneration. A release-note fragment is included under `docs/releases/vnext/`.

## Related issues
Closes #1473.

## Testing
New `Cql/CoreTests/CollapseOperatorTests.cs` with 23 behavior-lock tests over the integer and `CqlDateTime` overloads: overlap merge, meets merge, chained merge of three intervals (the spec's own §9.B example `collapse { Interval[1,4], Interval[4,8], Interval[7,9] }` → `{ Interval[1,9] }`), a chained merge extending across an absorbed interval (`{ [1,10], [3,5], [8,20] }` → `{ [1,20] }` — specifically kills the compare-against-previous-input mutant), two separate merge groups in one result, containment absorption, disjoint intervals, unsorted input, open boundaries, null elements, lazy sources (exercising the `Count()` fallback), empty input, and null input.

Because this PR claims no behavior change, the proof is the opposite of a regression test: **all 23 tests pass against develop's implementation and against this PR's implementation alike** (verified by swapping the operator file, rebuilding, and re-running). The null-element tests' doc comments note that keeping nulls in the result deviates from §9.B (which excludes them) — that is pre-existing behavior on develop, locked rather than changed here, and will get a follow-up issue.

Full CoreTests on net10.0 (Release): 1078 passed, 8 skipped, 0 failed (1086 total).

## FirelyTeam Checklist

---


baseTwo (commented):
## Review of `15cbffa86` — behaviour-preservation verified against `develop`, not just asserted

@alexzautke — 0 behind `develop`, so nothing stale here.

### The refactor is faithful

Lined the two loops up statement by statement. The old one took `queue[0]`, removed it, and then branched on
`result.Count > 0` to either append or merge into `result[^1]`; the new one iterates `sorted`, short-circuits the
first element, and otherwise does the same `TryCombine(result[^1], next)` with the same append-or-replace. Same
visit order, same merge target, same branch structure. `TryCombine`, the sort, and the null handling are untouched,
so the ordering of null elements — and the fact that they survive into the result, which is
[#1476](https://github.com/FirelyTeam/firely-cql-sdk/issues/1476) — is deliberately unchanged.

`RemoveAt(0)` on a `List<T>` is documented O(n), so draining n intervals was O(n²) element copies for a list only
ever read front to back. Removing it is structurally, not just empirically, the fix. Result list pre-sized to
`sorted.Count`.

### Verified rather than assumed

**I ran your 23 tests against `develop`'s implementation**, by checking the test file out onto `develop` and leaving
`CollapseHelper` as it is there — `RemoveAt(0)` and all:

```
Passed!  - Failed: 0, Passed: 23, Skipped: 0, Total: 23
```

and against this branch:

```
Passed!  - Failed: 0, Passed: 23, Skipped: 0, Total: 23
```

That is the right shape for a performance-only change: the tests lock the behaviour that already exists rather than
describing new behaviour, so they would catch a regression introduced by the rewrite. Coverage is genuinely broad —
both a numeric and a date/time overload, the spec's own `{[1,4],[4,8],[7,9]} → {[1,9]}` chain, absorption, two
separate merge groups, unsorted input, open boundaries, null elements, lazy sources, empty and null input.

### The count-probe change is described accurately

The old type tests asked about `IList<T>`/`T[]` — the interval *point* type — against a sequence of
`CqlInterval<T?>?`, so they could never match. Replacing them with `TryGetNonEnumeratedCount` is behaviour-neutral,
and both the code comment and the fragment say plainly that this is readability with no performance effect,
because the `Count()` fallback already short-circuits on `ICollection<T>`. I appreciate the fragment explicitly
*not* claiming a win here, and the issue carrying the correction that an earlier version of it overstated this.

The empty-versus-null distinction still holds: `count == 0` returns an empty array, while a null sort result
returns `null`. Worth keeping in mind that the probe is load-bearing for exactly that, so it cannot simply be
dropped in favour of sorting first.

### 1. Note, no action: a lazy source is still enumerated twice

`TryGetNonEnumeratedCount` fails for a lazy sequence, so `Count()` walks it, and `SortBy` then walks it again. That
is exactly what `develop` does, so it is not a regression and not something I want changed here — but it is the
same double-enumeration that [#1477](https://github.com/FirelyTeam/firely-cql-sdk/issues/1477) is about for
retrieve results, and `Collapse_LazySource_IsCollapsed` currently pins the two-pass behaviour. Worth a cross-link
so whoever takes #1477 knows this call site exists.

### 2. Note: `Collapse_NullElement_IsSortedFirstAndKeptSeparate` will need updating by #1476

That test asserts nulls survive into the result, which is the behaviour #1476 says the spec forbids. Correct to pin
it *now* — it is what `develop` does — but it is a behaviour lock that #1476 must deliberately flip, not a test
that should quietly keep passing. Worth a comment pointing at #1476 so the next person does not read a failing
assertion there as their own bug.

---

Both items are notes; nothing blocks. Approving once the checks finish — `Build, Sign & Package` and the test jobs
were still running at this head. Note this will still need a second approver to merge.

baseTwo (approved):
## Approving `15cbffa86` — all 8 checks green

@alexzautke — as flagged in my review, holding only for CI, which is now green across all 8 checks.

Recapping what was verified rather than assumed: I ran your 23 tests against **`develop`'s** implementation, with
`RemoveAt(0)` still in place — 23/23 pass there and here. That is the right shape for a performance-only change:
the tests lock behaviour that already exists, so they would catch a regression introduced by the rewrite rather
than merely describing the new code.

The refactor is faithful statement for statement, `TryCombine` and the null handling are untouched, and
`RemoveAt(0)` being documented O(n) makes the quadratic-to-linear claim structural rather than merely measured.

The two notes stand as notes, no action wanted here: lazy sources are still enumerated twice (unchanged from
`develop`, and the same double-enumeration #1477 addresses), and `Collapse_NullElement_IsSortedFirstAndKeptSeparate`
pins behaviour that #1476 will have to deliberately flip.

PR #1481, opened 2026-08-01
Compute the resource-type list once per process

### LINKED ISSUE #1479: CompareResourcesById enumerates every type in the model assembly on each call
## Summary

`CqlComparersExtensions.CompareResourcesById` ([`Cql/Cql.Firely/Extensions/CqlComparersExtensions.cs`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/Cql/Cql.Firely/Extensions/CqlComparersExtensions.cs)) recomputes the set of resource types from scratch on every call:

```csharp
var derivedFromResource = typeof(Patient).Assembly.GetTypes()
    .Where(t => typeof(Resource).IsAssignableFrom(t));
```

`Assembly.GetTypes()` materializes every type in the Firely model assembly (thousands) and filters for `Resource` descendants. That answer is fixed for the lifetime of the process, but the method runs per comparer setup — and comparer setup happens per `CqlContext`, which is per patient. A batch evaluation over N patients pays the full reflection sweep N times.

A working fix exists on the `claude/sdk-cql-performance-j3uz16` branch; this ticket tracks landing it.

## Impact

In thread-time profiles of a 900-measure CMS evaluation run, this reflection sweep accounted for roughly 5% of steady-state CPU.

## Root Cause

Fixed program metadata (the model assembly's resource types) computed eagerly inside a per-context setup path instead of once per process.

## Expected Behavior

- The resource-type list is computed once per process (`static readonly Lazy<Type[]>`) and reused; `CompareResourcesById` only performs the `Register` loop per call.
- Registration behavior is unchanged: the same types get the same `ResourceIdCqlComparer` with the caller's `StringComparer`.

## Acceptance Criteria

- [x] `CompareResourcesById` no longer calls `Assembly.GetTypes()` per invocation; the type sweep is behind a process-wide lazy computed at most once.
- [x] Behavior-lock tests cover the registration semantics — resources of several types (including at least one beyond `Patient`) compare equal/unequal by id after opting in, the supplied `StringComparer` is honored (e.g. case-insensitive ids), and calling `CompareResourcesById` more than once works — all passing on `develop` too (this change has no observable behavior).
- [x] No public API changes; `ICqlOperators` untouched (no `GeneratorToolVersion` bump).
- [x] A release-note fragment under `docs/releases/vnext/` (`## Improvements`) citing only the measured profile observation.

alexzautke:
## Description
`CqlComparersExtensions.CompareResourcesById` recomputed the set of resource types on every call:

```csharp
var derivedFromResource = typeof(Patient).Assembly.GetTypes()
    .Where(t =&gt; typeof(Resource).IsAssignableFrom(t));
```

`Assembly.GetTypes()` materializes every type in the Firely model assembly and filters for `Resource` descendants — an answer that is fixed for the process lifetime, recomputed per comparer setup, which happens per `CqlContext`, i.e. per patient in batch evaluations. In thread-time profiles of a 900-measure CMS evaluation run this sweep accounted for roughly 5% of steady-state CPU.

The sweep now lives in a `private static readonly Lazy` (thread-safe by default), computed at most once per process; `CompareResourcesById` only performs the `Register` loop per call. Registration behavior is unchanged — same types, same `ResourceIdCqlComparer` built per call with the caller&#39;s `StringComparer`. The one behavioral difference is on the failure path: `Lazy`'s default `ExecutionAndPublication` mode caches an exception from the value factory, so a partially loadable model assembly makes the `ReflectionTypeLoadException` permanent where the per-call sweep re-attempted it. That is recorded in the release-note fragment. No public API changes; no `ICqlOperators`/`GeneratorToolVersion` impact.

## Related issues
Closes #1479.

## Testing
`CompareResourcesById` previously had **zero test coverage** (its only references were the definition and one call site in `FhirCqlContext`). New `Cql/CoreTests/Fhir/CompareResourcesByIdTests.cs` adds 7 behavior-lock tests: Patient and Observation equal/unequal by id under `Ordinal` (with other properties deliberately differing, pinning that only the id is compared); a test-local `DerivedPatient : Patient` proving resolution via base-type registration; `Ordinal` vs `OrdinalIgnoreCase` deciding case-differing ids across two resource types; independent registrations on three fresh `CqlComparers` instances; last-registration-wins on double opt-in **for the types the sweep registers**; and its counterpart, pinning that a type already resolved through its base type keeps the comparer it first resolved to, because `CqlComparers.SelectComparer` memoizes that resolution with `TryAdd`.

This change has no observable behavior, so there is no fail-first proof — all 7 tests pass against develop&#39;s per-call implementation and against this PR&#39;s implementation alike (verified by swapping the file, rebuilding, and re-running).

What the suite does and does not catch is stated in the test class remarks rather than implied: it fails on a type list that loses types a runtime instance can have, and on a comparer that stops being built per call from the caller's `StringComparer`. It is deliberately blind to changes that leave every observable registration identical — restricting the sweep to concrete types, or moving the type list to a per-instance cache.

Full CoreTests on net10.0 (Release), with `develop` merged in: 1174 passed, 8 skipped, 0 failed (1182 total).

## FirelyTeam Checklist

---


baseTwo (commented):
## Review of `1fb2b6c55` — good, small things only

**Note on staleness:** this branch is still **90 commits behind `develop`** (the others in this batch were updated). Per repo convention the base merge is your call, not mine — flagging it so it is a decision. CI green here predates most of what has landed.

The change is correct and genuinely behaviour-preserving — I verified all 6 new tests pass against develop's implementation and that they do catch the real hazard (over-hoisting the comparer into the static) — but the test class's own `<remarks>` overstates its coverage: I proved by mutation that restricting the sweep to concrete types breaks nothing.

### Verified

- The optimization is structurally sound and the cache cannot be mis-keyed. `typeof(Patient)` binds at compile time, so `Cql.Firely` resolves it to exactly one `Hl7.Fhir.Model` assembly per AssemblyLoadContext; static fields are also per-ALC. There is no runtime-varying key that the static could serve from the wrong assembly. Under the old code the answer was equally fixed, so hoisting it changes nothing.
- The right thing was hoisted and the wrong thing was not. Only `Type[]` (immutable, process-global validity) moves into the static; the `ResourceIdCqlComparer`, which wraps the caller's `StringComparer`, is still constructed per call inside `CompareResourcesById` (CqlComparersExtensions.cs:96). Mutation test C confirms the suite fails loudly if someone later hoists the comparer too.
- The per-call cost claim is real. `FhirCqlContextOptions.ResourceIdComparer` defaults to `StringComparer.OrdinalIgnoreCase` (Cql/Cql.Firely/FhirCqlContextOptions.cs:32), i.e. non-null by default, so the guarded call at Cql/Cql.Firely/FhirCqlContext.cs:77-78 fires on essentially every `FhirCqlContext` creation. The release note's "ran per `CqlContext`, i.e. per patient in batch evaluations" is accurate.
- `Lazy<Type[]>` created via `new(Func<Type[]>)` uses `LazyThreadSafetyMode.ExecutionAndPublication`, so the PR description's "thread-safe by default" is correct; the factory runs at most once even under concurrent context creation.
- Deferral is correct and `Lazy` is the right tool rather than a plain `static readonly` field. `CqlComparersExtensions` is a static class whose other member `AddFhirComparers` is called unconditionally; a plain field initializer could be triggered by `beforefieldinit` type initialization on that path, making callers who never opt into by-id comparison pay the sweep. `Lazy` ties the cost to actual use.
- The `DerivedPatient` test is not decorative — it exercises a distinct code path. `EqualsValues` (CqlComparers.CqlComparerImplementation.cs:152) has no BaseType walk, but `CqlComparers` is constructed with `equalsImplementation: CqlComparerEqualsImplementation.Compare` (CqlComparers.cs:26-28), so `Equals` routes through `CompareValues` → `SelectComparer` → `SelectComparerUncached`'s BaseType recursion (lines 60-101). A subclass outside the model assembly genuinely resolves via its registered base type.
- No `GeneratorToolVersion` bump is required. The diff is 3 files touching only `Cql.Firely` runtime extension code, a test, and a doc — nothing in `Cql.Compiler`, `CodeGeneration.NET`, or `ICqlOperators`, so no emitted C# changes.
- Conventions hold. The new test file carries the standard header with the current year (2026) and its own `#nullable enable`; its three `using`s (`Hl7.Cql.Comparers`, `Hl7.Cql.Fhir.Extensions`, `Hl7.Fhir.Model`) are none of them duplicates of `Cql/CoreTests/GlobalUsings.cs`; `[TestCategory("UnitTest")]` matches sibling classes. The modified `CqlComparersExtensions.cs` correctly retains its original 2023 header. The missing trailing newline on that file is pre-existing on develop (verified by hexdump of `git show origin/develop:...`), not introduced here.

```
Built `dotnet build Cql/CoreTests/CoreTests.csproj -c Release -f net10.0` — 0 errors, 195 warnings (all pre-existing nullable warnings in PrimitiveTests.cs etc., not from this PR).

1. New tests on PR head: `dotnet test ... --filter "FullyQualifiedName~CompareResourcesByIdTests"` → `Passed! - Failed: 0, Passed: 6, Skipped: 0, Total: 6`.

2. Behaviour-lock verification (the claim the PR makes). I copied `origin/develop`'s `CqlComparersExtensions.cs` over the PR's version, keeping the new test file, rebuilt (0 errors), and re-ran: `Passed! - Failed: 0, Passed: 6`. The tests do lock existing behaviour, exactly as claimed. Restored with `git checkout --`.

3. Mutation test A — "restricted to concrete types". Added `&& !t.IsAbstract` to the Lazy's filter, rebuilt (0 errors), re-ran: `Passed! - Failed: 0, Passed: 6`. NO test failed. This contradicts the test class's `<remarks>` (finding 1).

4. Mutation test B — truncated sweep. Changed the filter to `t == typeof(Patient)`, rebuilt, re-ran: `Failed! - Failed: 3, Passed: 3`; failures were `Observations_CompareByIdAlone`, `SuppliedStringComparer_DecidesIdEquality`, `CalledOnSeparateInstances_EachRegistrationIsCompleteAndIndependent`. The suite does catch a wrong type list.

5. Mutation test C — the hazard named in the review brief (static-per-process where it should be per-instance). I hoisted the `ResourceIdCqlComparer` itself into a `private static readonly` built from `StringComparer.Ordinal`, so the caller's `StringComparer` is ignored. Rebuilt, re-ran: `Failed! - Failed: 3, Passed: 3`; failures were `SuppliedStringComparer_DecidesIdEquality`, `CalledOnSeparateInstances_EachRegistrationIsCompleteAndIndependent`, `CalledTwiceO
```
</details>

**Acceptance criteria:** All four acceptance criteria on issue #1479 are met, and all four are already ticked on the issue — no checkbox sync is needed.

AC1 — "`CompareResourcesById` no longer calls `Assembly.GetTypes()` per invocation; the type sweep is behind a process-wide lazy computed at most once": MET. The sweep now lives in `private static readonly Lazy<Type[]> ResourceTypes` (CqlComparersExtensions.cs:109-110) and the method body retains only the `Register` loop over `ResourceTypes.Value` (lines 96-101). `Lazy`'s default `ExecutionAndPublication` mode guarantees at-most-once evaluation.

AC2 — "Behavior-lock tests cover the registration semantics ... all passing on `develop` too": MET, and this is the criterion I verified hardest. Several types beyond `Patient` are covered (`Observations_CompareByIdAlone

**Release note:** Present, correctly named, correctly sectioned, and accurate. `docs/releases/vnext/1481-compare-resources-by-id-once.md` follows the `<PR-number>-<short-slug>.md` rule from `docs/releases/vnext/README.md` (1481 is the PR, not the issue — correct). Its `## Improvements` heading is a canonical template section (`docs/releases/release-notes-template.md:32, "#### Improvements"` under "### CQL SDK") and is already in use by two sibling fragments on develop (1469, 1518).

Every factual claim in it chec

1. **Test class `<remarks>` claims coverage the suite does not have — "restricted to concrete types" breaks nothing** — The `<remarks>` block says the tests exist so that "a future change to how the type list is computed (filtered differently, cached per instance, restricted to concrete types, ...) shows up as a test failure". Two of those three examples are not actually covered. Restricting the sweep to concrete types is unobservable through the public surface: runtime types are always concrete, so a concrete registration is always f
2. **`CalledTwiceOnSameInstance_LastRegistrationWins` asserts a property that does not hold for base-type-resolved types** — The test pins "last registration wins" using `Patient`, which the sweep registers directly, so `Register`'s `AddOrUpdate` overwrites it. For a resource type resolved through the base-type walk the property fails: the first comparison memoizes the resolved comparer onto the derived type via `Comparers.TryAdd(xType, comparer)` in `CqlComparers.SelectComparer` (Cql/Cql.Runtime/Comparers/CqlComparers.CqlComparerImplement

### Notes — no action needed

- **`Lazy<T>` permanently caches a `ReflectionTypeLoadException` where the old code retried per call** — `LazyThreadSafetyMode.ExecutionAndPublication` (the default for `new Lazy<T>(Func<T>)`) caches an exception thrown by the value factory and rethrows the same instance on every subsequent access. `Assembly.GetTypes()` throws `ReflectionTypeLoadException` when the model assembly is only partially loadable, so the old per-call code would re-attempt the sweep on each `CompareResour
- **Release fragment cites the issue number where sibling fragments cite the PR number** — The fragment is correctly named `1481-...` (the PR number, per `docs/releases/vnext/README.md`) but its trailing citation is `(#1479)`, the issue. Every fragment currently on develop cites the number matching its own filename: `1467-shared-bundle-index.md` → `(#1467)`, `1469-age-operator-patient-lookup.md` → `(#1469)`, `1518-patient-type-info.md` → `(#1518)`. Since these lines

alexzautke:
@baseTwo Both fixed in `d503798d5`, and I acted on both notes as well.

### Also 1 — remarks claimed coverage the suite doesn't have ✅ valid, fixed

Your mutation result is right, and the reasoning generalizes further than the remark stated: a runtime type is always concrete, so restricting the sweep to concrete types cannot change any observable registration — and *moving the type list to a per-instance cache* is equally invisible, for the same reason. Two of the three examples were unfalsifiable, not one.

The remarks now name the two breakages the suite genuinely catches — a type list that loses types a runtime instance can have, and a comparer that stops being built per call from the caller's `StringComparer` — and say explicitly which changes are invisible here and why. Stating the blind spot seemed better than manufacturing a test for a change that is, by construction, unobservable through the public surface.

### Also 2 — `LastRegistrationWins` over-generalized ✅ valid, fixed, and verified

Confirmed your mechanism empirically rather than taking it on trust: `SelectComparer` memoizes a base-type resolution with `Comparers.TryAdd(xType, comparer)` (`CqlComparers.CqlComparerImplementation.cs:70`), and `TryAdd` does not replace, so a second opt-in never reaches the derived type.

- Renamed to `CalledTwiceOnSameInstance_LastRegistrationWinsForTheRegisteredTypes`, with the doc pointing at the counterpart.
- Added `CalledTwiceOnSameInstance_KeepsTheComparerATypeAlreadyResolvedThroughItsBaseType`: `DerivedPatient` compared under `Ordinal`, then a second opt-in with `OrdinalIgnoreCase` — the derived type still compares case-sensitively while `Patient` itself flips. It passes, so the asymmetry is real and now pinned rather than latent.

### Notes — both acted on, since each was a one-line accuracy fix

- **`Lazy<T>` caching a `ReflectionTypeLoadException`.** This is a genuine behavioral difference on the failure path, and the fragment claimed "registration behavior is unchanged" without qualification. It now states it: a partially loadable model assembly makes the failure permanent where the per-call sweep re-attempted it. Worth a reader knowing, even though the healthy path is identical.
- **Fragment citation.** Changed to `(#1479, #1481)`. Rather than swapping issue for PR, this matches what the *published* notes actually carry — the consolidated 2.12.0 entries read `(#1450, #1451)` and `(#1454, #1455)`, i.e. issue and PR — and it keeps the fragment consistent with the same fix I made on #1474.

Also folded the exception-caching caveat and the corrected test description into the PR body.

### Staleness — already resolved

Head is `8913dcb8b`, a `develop` merge; `git rev-list --count HEAD..origin/develop` is **0**. Your review predated it.

**Validation:** `CompareResourcesByIdTests` 7/7; full `CoreTests` net10.0 Release **1174 passed, 8 skipped, 0 failed**.

---

baseTwo (approved):
## Approving `d503798d5` — both nits fixed

@alexzautke — the `<remarks>` now claims only the coverage the suite actually has, and
`CalledTwiceOnSameInstance_LastRegistrationWins` no longer asserts a property that fails for a type resolved
through the base-type walk.

Tests: filtered suite green, full `CoreTests` on net10.0 green.

One note, recorded not blocking: the rewritten `<remarks>` still generalizes very slightly past what the seven
tests exercise. And the `Lazy<T>` point stands as a note — it permanently caches a `ReflectionTypeLoadException`
where the old code retried per call. Neither is worth another round.

PR #1482, opened 2026-08-01
Format Iso8601 date/time strings lazily

### LINKED ISSUE #1478: Iso8601 date/time types eagerly format their string representation in every constructor
## Summary

`DateTimeIso8601`, `DateIso8601` and `TimeIso8601` (project `Cql/Iso8601`) build their full string representation with `Format(...)` inside every constructor, storing it in a `readonly string` field. The string is only needed when someone actually calls `ToString()` (display, serialization, string conversion) — but CQL evaluation constructs these types constantly (date arithmetic, `FhirDateTime` → `CqlDateTime` conversions, comparisons), and the overwhelming majority of those instances never have their string read.

A working implementation exists on the `claude/perf-execution-hot-path` branch (with 154 golden tests capturing the exact pre-change formatting); this ticket tracks landing it standalone. Note that branch also converted the files' line endings CRLF→LF — the landing PR must re-apply the semantic change onto the CRLF originals so the diff shows only the real ~18-line change per file.

**Update during implementation:** adversarial review showed the hot-path branch's version was *not* byte-identical for `strict: true` / explicit `DateTimePrecision.Unknown` constructor paths (the eager code formatted with the precision *argument*, not the derived `Precision`). The landing PR fixes this by recording per constructor the precision the eager `Format` call used, and the golden corpus was expanded from 154 to 426 rows (captured against develop) to pin those edges.

## Impact

Measured with BenchmarkDotNet-style micro-benchmarks: constructing a `CqlDateTime` costs 1232 ns / 2864 B eagerly vs 545 ns / 792 B lazily. On the 900-measure CMS harness this change alone removed ~66 MB (~10%) of total allocation. It is also the prerequisite for removing `FhirTypeConverter`'s process-wide `LRUCache<CqlDateTime>`: with cheap construction, the measured cost of running without that cache halves (from +68 MB to +37 MB per harness run), making removal a realistic follow-up.

## Root Cause

Constructor chains pass an eagerly computed `Format(Year, Month, …)` into the field initializer, unconditionally.

## Expected Behavior

- The string field becomes `private string? _string` with a lazy property: `_string ??= Format(...)`. The benign race (two threads formatting concurrently) is harmless — `Format` is pure and both produce the same string.
- Public constructors pass `null` (no eager formatting).
- The parse path (internal constructors taking the original input text) stores the literal string, so parse → `ToString()` round-trips stay byte-identical to today.
- `ToString()` output is **byte-identical** for every input — pinned by golden tests whose expected values were captured against the pre-change implementation, and which therefore pass on `develop` and on the new implementation alike.

## Acceptance Criteria

- [x] No constructor formats eagerly; the string is computed at most once, on first read.
- [x] Golden tests covering all three types and the parse round-trip pass against both `develop`'s and the new implementation (behavior lock).
- [x] The PR diff against `develop` shows only the lazy-string hunks — no line-ending churn, no header changes.
- [x] No public API changes; `ICqlOperators` untouched (no `GeneratorToolVersion` bump).
- [x] A release-note fragment under `docs/releases/vnext/` (`## Improvements`) citing only the measured numbers.

alexzautke:
## Description
`DateTimeIso8601`, `DateIso8601` and `TimeIso8601` built their full string representation with `Format(...)` inside every constructor, but CQL evaluation constructs these types constantly (date arithmetic, `FhirDateTime` → `CqlDateTime` conversions, comparisons) and the overwhelming majority of instances never have their string read.

The string field becomes `private string? _string` with a lazy property (`_string ??= Format(...)`); public constructors pass `null` instead of an eagerly formatted string. The parse path stores the original input literal, so parse → `ToString()` round-trips stay byte-identical. The `??=` race is benign — `Format` is pure and both racers produce the same string.

One subtlety, caught by adversarial review and fixed before this PR went up: the eager code formatted with the constructor's *precision argument*, which is **not** always the instance's derived `Precision` (under `strict: true` validation, or with an explicit `DateTimePrecision.Unknown`, the two diverge — develop's own `Precision` property already disagreed with its string in those cases). Each constructor therefore records the exact precision its eager `Format` call used (`_stringPrecision`) and the lazy property formats with that, reproducing develop's output byte-for-byte everywhere rather than silently changing `ToString`/`Equals`/`GetHashCode` on those public constructor paths. The self-consistency question (string vs `Precision` under `strict`) is deliberately left as-is for a separate discussion.

Measured: constructing a `CqlDateTime` goes from 1232 ns / 2864 B to 545 ns / 792 B; ~66 MB (~10%) less total allocation on a 900-measure CMS evaluation run. This is also the prerequisite for re-evaluating `FhirTypeConverter`'s process-wide `LRUCache<CqlDateTime>` (cheap construction halves the measured cost of running without that cache).

No public API changes (only `internal` constructor parameters changed `string` → `string?`); no `ICqlOperators`/generated-code impact, so no `GeneratorToolVersion` bump. The three files keep their original CRLF line endings and BOM — the diff is ~30 lines per file, no whole-file churn. A release-note fragment is included under `docs/releases/vnext/`.

## Related issues
Closes #1478.

## Testing
New `Cql/CoreTests/Iso8601LazyFormattingGoldenTests.cs`: a golden corpus of **426 rows** whose expected values were captured by *running develop's implementation* (never hand-computed), covering all three types across all seven precisions, `strict: true` and `strict: false`, explicit `DateTimePrecision.Unknown`, milliseconds 0/1/999, negative/zero/half-hour/absent offsets, the composing `DateTimeIso8601(DateIso8601, TimeIso8601?)` constructor, the `DateTimeOffset`-taking overloads, parse round-trips of unusual-but-valid literals, and two strict-validation throw cases.

The corpus passes against **both** develop's eager implementation and this PR's lazy implementation (verified by swapping the three files, confirming the DLL genuinely recompiled, and re-running). The added rows are discriminating, not decorative: replayed against an earlier draft of this change that formatted with a hardcoded precision, 68 of them fail — independently reproducing the divergence the adversarial review found by diffing 243 constructor scenarios.

Full CoreTests on net10.0 (Release): 1056 passed, 8 skipped, 0 failed (1064 total).

## FirelyTeam Checklist

---


baseTwo (commented):
## Review of `59e374ab7` — good, small things only

Re-checked at the new head after you merged `develop` in: **0 behind**, and every finding below was re-verified against `59e374ab7`, not carried over from the previous head.

The change is correct and genuinely behaviour-preserving — I diffed 105,704 constructor scenarios between develop's eager implementation and this branch and got zero differences across ToString, Precision, Equals, GetHashCode and every public component property — but the golden corpus records `Precision` for 154 rows and a constant `True` column for 84 more that the test silently discards, so the one property this design most easily breaks is captured and then thrown away.

### Verified

- Behaviour is byte-identical: 105,704 differential scenarios across ToString, Precision, Equals, GetHashCode and every public property produce zero diffs between develop's eager implementation, this branch, and the merged state.
- The golden corpus passes unmodified against develop's eager implementation (13/13 with origin/develop's three Iso8601 files swapped in and the DLL confirmed rebuilt) — it is a genuine behaviour lock, not a snapshot of the new code.
- The corpus is discriminating: replacing `_stringPrecision` with `Precision` in the lazy property makes exactly 68 of 426 rows fail, reproducing the number in the PR body.
- The `_stringPrecision`-vs-`Precision` subtlety is real and correctly handled. `new DateTimeIso8601(dto, DateTimePrecision.Unknown)` yields Precision=Millisecond but ToString()="2024" on develop, and the PR reproduces that exactly rather than 'fixing' it silently.
- No live null-string path exists: the only callers of the three internal `string?` constructors are the three TryParse methods, which all pass the original literal (grep over Cql/ for `new (DateTimeIso8601|DateIso8601|TimeIso8601)\(`). Iso8601.csproj's sole InternalsVisibleTo names `Cql.Primatives`, which matches no AssemblyName in the repo, so the internal surface is effectively assembly-private.
- Interaction with the merged #1458 work is clean. Develop's 6cf3b7a77 refactored Format's offset block into `FormatKnownOffset`; the PR's hunks are in the field/constructor region. Merge is conflict-free and the merged tree is behaviourally identical to both parents.
- No public API change: Cql/Iso8601/PublicAPI.Shipped.txt lists all 7 public constructors and is untouched, and the build is clean, so the PublicAPI analyzer confirms it. Only `internal` parameters went `string` -> `string?`.
- No generated-code impact: `ICqlOperators` untouched, no `*.g.cs` or `_CODE GENERATOR VERSION_.cs` in the diff, so no GeneratorToolVersion bump is required — correct.

```
Everything below ran in my own worktree C:/Drives/D/cql-review-wt/pr-1482, which I restored to a clean PR head afterwards (`git status --porcelain` empty).

1. Build: `dotnet build Cql/CoreTests/CoreTests.csproj -c Release` -> 0 Error(s), 390 Warning(s) (all pre-existing nullable warnings in PrimitiveTests.cs etc.).

2. Full suite: `dotnet test Cql/CoreTests/CoreTests.csproj -c Release -f net10.0` -> "Passed! - Failed: 0, Passed: 1056, Skipped: 8, Total: 1064, Duration: 33 s". Exactly matches the PR body's claim.

3. BEHAVIOUR-LOCK CHECK (does the new corpus pass on develop?): `git checkout origin/develop -- Cql/Iso8601/{DateIso8601,DateTimeIso8601,TimeIso8601}.cs`, confirmed the eager code was in place (`private readonly string String;` present at DateIso8601.cs:42, `_stringPrecision` count 0), rebuilt (confirmed Hl7.Cql.Iso8601.dll timestamp advanced to 5 s before the run), re-ran `--filter "FullyQualifiedName~Iso8601"` -> "Passed! - Failed: 0, Passed: 13". The 426-row corpus genuinely passes against develop's eager implementation. The claim is real.

4. DISCRIMINATING-POWER CHECK (does the corpus fail if the change is done wrong?): restored the PR files, then mutated the lazy property in all three files to format with `Precision` instead of `_stringPrecision` — i.e. the naive rewrite. Rebuilt, re-ran -> "Failed TextFormMatchesTheEagerImplementationsGoldens ... Assert.Fail failed. 68 of 426 golden rows diverged". Exactly the 68 the PR body claims. The corpus is discriminating, not decorative, on the string dimension.

5. FULL DIFFERENTIAL (my strongest check): I wrote a temporary harness (Cql/CoreTests/ZzDifferentialDump.cs, since deleted) that walks a cartesian product
```
</details>

**Acceptance criteria:** All five acceptance criteria on #1478 are ticked, and I independently confirmed all five.

1. "No constructor formats eagerly; the string is computed at most once, on first read." MET. All four public constructors across the three types now pass `null` to the internal constructor; the only `Format` call sites are the three `String` properties (DateIso8601.cs:57, DateTimeIso8601.cs:92, TimeIso8601.cs:115), each guarded by `??=`. Computed at most once per instance modulo the benign publication race, which I stress-tested (2000x16 threads, always one distinct string and hash).

2. "Golden tests ... pass against both develop's and the new implementation (behavior lock)." MET, and this is the criterion I checked hardest. With origin/develop's three Iso8601 files swapped into my worktree and the

**Release note:** Present, correctly named, and accurate apart from one wording nit.

docs/releases/vnext/1482-lazy-iso8601-formatting.md. Naming follows docs/releases/vnext/README.md ("`<PR-number>-<short-slug>.md` ... use the PR number") — the second commit on the branch, abc286657, exists precisely to rename it to the PR number, so that convention was applied deliberately. Section heading is `## Improvements`, which maps to `#### Improvements` under `### CQL SDK` in docs/releases/release-notes-template.m

1. **Row-count assert fires before the divergence report, hiding the useful failure** — `Assert.AreEqual(ExpectedRowCount, rows, ...)` at line 86 runs before the `failures.Count > 0` block at line 88. Anyone who both adds a row and breaks formatting sees only 'The golden corpus changed size' and loses the list of diverged rows, which is the information they actually need. Move the `if (failures.Count > 0) Assert.Fail(...)` block above the count assertion.
2. **Release fragment says the string is built on first `ToString()`; `Equals` and `GetHashCode` also trigger it** — The fragment reads 'it is now computed lazily on first `ToString()`'. `Equals(object?)` and `GetHashCode()` both read the same `String` property, so either of those on a never-printed instance also pays the format. Reword to 'computed lazily on first use of its text form (`ToString()`, `Equals` or `GetHashCode`)'. Worth being precise because a consumer benchmarking their own code needs to know that hashing a `CqlDate

### Notes — no action needed

- **The preserved `Unknown`-precision behaviour is a live equality bug, and the promised 'separate discussion' has no issue behind it** — The PR body says the string-vs-Precision self-consistency question is 'deliberately left as-is for a separate discussion', and the code comments in all three files say the same. Preserving it in a performance PR is the right call — but nothing tracks it, so `_stringPrecision` and its three-paragraph comment become permanent by default. The consequence is not cosmetic: two entir
- **Branch is 90 commits behind develop and needs updating before landing (no rework required)** — Not a defect in the change — recording it so it is a conscious decision rather than a surprise, and because per repo convention the base-merge is the author's/maintainer's call, not a reviewer's. Develop has moved 90 commits, including 6cf3b7a77 which refactored the offset-rendering block inside `Format` in two of the three files this PR touches. I confirmed the merge is confli

### Considered and dropped

- **Golden corpus records `Precision` for 154 rows and a constant `True` for 84 more, and the test asserts neither** — The finding's *data* observation is accurate, but its justification for `required` severity is demonstrably false, and its proposed fix is not what it claims.

WHAT I CONFIRMED (parsing the two verbatim strings in `C:/Drives/D/cql-review-wt/pr-1482/Cql/CoreTests/Iso8601LazyFormattingGoldenTests.cs` 
- **The newly nullable internal ctor `@string` parameter has an unstated, unenforced invariant; violating it silently truncates ToString()** — The finding does not hold up as a `required` fix. Four independent problems, all verified by reading the source in the read-only worktree (no build needed — every point is settled by the code and the committed golden corpus).

1) THE "EMPIRICAL CONFIRMATION" IS PRE-EXISTING, DELIBERATELY GOLDEN-PINN

alexzautke:
@baseTwo Both fixed in `3928e0cfd`, and I filed the missing tracker for the note.

### Also 1 — row-count assert hiding the divergence report ✅ valid, fixed and verified

Moved the `failures.Count > 0` block above the count assertion, with a comment saying why the order matters.

Verified it actually does what you asked rather than assuming: I broke both dimensions at once — mutated `DateIso8601`'s lazy property to format with `Precision` *and* set `ExpectedRowCount` to 999 — and the reported failure is now `Assert.Fail failed. 4 of 426 golden rows diverged:` followed by the rows. Before the reorder that same state reported only `Expected:<999>. Actual:<426>. The golden corpus changed size`, exactly the information loss you described.

### Also 2 — fragment crediting only `ToString()` ✅ valid, fixed

Confirmed against the source before rewording: `Equals(object? obj) => Equals(String, obj?.ToString())` and `GetHashCode() => String.GetHashCode()` (`DateIso8601.cs:170-171`), so both pay the format on a never-printed instance. The fragment now says the string is built on first use of the text form — `ToString()`, `Equals` or `GetHashCode` — and states the consequence plainly: an instance whose text form is never used never pays, one that is hashed or compared still does.

### Note — filed as #1524, and the code now points at it

You were right that nothing tracked the `Unknown`-precision divergence, and that this makes `_stringPrecision` plus its explanatory comment permanent by default. Filed **#1524**, and the comment in all three files now ends with *"Whether that divergence is the intended contract is tracked separately, in issue #1524."* — so the field reads as a recorded deferral rather than an inherited accident.

The issue captures what you noted about it not being cosmetic: since `Equals` and `GetHashCode` are both defined over the text form, two instances differing in their components can compare equal and hash identically when their truncated strings coincide, and `Precision` does not describe what equality is actually based on. It also records that the golden corpus already discriminates the two behaviours precisely (the 68 rows), so whichever answer is chosen can be pinned rather than assumed, and that changing it needs a `Potentially Breaking` note.

### One self-inflicted problem, caught and undone

My first pass at adding the issue reference rewrote the three `Cql/Iso8601` files through a text-mode write, which converted their CRLF endings to LF and produced a 1,117-line whole-file diff — precisely the churn this PR had gone out of its way to avoid. Reverted and redid it in binary mode; the diff is back to **+3/−1 per file**, endings and BOM intact.

### Staleness

Head is `59e374ab7`, a `develop` merge, 0 behind — consistent with your re-check note (the "90 commits behind" line in the Notes section reads as carried over from the earlier pass).

**Validation:** `Iso8601` tests 13/13; full `CoreTests` net10.0 Release **1168 passed, 8 skipped, 0 failed**.

---

baseTwo (approved):
## Approving `3928e0cfd` — both nits fixed

@alexzautke — the divergence report now fires before the row-count assert, so someone who both adds a row and
breaks formatting keeps the useful diagnostic instead of losing it to "the corpus changed size". The fragment no
longer claims the string is built only on first `ToString()`, which was inaccurate given `Equals` and
`GetHashCode` read it too.

Tests: filtered suite green, full `CoreTests` on net10.0 green.

Standing behind the original verification: 105,704 constructor scenarios diffed between develop's eager
implementation and this branch produced zero differences across `ToString`, `Precision`, `Equals`, `GetHashCode`
and every public component property.

One pre-existing note: the golden corpus records a `Precision` column nothing asserts. Not introduced here.

PR #1490, opened 2026-08-02
Fuse immediately-consumed Where/Select chains at codegen

### LINKED ISSUE #1484: Fuse immediately-consumed Where/Select chains during code generation
## Summary

`CqlOperators.Where`/`Select` materialize their result into a `List<T>` on every call. Generated CQL frequently consumes such a result exactly once, immediately, in another operator — `Exists(Where(s, p))`, `Select(Where(s, p), f)`, `Where(Select(s, f), p)`, `Distinct(Select(s, f))` — so the intermediate list exists only to be walked once and discarded. List growth during these materializations is the single largest CPU cost in the runtime layer (~20% of active thread-time on a 900-case CMS corpus measured with fixture noise removed), and the discarded lists are a large share of operator allocation.

The compiler should emit **fused operators** for these four shapes.

## Evidence (measured prototype)

A prototype implemented the four fused operators and rewrote the generated C# of the dqm-content-qicore-2025 corpus (1,340 fusions across 86 of 89 libraries), then ran an order-balanced steady-state A/B (900 cases × 100 reps, medians of reps 86–100, bundle-deserialization cached out of the loop):

| dop1 | base | fused | Δ |
|---|---|---|---|
| CPU | 525 ms | 496 ms | −5.5% |
| allocation | 407 KB/case | 382 KB/case | **−6.1%** (all 8 round-pairs within −5.7…−6.6%) |
| gen0 / GC pause | 22 / 17.1 ms | 20 / 15.2 ms | −8% / −11% |

dop4 medians were −9% wall/CPU (one adverse round; direction consistent). The hottest library (CMS117, n³ cross-join pipelines) improved 7.7%. Correctness was gated on a per-case oracle: 900 cases × 3 reps produced **byte-identical** outcome files for base vs fused.

## Design constraints (from the prototype)

- **Non-short-circuiting semantics, by construction.** Each fused operator evaluates the same lambdas over the same elements in the same order as the composition it replaces — `WhereAny` does **not** stop at the first match. This keeps `Message` side effects, exception surfaces, and results byte-identical without any per-site safety analysis. (A short-circuiting tier is possible under the spec's rewrite license — Author's Guide: implementations "may use short-circuit evaluation to reduce computation" — but it is explicitly out of scope here.)
- One acceptable, documented nuance: single-pass `WhereSelect`/`SelectWhere` interleave the two lambdas per element (`p(x₁), f(x₁), p(x₂)…`) instead of running all `p` then all `f`. No evaluation is added or skipped; only if *two different lambdas would both throw* can the surfacing exception differ. The corpus oracle showed no observable difference.
- Fusion applies only when the producer result is consumed exactly once as the consumer's source argument. The code generator emits strict SSA and never reuses `Where`/`Select` results (verified corpus-wide: 0 multi-use sites), so a structural pattern-match suffices.
- Fusion should happen at the **expression-tree level** in `Cql.Compiler` (a pass over the built operator-call expressions, or pattern-matching at build time) — not as a source-text rewrite of emitted C#.

## Acceptance Criteria

- [x] Four fused operators on `ICqlOperators`/`CqlOperators`: `WhereAny`, `WhereSelect`, `SelectWhere`, `SelectDistinct`, each semantically identical to its composition (null-source/null-element rules, result collection types — e.g. `SelectDistinct` must match `Distinct`'s output shape — and lambda evaluation traces), covered by unit tests that compare each fused operator against its literal composition including lambda-invocation order and exception behavior. *(29 differential tests; operators byte-identical to the proven prototype.)*
- [x] The compiler emits fused calls for the four patterns; compiler-level tests cover each pattern and at least one chain (`Select→Where→Select→Distinct` collapsing to `SelectWhere` + `SelectDistinct`) plus a negative case (producer result consumed as a non-source argument must NOT fuse). *(11 binder tests incl. five negatives: non-source position, Where-over-Where, cast-wrapped, null-conditional, generic mismatch.)*
- [x] `GeneratorToolVersion` gets a **minor** bump (new `ICqlOperators` members, non-breaking addition), and the current `LibraryInstanceInvoker` supported range (`MinSupportedGeneratorToolVersion`/`FirstUnsupportedGeneratorToolVersion`) is verified to cover the new version. *(Amended per adversarial-review adjudication: the minor bump — 5.2.0.0 — required **widening** `FirstUnsupportedGeneratorToolVersion` to 5.3.0.0, since 5.2.0.0 was exactly the old boundary. Minor-with-widening was chosen over a patch bump because generated assemblies now hard-depend on new runtime API: under a patch bump an older runtime would load them and fail with `MissingMethodException` mid-evaluation, whereas the minor bump makes older runtimes skip such libraries cleanly at load. Rationale recorded on the invoker field and in the PR.)*
- [x] Every checked-in `*.g.cs` is regenerated in the same PR; only files whose content genuinely changed are committed (header-only churn reverted per the repo convention). *(128 committed, 21 header-only reverted; line-level audit: every changed line is version-header, removed composition, or added fused call; removed compositions = exactly 2× the 2,040 fusions.)*
- [x] Full CoreTests suite green; `PublicAPI.Unshipped.txt` updated for the new interface members. *(1095/0/8 net10.0; corpus oracle byte-identical, re-verified independently by the review against this branch's compiler output.)*
- [x] A release-note fragment (`## Improvements`) citing only the measured numbers above, with the instrument caveats (fixture-free CMS corpus, steady-state protocol). *(Plus `## New Public API` and `## Potentially Breaking` sections per review.)*

## Non-goals (follow-up candidates, evidence pending)

- `UnionMany` (634 corpus sites of left-associated `Union` chains) and `SortBy` fusions (`WhereSortBy`, `SortByFirst/Last` — the latter only pay with short-circuiting).
- The short-circuiting optimization tier and the evaluation-contract documentation it requires.

**PR: #1490.**

alexzautke:
## Description
Generated CQL frequently materializes a `Where`/`Select` result into a `List` only for it to be consumed exactly once by the next operator and discarded — list growth during those materializations was the largest single CPU cost in the runtime layer (~20% of active thread-time, measured fixture-free on the 900-case CMS corpus). This PR fuses the four dominant shapes at code generation:

- **Four new fused operators** on `ICqlOperators`/`CqlOperators`: `WhereAny` (≡ `Exists(Where(s,p))`), `WhereSelect` (≡ `Select(Where(s,p),f)`), `SelectWhere` (≡ `Where(Select(s,f),p)`), `SelectDistinct` (≡ `Distinct(Select(s,f))`). All are **non-short-circuiting by construction**: they evaluate the same lambdas over the same elements in the same order as the composition they replace and only skip the intermediate list. While every lambda runs to completion, results, `Message` side effects and exception surfaces are identical. One documented nuance: single-pass `WhereSelect`/`SelectWhere` interleave the two lambdas per element instead of running all of one then all of the other. That is observable when a lambda throws — the fused form has already invoked the *other* lambda for elements preceding the failure (same surfacing exception), and in the corner case where *both* lambdas would throw on different elements, the fused form surfaces whichever throw its single pass reaches first, which need not be the one the composition reaches. Both are documented in the operator remarks and the release fragment, and pinned by divergence tests.
- **The fusion pass** lives in `Cql/Cql.Compiler/CqlOperatorsBinder.Fusion.cs`, hooked at the point every operator invocation is bound. It fuses only when the producer is *literally* the consumer's source argument (casts, null-conditional access, and non-source positions decline), cross-checks generic type arguments, declines rather than throws on any construction failure, and reaches chain fixpoint bottom-up (e.g. `Select→Where→Select→Distinct` collapses to `SelectWhere` + `SelectDistinct`). All `ICqlOperators` method lookups go through `ReflectionUtility.GenericMethodDefinitionOf`.
- **Version:** `GeneratorToolVersion` 5.1.7.0 → **5.2.0.0**, with `LibraryInstanceInvoker_5_0`'s `FirstUnsupportedGeneratorToolVersion` widened to 5.3.0.0. A minor (not patch) bump is deliberate, per adversarial review: generated assemblies now hard-depend on the four new interface members, so an older runtime must *skip them at load with a warning* rather than fail with `MissingMethodException` mid-evaluation — which is exactly what the version gate provides under a minor bump. Recorded in the invoker file and the release fragment (`## Potentially Breaking`).
- **Regeneration:** all checked-in generated sets regenerated with the packager (`CoreTests/CSharp`, `LibrarySets/RR23`, the Demo sets, dqm-content-qicore-2025, the packaged example inputs): the remaining 5.1.x-headed files are exactly the header-only set. **2,040 fusions** corpus-wide (885 `WhereAny`, 543 `SelectDistinct`, 444 `WhereSelect`, 168 `SelectWhere`), net −2,040 generated lines. A line-level audit classified every changed line as version-header, removed composition, or added fused call — removed compositions = exactly 2 × fusions, independently confirming no producer was ever shared between consumers.

Measured effect (prototype of this exact transform, 1,340 fusions on the dqm corpus — the identical configuration this PR's compiler produces there; steady-state, order-balanced, fixture-free, dop1): **−5.5% CPU, −6.1% allocation (−25 KB/case, all round-pairs within −5.7…−6.6%), −8% gen0, −11% GC pause**; the hottest library (CMS117) −7.7%. Details and caveats in the fragment.

## Related issues
Closes #1484.

## Testing
- **Behavioral proof, end-to-end:** the 900-case CMS corpus (× 3 reps) produces **byte-identical per-case outcomes** with fused resources vs the committed baseline (sha-verified, independently re-run by the adversarial review against this branch's compiler output). A post-version-bump smoke pass (900 × 1) confirms 5.2.0.0-headed assemblies load and run under the widened invoker range — including mixed-version assemblies, exercising the range convention as designed.
- **`FusedOperatorTests` (33):** each fused operator differentially tested against its literal composition — value, per-lambda invocation trace, concrete result type, and exception behavior — over null sources, null elements, duplicates, and value-type projections; non-short-circuiting and interleaving explicitly pinned, including the throwing-lambda divergences (side effects already fired for preceding elements; both-throw surfacing-exception difference in each direction).
- **`CqlOperatorsBinderFusionTests` (12):** each fusion shape through the real binder, the chain-collapse case, and negative cases: `Where`-over-`Where`, cast-wrapped producer, null-conditional producer, a covariant generic mismatch (with the actual decline mechanism traced and asserted, not assumed), a non-fusable consumer, and `ProducerInANonSourceArgument_IsUnrepresentableInTheIR`, which pins that the source-position guard is defence-in-depth by proving the shape cannot be constructed in the IR.
- **`FusedOperatorsCqlTest` (5):** a real CQL fixture (`FusedOperatorsTest.cql` + checked-in ELM) whose generated output exercises all four fused operators end-to-end, plus a shape assertion that regenerates from the ELM and fails if any fusable composition survives unfused.
- **Suites:** CoreTests 1217/0/8 (net10.0), CqlToElmTests 2497/0/33, full `Cql-Sdk.slnf` Release build 0 errors.
- An adversarial review (15-case probe against the binder internals, regeneration audit, independent corpus regeneration byte-compare, oracle re-run) found zero blocking issues; its version ruling and nits are incorporated.

## FirelyTeam Checklist

---


alexzautke:
@copilot Fix the merge conflict

baseTwo (changes requested):
## Review of `9ece795c0` — changes requested

Re-checked at the new head after you merged `develop` in: **0 behind**, and every finding below was re-verified against `9ece795c0`, not carried over from the previous head.

The fusion itself is correct and unusually well-evidenced — I independently confirmed the version gate, the generated-corpus arithmetic and the operator semantics — but the safety invariant the whole design rests on ("no evaluation is added or skipped") is demonstrably false when a lambda throws, and three of the issue's acceptance criteria are not actually met (decorative non-source negative test, two unregenerated example `.g.cs`, missing Breaking classification for the `ICqlOperators` additions).

### Verified

- GeneratorToolVersion bumped 5.1.7.0 → 5.2.0.0 in `Cql/CodeGeneration.NET/_CODE GENERATOR VERSION_.cs`, AND the invoker range widened: `LibraryInstanceInvoker_5_0.MinSupportedGeneratorToolVersion = 5.1.0.0`, `FirstUnsupportedGeneratorToolVersion` 5.2.0.0 → 5.3.0.0, `SupportsVersion` = `>= min && < firstUnsupported`, so 5.2.0.0 is inside the range. Verified BOTH explicitly. Reverting the widening fails 17 ToolkitTests, so the pairing is guarded by `ElmToolkit_GeneratorToolVersion_IsWithinInvocationToolkitSupportedRange`.
- The release note's claim that an older runtime 'skips such libraries at load with a warning rather than half-working' is accurate: `Cql/Cql.Invocation/Toolkit/LibraryInvoker.cs:152-158` calls `logger?.LogWarning("Skipping type {type} because it was generated with an unsupported version {ver}…")` and returns false.
- Fused-operator semantics match their compositions line by line against `CqlOperators.cs:221` (`Select`), `:236` (`Where`), `CqlOperators.ListOperators.cs:42` (`Distinct`), `:111` (`Exists`) — including `Where(null)=null`/`Exists(null)=false` giving `WhereAny(null)=false`, `Exists` ignoring null elements (`(lambda(e) ?? false) && e != null`), and `SelectDistinct` reproducing `Distinct`'s exact shape (`List<object?>` + `HashSet<object>(EqualityComparer)` + at-most-one-null + `CastToArray<object?, TR>()` returning an array, not a List).
- All four fused signatures return the same .NET type as the consumer they replace (`bool?`, `IEnumerable<TR>`), so parent IR nodes are unaffected; `CqlOperatorsBinderFusionTests` asserts `call.Type` for each.
- The 'single use is structural' argument survives the emitter. `CSharpEmitter.Scope.Hoist` deduplicates hoisted subexpressions by *printed text*, not node reference — so a shared producer could in principle have collapsed to one evaluation pre-fusion and two post-fusion. It cannot: `HoistLocalFunction` allocates a fresh name per `CodeLambda` linearization and never dedups, and `Where`/`Select` always take a `CodeLambda` (enforced in `CqlOperatorsBinder.Specific.cs:281,296`), so two `Where`/`Select` calls can never print identically. Fusion cannot change evaluation counts through that path.
- `TryBuild`'s `catch (ArgumentException)` is the right net: `CodeInvoke`'s ctor and `CodeTypeRules.ValidateAssignment` (`CodeTypeRules.cs:79-90`) both throw `ArgumentException`, as does `MakeGenericMethod` on a constraint violation.
- All four `Fuse*` helpers cross-check generic arguments against the producer, so a covariantly-bound consumer declines even if the binder ever stopped inserting a cast.
- New files carry the 2026 copyright header and their own `#nullable enable`; `CqlOperators.FusedOperators.cs` puts `#pragma warning disable CS1591` above the header, matching `CqlOperators.ListOperators.cs`. No `using` duplicates a GlobalUsings entry in Cql.Runtime / Cql.Compiler / CoreTests. `PublicAPI.Unshipped.txt` has all four members with correct nullability annotations.

```
All work done in the assigned worktree `C:/Drives/D/cql-review-wt/pr-1490`; worktree left clean (`git status --porcelain` empty).

BUILD
- `dotnet build Cql/CoreTests/CoreTests.csproj -c Release` → 0 errors, 394 warnings (all pre-existing nullable warnings in PrimitiveTests.cs etc.).
- `dotnet build Demo/Measures.dqm-content-qicore-2025/Measures.dqm-content-qicore-2025.csproj -c Release` → **Build succeeded, 0 Warning(s), 0 Error(s)**. This is the corpus that actually carries the fused calls (695 `WhereAny`, 290 `WhereSelect`, 125 `SelectWhere`, 230 `SelectDistinct` = the 1,340 the PR cites), so the emitted fused C# demonstrably type-checks against the real `ICqlOperators`.

TESTS
- Fusion suites: `dotnet test … -f net10.0 --filter "FullyQualifiedName~FusedOperatorTests|FullyQualifiedName~CqlOperatorsBinderFusionTests"` → **Passed! Failed: 0, Passed: 40, Skipped: 0** (29 + 11, matching the PR's claim).
- Full CoreTests: `dotnet test Cql/CoreTests/CoreTests.csproj -c Release -f net10.0` → **Passed! Failed: 0, Passed: 1145, Skipped: 8, Total: 1153, Duration: 55 s**. (PR claimed 1095/0/8 pre-merge; 1145 now.)

BREAK-THE-THING VERIFICATION (each change reverted afterwards, worktree confirmed clean)
1. Made `WhereAny` short-circuit (`any = true` → `return true`): **3 FusedOperatorTests failed** — `WhereAny_DoesNotShortCircuitAfterTheFirstMatch`, `WhereAny_NullElements_MatchesComposition`, `WhereAny_PredicateThrows_MatchesComposition`. The non-short-circuit pin is real.
2. Deleted the `&& !invoke.NullConditional` clause from `IsOperatorsCall`: **`NullConditionalProducer_DoesNotFuse` failed** (1/11). Real guard.
3. Deleted the source-position guard entirely — replaced `consumer.
```
</details>

**Acceptance criteria:** Issue #1484 has all six AC boxes ticked. Assessed against the code:

1. Four fused operators + differential unit tests (value, lambda trace, result type, exceptions) — **MET.** 29 tests, and I confirmed they bite: short-circuiting `WhereAny` fails 3 of them.
2. Compiler emits fused calls; tests for each pattern, a chain, **and a negative case (producer in a non-source argument must NOT fuse)** — **PARTIALLY MET.** The four shapes, the `Select→Where→Select→Distinct` chain, and the cast/null-conditional/generic-mismatch declines are genuinely covered. The non-source-argument negative case is not: `WhereResultAsNonSourceArgument_DoesNotFuse` passes with the position guard deleted entirely (finding #3).
3. Minor `GeneratorToolVersion` bump + `LibraryInstanceInvoker` range verified — **MET, and

**Release note:** Yes — `docs/releases/vnext/1490-operator-fusion.md`, correctly named `<PR>-<slug>.md` per `docs/releases/vnext/README.md`, and it uses only template-defined section names: `## New Public API`, `## Improvements`, `## Potentially Breaking`.

Accuracy check on each claim I could test:
- "`LibraryInstanceInvoker_5_0`'s supported range widens to `[5.1.0.0, 5.3.0.0)`" — **accurate** (`LibraryInvoker.5.0.cs:34,47`).
- "such libraries are … skipped at load with a warning rather than half-working" — **ac

### Required — @alexzautke please fix these

#### 1. Eight `ICqlOperators` methods resolved by string name + arity instead of via `ReflectionUtility` — `Cql/Cql.Compiler/CqlOperatorsBinder.Fusion.cs:75`

CLAUDE.md requires reflection to go through `ReflectionUtility` and states that `nameof` in a name-based lookup is not a substitute. Replace the eight `MethodInfo` fields and delete the `ICqlOperatorsMethod` helper: `private static readonly MethodInfo WhereMethod = ReflectionUtility.GenericMethodDefinitionOf(() => default(ICqlOperators)!.Where<object>(default!, default!));` and the same shape for `Select<object, object>`, `Exists<object>`, `Distinct<object>`, `WhereAny<object>`, `WhereSelect<object, object>`, `SelectWhere<object, object>`, `SelectDistinct<object, object>`; add `using Hl7.Cql.Compiler.Infrastructure;`. I applied exactly this and it compiled with 0 errors and 40/40 fusion tests passing. Secondary benefit: `ICqlOperatorsMethod` throws from a *static field initializer*, so a future second 2-parameter `Where`/`Select` overload becomes a `TypeInitializationException` at compiler startup instead of a compile error.

*Evidence:* `CqlOperatorsBinder.Fusion.cs:75-91`. Develop has just landed the opposite sweep in the same files — `git diff HEAD...origin/develop -- Cql/Cql.Compiler/CqlOperatorsBinder.Specific.cs` shows `typeof(ICqlOperators).GetMethod(nameof(ICqlOperators.Retrieve))!` replaced by `ReflectionUtility.GenericMethodDefinitionOf(() => default(ICqlOperators)!.Retrieve<object>(default!))`, so this branch reintroduces the pattern being removed.

#### 2. `WhereResultAsNonSourceArgument_DoesNotFuse` passes with the source-position guard removed entirely — `Cql/CoreTests/CqlOperatorsBinderFusionTests.cs:211`

The test uses `Except` as the consumer, and `Except` is not in the fusable consumer set, so control never reaches the position guard — the switch on `(consumerDefinition, producerDefinition)` returns null regardless. It cannot fail. This is the negative case AC #2 of issue #1484 explicitly requires. Either exercise the guard for real (construct the mismatched shape directly, e.g. `new CodeInvoke(OperatorsReceiver, SelectMethod.MakeGenericMethod(t, tr), [selectorLambda, whereInvoke])` through an internals-visible seam) or rename it to what it actually pins and add a comment recording that the guard is currently unreachable through `Exists`/`Select`/`Where`/`Distinct` because their only non-source parameter is a lambda — so the guard is defence-in-depth, not a live discriminator.

*Evidence:* I replaced `consumer.Arguments[0] is not CodeInvoke producer || !IsOperatorsCall(producer)` (Fusion.cs:103-106) with `consumer.Arguments.OfType<CodeInvoke>().FirstOrDefault(IsOperatorsCall)`, rebuilt, and got `Passed!  - Failed: 0, Passed: 11` on `CqlOperatorsBinderFusionTests` — this test included.

#### 3. Two checked-in `*.g.cs` example inputs were not regenerated and still contain unfused patterns — `Examples/CqlSdkExamples/230 Package CQL+ELM+C#+DLL from dirs, Save FHIR Resources to dir/input/cs/DevDays-2025.0.0.g.cs:96`

AC #4 and CLAUDE.md require every checked-in `*.g.cs` to be regenerated in the same PR. `Examples/CqlSdkExamples/230 Package CQL+ELM+C#+DLL from dirs, Save FHIR Resources to dir/input/cs/DevDays-2025.0.0.g.cs` is still at 5.1.0.0 with `Exists(Where(...))` at lines 96-97 and 132-133; `.../input/cs/FHIRHelpers-4.0.001.g.cs` is still at 5.1.4.0 with `Distinct(Select(...))` at line 175. Regenerate both. Not a runtime break — 5.1.x is inside the widened `[5.1.0.0, 5.3.0.0)` range — but the PR body's "all checked-in generated sets regenerated … 21 header-only files reverted" is inaccurate (23 files remain on 5.1.x, and 2 of them have genuinely fusable content).

*Evidence:* `grep -rl 'GeneratedCode(".NET Code Generation", "5\.1' --include=*.g.cs .` lists 23 files; inspecting the two example inputs shows live `context.Operators.Where<Coding>(…)` / `Operators.Exists<Coding>(g_)` pairs and `Operators.Distinct<CqlCode>(d_)` over a `Select`. Direct precedent that these files are in scope for regeneration: commit `460e7750c "fix: regenerate packaged example C# inputs"` touched exactly these two files after an earlier generator bump.

#### 4. "No evaluation is added or skipped" is false when a lambda throws — the fused form runs the *other* lambda for earlier elements — `Cql/Cql.Runtime/Operators/CqlOperators.FusedOperators.cs:26`

In `Select(Where(s,p),f)` the selector never runs at all if `p` throws, because `Where` completes first. In `WhereSelect` the selector has already run for every element before the throwing one, so its `Message` side effects fire where the composition emits none. Symmetric for `SelectWhere` when the selector throws. Practical impact is low (a CQL runtime error aborts the evaluation), but this doc is the safety argument for the whole optimization, so it must be exact. Narrow both statements to something like "identical unless one of the two lambdas throws, in which case the fused form may have invoked the other lambda for elements preceding the failure" — in the `<remarks>` here and in the fragment's Improvements paragraph (`docs/releases/vnext/1490-operator-fusion.md:23`). The existing `WhereSelect_PredicateThrows_MatchesComposition` / `SelectWhere_SelectorThrows_MatchesComposition` tests only compare the exception, not the traces, so they don't catch this.

*Evidence:* Probe run in the worktree (temporary test, since deleted). Source `["a","b"]`, predicate throws on `"b"`: `composed selector invocations: []` vs `fused    selector invocations: [f:a]`. Source `["a","b"]`, selector throws on `"b"`: `composed predicate invocations: []` vs `fused    predicate invocations: [p:1]`. Both asserts failed.

1. **Three of the four fused operators get no end-to-end execution coverage in the open test corpora** — `Cql/CoreTests/CSharp/` has 132 `SelectDistinct` call sites and zero `WhereAny`/`WhereSelect`/`SelectWhere`; `LibrarySets/RR23/CSharp` has 1 `SelectDistinct` and zero of the others. All 1,340 sites for the other three live in `Demo/Measures.dqm-content-qicore-2025/CSharp`, which is not in `Cql-Sdk.slnf` and whose runtime tests live in the private Integration Runner submodule — so a public-repo contributor can regress

### Notes — no action needed

- **`WhereAny` keeps re-testing `element != null` after the answer is already known** — Once `any` is true the null test is redundant; only the predicate must keep running (for side effects). `var matched = lambda(element) ?? false; if (!any && matched && element != null) any = true;` is the equivalent shape. Micro-optimisation in an optimisation PR — mentioning it only for completeness, no action needed.

### Considered and dropped

- **Adding four abstract members to the public `ICqlOperators` is a Breaking change for implementers and is not recorded as one** — REFUTED as written (severity `required`, remedy `## Breaking` section + `breaking change` label). The premise is true but the finding's own cited evidence contradicts its conclusion, and the change is in fact recorded.

WHAT IS TRUE: `git diff origin/develop...HEAD -- Cql/Cql.Runtime/Operators/ICqlO

alexzautke:
@baseTwo All four Required items and the Also item are addressed in `d1e951a9f`. Three were valid as written; one had a correct diagnosis but an impossible remedy, with harder evidence than the finding assumed.

### Required 1 — reflection through `ReflectionUtility` ✅ valid, fixed

All eight lookups now use `ReflectionUtility.GenericMethodDefinitionOf(() => default(ICqlOperators)!.X<…>(default!, …))`; `ICqlOperatorsMethod` is deleted. Confirmed each of the eight members has exactly one overload, so closing over `<object>`/`<object, object>` is unambiguous, and that `CqlOperatorsBinder.Specific.cs:240` is the precedent shape. Your secondary point is the sharper one: the old helper threw from a `static readonly` initializer, so an added overload was a startup `TypeInitializationException`; the expression form makes it a compile error.

### Required 2 — decorative negative test ⚠️ correct diagnosis, remedy impossible — fallback taken

The test genuinely cannot fail, as you showed. But the preferred fix cannot be built, and not merely because the binder blocks it — **the shape is unrepresentable in the IR**:

- All four fusable consumers take their list at argument 0. `Exists<T>` and `Distinct<T>` have exactly one parameter; `Where<T>` and `Select<T,TR>` declare a delegate as their only other parameter.
- `CodeInvoke`'s constructor runs `CodeTypeRules.ValidateAssignment` per argument, and an `IEnumerable<…>`-typed producer is not assignable to a `Func<…>` parameter — so the call **throws at construction**. Exposing `Fuse` through an internals seam would buy nothing.

I ran your probe verbatim (guard replaced with `consumer.Arguments.OfType<CodeInvoke>().FirstOrDefault(IsOperatorsCall)`): fusion suites unchanged at 41/41, full CoreTests **1208 passed / 8 skipped / 0 failed**, no golden-output change. Nothing in the suite bites that guard, confirming it is unreachable rather than merely untested. Probe reverted.

So, the fallback — but with an assertion rather than only a comment:
- Renamed to `WhereResultConsumedByNonFusableOperator_DoesNotFuse`, documenting the half it actually pins.
- Added `ProducerInANonSourceArgument_IsUnrepresentableInTheIR`, which asserts `Exists`/`Distinct` arity **and** that constructing `Select(source, whereInvoke)` / `Where(source, selectInvoke)` throws `ArgumentException`. That starts failing the day a fused consumer takes a list in a second position — exactly when the guard becomes load-bearing.
- Corrected the `Fusion.cs` paragraph: the `Except` example was a consumer-set case, not a source-position one.

### Required 3 — unregenerated example inputs ✅ regenerated for real

Both are now on **5.2.0.0** with fused calls: `DevDays-2025.0.0.g.cs` two `Exists(Where(…))` → `WhereAny<Coding>`, `FHIRHelpers-4.0.001.g.cs` `Distinct(Select(…))` → `SelectDistinct<Coding, CqlCode>` (plus SSA renumbering and the local-function elimination accumulated since 5.1.4). Regenerated through `ElmToolkit` from the checked-in ELM, not hand-edited. `grep` for 5.1.x across `*.g.cs` is now **21 files** — exactly the header-only set — and none of those were touched.

### Required 4 — the safety claim ✅ valid, narrowed, and now tested

Your probe values are the real behaviour. Both statements now say: identical while every lambda runs to completion; when one throws, the fused form has already invoked the *other* for the elements preceding the failure, with its `Message` side effects, where the composition never reaches it. Same exception surface either way.

Two tests added asserting the **actual fused traces** rather than parity: `WhereSelect_PredicateThrows_SelectorHasAlreadyRunForPrecedingElements` (fused `["f:a"]` vs composed `[]`) and the symmetric `SelectWhere_SelectorThrows_…` (fused `["p:1"]` vs composed `[]`). The release fragment's old wording ("only if *two different lambdas would both throw* can the surfacing exception differ") was both too strong and mis-stated the condition; replaced.

### Also — end-to-end coverage ✅ closed with a real CQL fixture

The ELM toolchain is available here, so no lesser fix was needed. Added `Cql/CoreTests/Input/ELM/HL7/FusedOperatorsTest.cql` whose generated output contains all four fused operators, and `FusedOperatorsCqlTest.cs` with five tests: four execute the generated library and assert real results (`[20,40,20]` uses `return all`, so a silent distinct fails), plus `GeneratedCSharp_UsesAllFourFusedOperators`, which regenerates from the checked-in ELM and asserts all four fused calls are present and no unfused `Where<`/`Select<`/`Exists<`/`Distinct<` survives. Break-tested: stubbing `Fuse` to return its input makes that last test fail while the other four still pass — which is exactly why the shape assertion is there.

One gotcha worth recording for whoever regenerates ELM next: `GenerateElmFiles`' incremental `Inputs`/`Outputs` means a partial run deletes all ELM but pretty-prints only the out-of-date subset. Every `.json` must be deleted first for a correctly formatted full run. After a full run the 13 pre-existing ELM files are byte-identical; restored, so no churn. The 10 header-only `CoreTests/CSharp` rewrites were likewise reverted per convention.

### Note — no action taken, deliberately

The redundant `element != null` retest in `WhereAny` after `any` is true: left as is. It is one predictable branch per element, and the current shape keeps the parity with `Exists`'s `(lambda(e) ?? false) && e != null` visible line-for-line, which is what makes the operator reviewable against its composition.

**Validation:** fusion suites **43/43** (baseline 40: +1 binder, +2 divergence); with the new fixture class **48/48**; full CoreTests net10.0 Release **1215 passed / 8 skipped / 0 failed**; `Cql-Sdk.slnf`, `Demo/Measures.dqm-content-qicore-2025` and `Examples/CqlSdkExamples` all build clean.

One correction to the PR body's numbers while I was here: the full-suite baseline in a current checkout is 1207 passed / 8 skipped, not the 1095 the body claims — that figure predates the develop merge.

---

baseTwo (changes requested):
## Re-review of `d1e951a9f` — all five findings fixed, but the doc narrowing overshot into a new false claim

@alexzautke — all five items from my last review are genuinely resolved, and two of them you took further than I
asked:

- All eight lookups now go through `ReflectionUtility.GenericMethodDefinitionOf`, and `ICqlOperatorsMethod` is gone.
- The two example inputs are regenerated to `5.2.0.0` and now carry fused output; 21 `.g.cs` remain on 5.1.x, all
  header-only.
- The guard test was renamed **and** backed by `ProducerInANonSourceArgument_IsUnrepresentableInTheIR`, which
  proves the shape cannot exist in the IR rather than merely being unreachable. That is a better answer than the
  one I suggested.
- The e2e coverage nit was closed with a real `FusedOperatorsCqlTest` plus CQL/ELM inputs.

Tests green: filtered and full `CoreTests` on net10.0.

### 1. Required — the exception claim is now false in a different way, and a correct caveat was deleted

Narrowing the "no evaluation is added or skipped" claim was right, but it went one step too far.
`CqlOperators.FusedOperators.cs:43-44` now ends with:

> The exception that surfaces is the same one either way — what differs is the work already done when it does.

and `docs/releases/vnext/1490-operator-fusion.md:24` says "The same exception still surfaces, but …". **Both are
false when the two lambdas would each throw on different elements.** In `WhereSelect` the fused loop calls
`select(element)` as soon as `lambda(element)` is true, so a selector throwing on an earlier element wins over a
predicate throwing on a later one.

Demonstrated over `["a","b"]` with a predicate throwing on `"b"` and a selector throwing on `"a"`:

```
composed = 'predicate-boom'
fused    = 'selector-boom'
```

The pre-fix fragment actually said this correctly — `git diff 9ece795c0..d1e951a9f -- docs/releases/vnext/1490-operator-fusion.md`
shows the deleted sentence "only if *two different lambdas would both throw* can the surfacing exception differ".
That caveat was right and should come back, in both places. Something like:

> …unless both lambdas would throw, in which case the fused form surfaces the earlier one it reaches.

This matters more than a doc nit because the fragment is the consumer-facing description of a potentially-breaking
codegen change — it is the one statement that most needs to be true. The existing exception-parity tests
(`FusedOperatorTests.cs:407-457`) only ever arm one lambda, so nothing in the suite would catch the difference; a
both-throw case would be worth adding alongside.

### 2. Nit — the renamed test's comment mis-states which half it pins

`CqlOperatorsBinderFusionTests.cs:205-210` says it "pins the consumer half of the match … not the source-position
guard". It pins neither: the test binds `Except(left, Where(right, predicate))` with `left` a `CodeConstant`, so
`Fuse` returns at the `consumer.Arguments[0] is not CodeInvoke producer` check and the consumer/producer switch is
never consulted. I confirmed it is insensitive to both mechanisms — adding a catch-all arm that removes the
consumer half entirely leaves this test passing (only `WhereOverWhere_HasNoFusedEquivalent_AndIsLeftAlone` fails).
The consumer half is really pinned by `WhereOverWhere_…`. Either reword the comment to say the shape is declined by
whichever check fires first and point at `WhereOverWhere_…`, or put the producer in argument 0 of a non-fusable
consumer that accepts a list there.

The test name is a true statement and the asserted behaviour is correct, so this is only the comment.

---

Item 1 is the only thing holding this. Everything else is done.

alexzautke:
@baseTwo Both items addressed in `b5a195535`.

### 1. Required — exception claim ✅ your both-throw case is real, caveat restored and tightened

Reproduced your probe before changing anything: `["a","b"]`, predicate throws on `"b"`, selector throws on `"a"` → composed `predicate-boom`, fused `selector-boom`. The statement in both places now has three tiers, each exactly true:

- every lambda completes → identical results, `Message` side effects and exception surfaces;
- exactly one lambda throws → the same exception surfaces either way, and only the work already done differs (the part of the previous wording that was correct is kept verbatim);
- **both lambdas would throw, on different elements → the surfacing exception itself can differ: the fused form surfaces whichever throw its single pass reaches first, which need not be the one the composition reaches.** The `<remarks>` extend the existing `["a","b"]` example with the selector also throwing on `"a"`, so the divergence is concrete in the doc itself.

The restored caveat is deliberately not the deleted sentence verbatim — "only if two different lambdas would both throw can the surfacing exception differ" stated the trigger but not the direction; the new wording adds which exception wins.

Two divergence tests added alongside the existing single-throw ones, asserting the *actual* exception of each form rather than parity: `WhereSelect_BothLambdasThrow_FusedSurfacesTheSelectorsThrowFromTheEarlierElement` and the symmetric `SelectWhere_BothLambdasThrow_FusedSurfacesThePredicatesThrowFromTheEarlierElement`. Break-tested: rewriting either to assert exception parity fails with `Expected:<predicate-boom>. Actual:<selector-boom>` (and the mirror image), then restored.

### 2. Nit — test comment ✅ reworded

You're right that the `Except` shape never reaches the consumer/producer switch — `Fuse` returns at `consumer.Arguments[0] is not CodeInvoke producer` because `left` is a `CodeConstant`. The comment now says the shape is declined at whichever structural check fires first, that the test pins the outcome (a non-fusable consumer is left alone) rather than any one check, and points at `WhereOverWhere_HasNoFusedEquivalent_AndIsLeftAlone` for the consumer half and `ProducerInANonSourceArgument_IsUnrepresentableInTheIR` for the source-position guard. Name and assertions untouched. I checked the alternative you offered — a producer in argument 0 of a non-fusable list-accepting consumer — but any such shape either enters the fusable set or is already covered by `WhereOverWhere_…`, so the reword is the honest fix.

**Validation:** fusion suites 50/50 (baseline 48 + 2), full CoreTests net10.0 Release **1217 passed / 8 skipped / 0 failed**.

---

baseTwo (approved):
## Approving `b5a195535` — the exception semantics are now stated exactly, and pinned by tests

@alexzautke — both items from the last round are resolved, and the required one is resolved properly rather than
by softening the wording until it was unfalsifiable.

**The claim is now correct.** The `<remarks>` and the fragment both say the same thing: while only one lambda
throws the surfacing exception is shared and only the work already done differs; when both would throw on
different elements, the fused form surfaces whichever throw its single pass reaches first, which need not be the
one the composition reaches. That is precisely the behaviour, and the remarks even extend the running `["a","b"]`
example to show it.

**And it is now pinned.** `WhereSelect_BothLambdasThrow_FusedSurfacesTheSelectorsThrowFromTheEarlierElement` and
its `SelectWhere` twin assert `composed = "predicate-boom"` against `fused = "selector-boom"` — the exact pair I
reproduced when raising this. The suite would now catch a regression here, which it could not before, since every
prior exception-parity test armed only one lambda.

**The test comment is fixed correctly too.** It no longer claims to pin the consumer half; it says the shape is
declined by whichever structural check fires first, and points at `WhereOverWhere_HasNoFusedEquivalent_AndIsLeftAlone`
and `ProducerInANonSourceArgument_IsUnrepresentableInTheIR` for the two mechanisms individually. That is an honest
description of an over-determined test rather than a rename that papers over it.

Re-confirmed from the previous round: all eight lookups still go through `ReflectionUtility`, and 21 `.g.cs` remain
on 5.1.x — all header-only, with the two that carried genuinely fusable content regenerated to 5.2.0.0.

Tests: fusion suites 50/50 locally; all 8 checks green at this head.

Worth saying plainly, since this PR took several rounds: every finding here was answered by making the code or the
documentation more precise, never by weakening a test or narrowing a claim to dodge it. The `ProducerInANonSourceArgument_IsUnrepresentableInTheIR`
answer in particular was better than the fix I proposed.

PR #1492, opened 2026-08-03
Bump Integration Runner submodule to current main

alexzautke:
## Primary Work

Advances the `submodules/Firely.Cql.Sdk.Integration.Runner` gitlink from `b2d076eb` to `4c7f3a02`, moving the recorded pointer onto the runner's `main` proper. This is the hygiene step that stages the superproject for the bump that will pick up runner PR FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108 — **it does not contain that work**.

> **Update.** `4c7f3a02` was `main`'s tip when this PR was opened. That runner PR has since merged, so `main` is now at `37c14091a` and `4c7f3a02` sits one merge behind it. This PR is unaffected and still does its job: after it lands, the follow-up bump reads as a clean `main` → `main` move from `4c7f3a02` to `37c14091a` whose diff is exactly that PR's merge — which was the entire point of separating the hygiene step. The follow-up bump and the solution wiring it needs are tracked in #1495.

**What the pointer was, precisely.** `b2d076eb` was the pre-merge tip of the runner's `claude/overlaps-after-passes` branch (runner PR FirelyTeam/Firely.Cql.Sdk.Integration.Runner#107, the `passed.jsonl` bump for the 15 tests fixed by the `OverlapsAfter` null-boundary fix). Since that PR merged, `b2d076eb` is the second parent of `4c7f3a02` and therefore already an ancestor of `main` — nothing was broken or unreachable. The pointer was simply sitting on an off-mainline commit mid-history instead of on a branch tip.

**Why move it anyway, and why now.** Because `b2d076eb` had itself already merged `origin/main`, its tree is identical to `4c7f3a02`'s: `git compare` reports the two commits one apart with **zero files changed**. So the move costs nothing content-wise, and it buys the thing the next bump needs — a pointer on a stable branch tip, so that picking up FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108 reads as a clean `main` → `main` gitlink move whose diff is exactly FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108's merge, rather than a move off a merge parent that mixes "catch up to mainline" with "adopt the new work". Doing that hygiene separately keeps the FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108 bump reviewable as the single decision it is.

**What the follow-up bump will bring in.** Runner PR FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108 makes the instruments from the CQL-execution performance campaign durable parts of the runner repo, alongside the existing `IntegrationRunner.Benchmarks`:

- **`IntegrationRunner.Harness`** — in-process steady-state A/B measurement: N corpus cases × M reps in one process, one line per rep (wall / CPU / allocation / lock contention / GC counts and pause), with parallel evaluation and a realized-concurrency counter.
- **`IntegrationRunner.Verify`** — a corpus-wide behavioral oracle: every case's own `MeasureReport` comparison recorded to a status file whose SHA-256 is a behavior fingerprint, so a supposedly behavior-preserving refactor is gated on the hash matching byte-for-byte.
- **Env-gated bundle cache in `CqlTestCaseLoader`** (`CQL_CMS_TEST_RUNNER_CacheBundles` = `share` | `copy` | unset) — removes the fixture cost (~55% of measured wall, ~48% of allocation) that otherwise drowns SDK-level effects; unset is today's behavior verbatim.
- **A methodology README** covering tiered-JIT convergence, medians over the steady tail, order-balanced interleaved rounds, and which metrics are trustworthy at what resolution.

That matters to this repo specifically because `Cql-Sdk-All.slnx` / `Cql-Sdk-All.sln` already reference `IntegrationRunner.Benchmarks` through the submodule path, and the [`run-integration-benchmarks`](.claude/skills/run-integration-benchmarks/SKILL.md) skill drives it from here. Once FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108 merges and the pointer advances, `Harness` and `Verify` become available from an SDK working tree on the same footing — the measurement and behavior-preservation instruments for SDK-side performance work land where the SDK is actually built.

---

## Auxiliary Work

None. Single-gitlink change; no superproject source, project, or build-script edits.

## Related issues

None — submodule housekeeping. Tracks runner PR FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108 as the follow-on work.

## Testing

Pointer-only change (one gitlink), so there is nothing to exercise beyond verifying the pointer itself:

- `4c7f3a02` is on the runner's `main` (it was `main`'s tip when this PR was opened; `main` has since advanced to `37c14091a`).
- `b2d076eb` is an ancestor of `4c7f3a02` (its second parent), so this is a forward move, not a re-point.
- `b2d076eb...4c7f3a02` compares as **one commit ahead, zero files changed** — the submodule tree is byte-identical before and after, confirming no runner content enters the superproject with this bump.
- The superproject diff is the single `submodules/Firely.Cql.Sdk.Integration.Runner` gitlink line; no other tracked file changes.

## FirelyTeam Checklist


baseTwo:
## Review — submodule pointer bump

Verified the pointer itself, and reviewed the submodule work this bump stages for (FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108) from the API rather than locally.

**The bump is correct and safe to merge.**

- `4c7f3a02` is the current tip of the runner's `main`, and `main` is the only branch whose head it is.
- `b2d076eb` is the second parent of `4c7f3a02`, so this is a forward move along `main`'s history, not a re-point.
- `b2d076eb...4c7f3a02` compares as **1 commit ahead, 0 files changed** — the submodule tree is byte-identical before and after, so no runner content enters the superproject here.
- The superproject diff is exactly one gitlink line; nothing else is touched. All 8 Azure checks pass.

### One follow-up to carry into the next bump (not a change request on this PR)

`Cql-Sdk-All.slnx` (line 87) and `Cql-Sdk-All.sln` reference `IntegrationRunner.Benchmarks.csproj` through the submodule path, and the [`run-integration-benchmarks`](.claude/skills/run-integration-benchmarks/SKILL.md) skill drives it from this repo. Runner FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108 adds two further projects — `IntegrationRunner.Harness` and `IntegrationRunner.Verify` — to the submodule's own `IntegrationRunner-All.sln`, but the superproject solutions will not know about them.

So the bump that adopts FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108 should either add both projects to `Cql-Sdk-All.slnx`/`.sln` (matching the `IntegrationRunner.Benchmarks` precedent) or record why they are deliberately left out. Worth deciding then rather than discovering it later.

No TFM concern: both new projects are `net10.0`-only, the same as `IntegrationRunner.Benchmarks` and `IntegrationRunner` itself, so the .NET 8 test job is unaffected.

### On the submodule work

I posted a full review with 9 findings on FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108. The two that matter most for this repo, since these instruments are intended to gate SDK-side performance refactors:

1. **The behavioural fingerprint is a sequential-only guarantee.** `IntegrationRunner.Verify` has no `dop` argument and runs a plain sequential `foreach`, while the harness's headline capability is parallel evaluation. So `share` mode + `dop > 1` — the combination the README recommends for SDK measurement — is unverified. That matters here specifically because `share` hands one `Bundle` POCO to every concurrent case, and our retrieve path is actively growing per-bundle caches (#1451's coding cache, #1467's shared bundle index).
2. **The harness aborts the whole run on the first case exception**, losing every completed repetition's output — most likely to fire exactly when A/B-testing a change that breaks a case.

Neither blocks this pointer bump, since FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108 is not in it.

baseTwo:
@copilot Please extend this PR to wire the Integration Runner's two new measurement projects into the superproject solutions, so they are reachable from an SDK working tree the same way `IntegrationRunner.Benchmarks` already is.

### Read this first — there is a hard ordering dependency

The two projects **do not exist at the commit this PR currently pins.** `4c7f3a02` is the merge of runner PR FirelyTeam/Firely.Cql.Sdk.Integration.Runner#107; `IntegrationRunner.Harness` and `IntegrationRunner.Verify` are added by runner PR FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108, which is still open.

So adding solution references now, on top of the current pointer, would point at paths that are not on disk and would break both solutions for everyone. The sequence has to be:

1. Runner FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108 merges (it currently has 13 requested changes from my review — please address those first).
2. Advance `submodules/Firely.Cql.Sdk.Integration.Runner` to the new `main` tip, which then contains both projects.
3. Add the solution entries described below, in the same commit or a follow-up commit on this branch.
4. Verify both solutions still load and build.

Please don't do step 3 before step 2. If you'd rather keep this PR as the pointer-hygiene change it is today and do the wiring in the follow-up bump PR, that's fine too — say so and I'll retarget the ask.

### What to add

Two projects, from the submodule:

- `submodules/Firely.Cql.Sdk.Integration.Runner/IntegrationRunner.Harness/IntegrationRunner.Harness.csproj`
- `submodules/Firely.Cql.Sdk.Integration.Runner/IntegrationRunner.Verify/IntegrationRunner.Verify.csproj`

### Where to add it

**1. `Cql-Sdk-All.slnx`** — into the existing `<Folder Name="/submodules/Firely.Cql.Sdk.Integration.Runner/">` block (lines 84-89). Entries in that folder are path-sorted, so both go *after* `IntegrationRunner.Benchmarks/...` and *before* `IntegrationRunner/IntegrationRunner.csproj` (`.` sorts before `/`):

```xml
<Project Path="submodules/Firely.Cql.Sdk.Integration.Runner/IntegrationRunner.Benchmarks/IntegrationRunner.Benchmarks.csproj" />
<Project Path="submodules/Firely.Cql.Sdk.Integration.Runner/IntegrationRunner.Harness/IntegrationRunner.Harness.csproj" />
<Project Path="submodules/Firely.Cql.Sdk.Integration.Runner/IntegrationRunner.Verify/IntegrationRunner.Verify.csproj" />
<Project Path="submodules/Firely.Cql.Sdk.Integration.Runner/IntegrationRunner/IntegrationRunner.csproj" />
```

**2. `Cql-Sdk-All.sln`** — follow the `IntegrationRunner.Benchmarks` precedent exactly. For each new project that means three edits:

- A `Project(...) = ...` entry next to line 142, with **freshly generated** GUIDs (do not reuse the GUIDs from the submodule's own `IntegrationRunner-All.sln` — they are a different solution's identifiers), using backslash path separators like the neighbouring entries.
- Config mappings in `ProjectConfigurationPlatforms`, next to lines 504-515.
- A `NestedProjects` entry next to line 572, mapping each new project GUID to `{0E5F5704-5289-4C65-AD4E-364CD5F60C44}` (the `Firely.Cql.Sdk.Integration.Runner` solution folder), so they don't land at the solution root.

⚠️ **One thing to get right:** this solution's configuration set is *not* the same as the submodule's. `Cql-Sdk-All.sln` gives each project **6** mappings (`Debug`/`Release` × `Any CPU`/`x64`/`x86`), whereas `IntegrationRunner-All.sln` gives **24**, because it also carries `Debug Use Project References` and `FullDebug`. Copy the 6-mapping shape from `{1200F914-0318-41E4-80CF-29BEF88DA7FE}` (lines 504-515), not the 24-mapping shape from the submodule PR.

### Where *not* to add it

Leave `Cql-Sdk.slnf` and `Cql-Sdk-Demos-Examples.slnf` alone. Neither references any submodule project today, and `Cql-Sdk.slnf` is the CI build target per `CLAUDE.md` — these are developer-facing measurement tools, not part of the SDK build. This mirrors the submodule side, where FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108 correctly leaves `IntegrationRunner.slnf` untouched.

### Verification

- `dotnet build Cql-Sdk-All.sln` and confirm both new projects build (they are `net10.0`-only, same as `IntegrationRunner.Benchmarks` and `IntegrationRunner` itself, so the .NET 8 job is unaffected).
- Confirm `Cql-Sdk-All.slnx` still loads and that the two projects appear under the `submodules/Firely.Cql.Sdk.Integration.Runner` folder rather than at the root.
- Keep the two solution files consistent with each other — same four-plus-two project set in both.

No release-note fragment needed: solution membership is not a public API, generated-output, Packager CLI or build behaviour change, and `Cql-Sdk.slnf` (what CI builds) is untouched.

Finally, please update this PR's description to cover the added scope — it currently describes a pointer-only change.

baseTwo:
@copilot Agreed — deferring is the right call, and it's the alternative I offered. Your reasoning is correct: the two projects don't exist at `4c7f3a02`, so wiring them here would reference paths that aren't on disk.

Two corrections to my earlier ask, then how to proceed.

### Corrections to what I asked for

1. **Ignore my last line about updating the PR description.** That was conditional on this PR taking the extra scope. It isn't, so the description stays as a pointer-only change — I've already rewritten it, no action needed from you.
2. **Reference convention, please adopt it going forward.** A bare `#108` written in *this* repo resolves to `FirelyTeam/firely-cql-sdk#108` — a closed PR titled "Add a public non-generic Convert on TypeConverter", nothing to do with the runner. `#107` likewise resolves to a Dependabot bump. Your reply used the bare form, and so did my ask; I've since rewritten this PR's body and both of my comments to use the fully qualified `FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108` form throughout. Please use `owner/repo#number` for anything in the submodule repo (and `FirelyTeam/firely-cql-sdk#number` when writing in the runner repo).

### How to proceed

**This PR:** nothing further. It stays a one-line gitlink change, it is `MERGEABLE`, all 8 Azure checks pass, and it is waiting only on a human approval — so it can merge independently of the runner work. Please leave it alone rather than holding it.

**The deferred wiring needs somewhere to live.** Right now it exists only in a PR comment, which is exactly how deferred work gets lost. Please open a tracking issue in `FirelyTeam/firely-cql-sdk` that captures it, following the repo's issue conventions:

- Title along the lines of *"Add IntegrationRunner.Harness and IntegrationRunner.Verify to the Cql-Sdk-All solutions"*.
- Body: carry over the full spec from [my earlier comment](https://github.com/FirelyTeam/firely-cql-sdk/pull/1492#issuecomment-5166018946) verbatim — both file targets, the path-sorted `.slnx` position, the three `.sln` edits (Project entry with fresh GUIDs, config mappings, `NestedProjects` under `{0E5F5704-5289-4C65-AD4E-364CD5F60C44}`), the 6-mapping-not-24-mapping warning, the two `.slnf` files to leave alone, and the verification steps.
- State the blocking dependency explicitly: it cannot start until `FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108` merges and the submodule pointer advances to a commit containing it.
- Reference this PR and that runner PR (fully qualified), and add acceptance criteria as checkboxes so they can be ticked as the work lands.

Then link the issue back here in a comment so the trail is complete.

**Priority order.** The tracking issue is bookkeeping and takes minutes. The real work is the **13 requested changes** on `FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108` — please treat those as the priority, and in particular the two architectural ones rather than the local fixes:

- **F1** — `IntegrationRunner.Verify` has no `dop` argument and runs a sequential `foreach`, so every fingerprint in that PR (including the `507669b5a173…` baseline) only establishes behaviour preservation for sequential execution. `share` + `dop > 1` is the mode the README recommends for SDK measurement and it is unverified — which matters because `share` hands one `Bundle` POCO to every concurrent case while our retrieve path is actively growing per-bundle caches (#1451, #1467).
- **F2** — the harness has no per-case `catch`, so one exception discards an entire run's output including every already-completed repetition.

Those two decide whether these instruments can be trusted to gate SDK performance refactors at all, so they're worth more care than the nits.

baseTwo (approved):
Approving. Verified independently rather than from the description:

- `b2d076eb` is the second parent of `4c7f3a02`, so this is a forward move along `main`'s history, not a re-point.
- `b2d076eb...4c7f3a02` compares as **1 commit ahead, 0 files changed** — the submodule tree is byte-identical before and after, so no runner content enters the superproject here.
- The superproject diff is exactly one gitlink line; no other tracked file changes.
- All 8 Azure checks pass.

**One note on timing, which I've also corrected in the description.** `4c7f3a02` was `main`'s tip when this PR was opened, but `FirelyTeam/Firely.Cql.Sdk.Integration.Runner#108` merged a few minutes ago, so `main` is now at `37c14091a` and this pointer is one merge behind.

That does not undermine this PR — it is exactly the scenario it was built for. Landing `4c7f3a02` means the follow-up bump is a clean `main` → `main` move to `37c14091a` whose diff is precisely that runner PR's merge, instead of one move mixing "catch up to mainline" with "adopt the new work". Separating the two was the whole rationale, and it still holds.

The follow-up bump, plus wiring `IntegrationRunner.Harness`, `IntegrationRunner.Verify` and the new shared `IntegrationRunner.Measurement` library into `Cql-Sdk-All.sln` / `Cql-Sdk-All.slnx`, is tracked in #1495 — now unblocked.

PR #1499, opened 2026-08-04
Fix FHIR stratifier invariant violation in generated measures

### LINKED ISSUE #1500: Generated Measure stratifiers violate the FHIR code/description xor component invariant
## Summary

`Measure` resources generated from `@stratifier` annotations are invalid: the container stratifier is emitted with `code` and `description` set *and* with `component` entries, violating the FHIR invariant on `Measure.group.stratifier`:

```
group.stratifier.all((code | description | criteria).exists() xor component.exists())
```

A stratifier must be either a single criteria (code/description/criteria) or a set of criteria components — never both.

## Impact

Every generated Measure that uses `@stratifier` annotations fails FHIR validation against this invariant, making the Packager's Measure output unusable with validating consumers.

## Observed Behavior

`GetOrCreateStratifier` in `Cql/Cql.Packaging/FhirMeasureExtensions.cs` creates the per-group container stratifier (`<group>-Stratifier`) with `Code` (text-only) and `Description` both set to the container id, and the `@stratifier` dimensions are then added as `component` entries under that same container.

## Repro Steps

1. Package a CQL library containing a definition annotated with `@group` and `@stratifier` (plus the `@measure`/`@year` tags required for Measure generation).
2. Validate the generated `Measure` resource.
3. Validation fails on the stratifier invariant because `group.stratifier` has `code`, `description`, and `component` simultaneously.

## Expected Behavior

Since the generated container stratifier always holds components, it must not also carry `code`, `description`, or `criteria` — only its element id. The components themselves keep their code, description, and criteria.

## Acceptance Criteria

- [x] The container stratifier is emitted with only its element id (`<group>-Stratifier`); no `code`, `description`, or `criteria`.
- [x] Stratifier components are unchanged (id, text-only code, description, `text/cql-identifier` criteria).
- [x] Tests assert the invariant holds for all generated stratifiers.
- [x] `docs/cql-packager.md` documents the container shape, and a release-note fragment records the output change.

alexzautke:
Fixes #1500

## Primary Work

Generated `Measure` resources violated the FHIR R4 invariant `mea-1` on `Measure.group.stratifier`:

```
group.stratifier.all((code | description | criteria).exists() xor component.exists())
```

The Packager set `Code` and `Description` on the per-group container stratifier (`<group>-Stratifier`) while also adding the `@stratifier` dimensions as `component` entries under it — so every generated Measure with stratifiers had `code`, `description`, and `component` simultaneously and failed validation.

Changes (`Cql/Cql.Packaging/FhirMeasureExtensions.cs`):

- The container stratifier now carries only its element id; `Code` and `Description` are no longer set. The components are unchanged: they keep their id (`<group>-StratifierComponent-<value>`), text-only code, description, and `text/cql-identifier` criteria — the "set of criteria components" side of the invariant's XOR.
- The container also cannot be emitted **without** components (which would satisfy neither side of the XOR): `GetOrCreateStratifier` was replaced by `AddStratifierComponent`, which runs the duplicate-component check first and only then creates the container together with its first component, so no code path can leave a componentless container behind.

Tests (`Cql/CoreTests/Packaging/FhirMeasureExtensionsTests.cs`):

- `GeneratedStratifiers_SatisfyFhirStratifierInvariant` evaluates the **literal FhirPath invariant expression** against the generated Measure via `FhirPathCompiler` (no hand-rolled C# translation), plus targeted per-half assertions so a failure identifies which side broke — including that every container has ≥ 1 component.
- `SingleStratifierDefinition_CreatesContainerWithOneComponent` asserts the container's `Code`/`Description` are null.
- All 22 tests in the class pass.

End-to-end coverage:

- `LibrarySets/Demo/Cql/MeasureExample.cql` gained a `@stratifier`-tagged definition (`Gender Stratifier`), with its ELM (`LibrarySets/Demo/Elm/MeasureExample.json`) and generated C# (`Demo/Measures.Demo/CSharp/MeasureExample-0.0.1.g.cs`) regenerated — so a packaged artifact now exercises the stratifier path, and the packaged `Measure-MeasureExample-0.0.1.json` was verified to satisfy the invariant. Only files with real content changes are committed; header-only `GeneratedCodeAttribute` churn was reverted per repo convention.

---

## Auxiliary Work

- **`docs/cql-packager.md`** — documents that the container is id-only and why; that the human-readable labels live on `stratifier.component.code`/`.description`; and that the components form a single composite stratification (a `MeasureReport` stratum combines all the group's dimension values), with links to the R4 definitions and CQF Measures IG.
- **`docs/releases/vnext/1499-measure-stratifier-invariant.md`** — release-note fragment with a `## Breaking Changes` section (generated `Measure` JSON shape change) and `## Fixes`.
- **`Demo/Cql/Build/CqlTooling.Targets.xml`** — fixed CQL-to-ELM generation on Linux/macOS: the Java classpath wildcard is now quoted so Unix shells don't expand it (previously the first dependency jar was misread as the main class).

Review by @baseTwo (5 required findings) is fully addressed in d22a467; threads answered and resolved. Copilot's #1501 was opened for the same findings and is redundant.

https://claude.ai/code/session_01TKXe5yP5wzM1PkxPcaLEWd

### DIFF
--- a/Cql/CoreTests/Packaging/FhirMeasureExtensionsTests.cs
+++ b/Cql/CoreTests/Packaging/FhirMeasureExtensionsTests.cs
@@ -10,6 +10,7 @@
 
 using Hl7.Cql.Elm;
 using Hl7.Cql.Packaging;
+using Hl7.FhirPath;
 using ElmAnnotation = Hl7.Cql.Elm.Annotation;
 using ElmLibrary = Hl7.Cql.Elm.Library;
 using FhirLibrary = Hl7.Fhir.Model.Library;
@@ -269,11 +270,17 @@ public class FhirMeasureExtensionsTests
             "the definition should also contribute a stratifier component");
     }
 
+    /// <summary>
+    /// The FHIR R4 invariant mea-1 on Measure ("Stratifier SHALL be either a single
+    /// criteria or a set of criteria components"), evaluated as the literal FhirPath
+    /// expression rather than a hand-rolled C# translation of it.
+    /// </summary>
+    private const string StratifierInvariant =
+        "group.stratifier.all((code | description | criteria).exists() xor component.exists())";
+
     [TestMethod]
     public void GeneratedStratifiers_SatisfyFhirStratifierInvariant()
     {
-        // FHIR invariant on Measure.group.stratifier:
-        // (code | description | criteria).exists() xor component.exists()
         var measure = CreateMeasure(BaseStatements(
             Def("Region Stratifier",
                 CreateTag("group", "RateA"),
@@ -284,15 +291,21 @@ public class FhirMeasureExtensionsTests
                 CreateTag("stratifier", "AgeBand"),
                 CreateTag("description", "Stratifies by age band"))));
 
+        new FhirPathCompiler().Compile(StratifierInvariant)
+            .Predicate(measure, new EvaluationContext())
+            .Should().BeTrue("generated measures must satisfy invariant mea-1: {0}", StratifierInvariant);
+
+        // Targeted halves of the invariant, so a failure identifies which side broke:
+        // the container carries no own content, while every component is fully populated.
         var stratifiers = measure.Group.SelectMany(g => g.Stratifier).ToList();
         stratifiers.Should().NotBeEmpty();
-        foreach (var stratifier in stratifiers)
-        {
-            var hasOwnContent = stratifier.Code != null || stratifier.Description != null || stratifier.Criteria != null;
-            var hasComponents = stratifier.Component.Count > 0;
-            (hasOwnContent ^ hasComponents).Should().BeTrue(
-                "stratifier {0} must have either code/description/criteria or components, never both", stratifier.ElementId);
-        }
+        stratifiers.Should().OnlyContain(s => s.Code == null && s.Description == null && s.Criteria == null,
+            "container stratifiers must not carry code/description/criteria");
+        stratifiers.Should().OnlyContain(s => s.Component.Count > 0,
+            "a container with no components would satisfy neither side of the invariant's xor");
+        stratifiers.SelectMany(s => s.Component).Should().OnlyContain(
+            c => c.Code != null && c.Description != null && c.Criteria != null,
+            "the stratification's meaning lives on the components");
     }
 
     [TestMethod]
--- a/Cql/CoreTests/Packaging/FhirMeasureExtensionsTests.cs
+++ b/Cql/CoreTests/Packaging/FhirMeasureExtensionsTests.cs
@@ -81,8 +81,8 @@ public class FhirMeasureExtensionsTests
         var group = measure.Group.Single(g => g.ElementId == "RateA");
         var container = group.Stratifier.Should().ContainSingle().Subject;
         container.ElementId.Should().Be("RateA-Stratifier");
-        container.Code!.Text.Should().Be("RateA-Stratifier");
-        container.Description.Should().Be("RateA-Stratifier");
+        container.Code.Should().BeNull("a stratifier with components must not also have a code (FHIR stratifier invariant)");
+        container.Description.Should().BeNull("a stratifier with components must not also have a description (FHIR stratifier invariant)");
         container.Criteria.Should().BeNull("the container stratifier only holds components");
 
         var component = container.Component.Should().ContainSingle().Subject;
@@ -267,6 +267,32 @@ public class FhirMeasureExtensionsTests
         var component = group.Stratifier.Single().Component.Single();
         component.Criteria.Expression_.Should().Be("Denominator Stratifier",
             "the definition should also contribute a stratifier component");
+    }
+
+    [TestMethod]
+    public void GeneratedStratifiers_SatisfyFhirStratifierInvariant()
+    {
+        // FHIR invariant on Measure.group.stratifier:
+        // (code | description | criteria).exists() xor component.exists()
+        var measure = CreateMeasure(BaseStatements(
+            Def("Region Stratifier",
+                CreateTag("group", "RateA"),
+                CreateTag("group", "RateB"),
+                CreateTag("stratifier", "Region")),
+            Def("Age Band Stratifier",
+                CreateTag("group", "RateA"),
+                CreateTag("stratifier", "AgeBand"),
+                CreateTag("description", "Stratifies by age band"))));
+
+        var stratifiers = measure.Group.SelectMany(g => g.Stratifier).ToList();
+        stratifiers.Should().NotBeEmpty();
+        foreach (var stratifier in stratifiers)
+        {
+            var hasOwnContent = stratifier.Code != null || stratifier.Description != null || stratifier.Criteria != null;
+            var hasComponents = stratifier.Component.Count > 0;
+            (hasOwnContent ^ hasComponents).Should().BeTrue(
+                "stratifier {0} must have either code/description/criteria or components, never both", stratifier.ElementId);
+        }
     }
 
     [TestMethod]
--- a/Cql/Cql.Packaging/FhirMeasureExtensions.cs
+++ b/Cql/Cql.Packaging/FhirMeasureExtensions.cs
@@ -231,11 +231,12 @@ internal static partial class FhirMeasureExtensions
         if (existing != null)
             return existing;
 
+        // Only the element id is set: the FHIR invariant on Measure.group.stratifier
+        // ((code | description | criteria).exists() xor component.exists()) forbids
+        // code/description/criteria on a stratifier that holds components.
         var container = new FhirMeasure.StratifierComponent
         {
             ElementId = id,
-            Code = new CodeableConcept { Text = id },
-            Description = id,
         };
         group.Stratifier.Add(container);
         return container;
--- a/Cql/Cql.Packaging/FhirMeasureExtensions.cs
+++ b/Cql/Cql.Packaging/FhirMeasureExtensions.cs
@@ -224,22 +224,27 @@ internal static partial class FhirMeasureExtensions
         }
     }
 
-    private static FhirMeasure.StratifierComponent GetOrCreateStratifier(FhirMeasure.GroupComponent group)
+    private static void AddStratifierComponent(FhirMeasure.GroupComponent group, FhirMeasure.ComponentComponent component)
     {
         var id = $"{group.ElementId}-Stratifier";
-        var existing = group.Stratifier.FirstOrDefault(s => s.ElementId == id);
-        if (existing != null)
-            return existing;
+        var container = group.Stratifier.FirstOrDefault(s => s.ElementId == id);
+        if (container?.Component.Any(c => c.ElementId == component.ElementId) == true)
+            throw new InvalidOperationException($"Stratifier component {component.ElementId} is defined twice for this measure.");
 
-        // Only the element id is set: the FHIR invariant on Measure.group.stratifier
-        // ((code | description | criteria).exists() xor component.exists()) forbids
-        // code/description/criteria on a stratifier that holds components.
-        var container = new FhirMeasure.StratifierComponent
+        if (container == null)
         {
-            ElementId = id,
-        };
-        group.Stratifier.Add(container);
-        return container;
+            // Only the element id is set: the FHIR invariant on Measure.group.stratifier
+            // ((code | description | criteria).exists() xor component.exists()) forbids
+            // code/description/criteria on a stratifier that holds components — and a
+            // container with no components satisfies neither side of the xor, so the
+            // container is only ever created together with its first component.
+            container = new FhirMeasure.StratifierComponent
+            {
+                ElementId = id,
+            };
+            group.Stratifier.Add(container);
+        }
+        container.Component.Add(component);
     }
 
     private static void AnnotateMeasureStratifiers(FhirMeasure fhirMeasure, ElmLibrary library, string? measureGroupCodeSystem)
@@ -276,13 +281,9 @@ internal static partial class FhirMeasureExtensions
             foreach (var tuple in tuples)
             {
                 var group = GetOrCreateGroup(fhirMeasure, tuple.Group, measureGroupCodeSystem);
-                var container = GetOrCreateStratifier(group);
 
                 var componentId = $"{tuple.Group}-StratifierComponent-{tuple.Stratifier}";
-                if (container.Component.Any(c => c.ElementId == componentId))
-                    throw new InvalidOperationException($"Stratifier component {componentId} is defined twice for this measure.");
-
-                container.Component.Add(new FhirMeasure.ComponentComponent
+                AddStratifierComponent(group, new FhirMeasure.ComponentComponent
                 {
                     ElementId = componentId,
                     Code = new CodeableConcept { Text = tuple.Stratifier },
--- a/docs/cql-packager.md
+++ b/docs/cql-packager.md
@@ -216,7 +216,10 @@ ELM `annotation` element. See `LibrarySets/Demo/Cql/MeasureExample.cql` for a co
 
 All stratifier-tagged definitions of a group collapse into a single container stratifier
 (`<group>-Stratifier`) whose dimensions are `stratifier.component` entries — including when a group
-has only one dimension. Each component gets id `<group>-StratifierComponent-<value>`, a text-only
+has only one dimension. The container carries only its element id — no `code`, `description`, or
+`criteria` — because the FHIR invariant on `Measure.group.stratifier`
+(`(code | description | criteria).exists() xor component.exists()`) forbids combining them with
+components. Each component gets id `<group>-StratifierComponent-<value>`, a text-only
 code with the `@stratifier` value, and a `text/cql-identifier` criteria referencing the definition.
 The definition's return value is the stratum value for the subject; any CQL type is allowed.
 
--- a/docs/cql-packager.md
+++ b/docs/cql-packager.md
@@ -219,9 +219,20 @@ All stratifier-tagged definitions of a group collapse into a single container st
 has only one dimension. The container carries only its element id — no `code`, `description`, or
 `criteria` — because the FHIR invariant on `Measure.group.stratifier`
 (`(code | description | criteria).exists() xor component.exists()`) forbids combining them with
-components. Each component gets id `<group>-StratifierComponent-<value>`, a text-only
-code with the `@stratifier` value, and a `text/cql-identifier` criteria referencing the definition.
+components. The human-readable labels therefore live on the components: each component gets id
+`<group>-StratifierComponent-<value>`, a text-only code with the `@stratifier` value (and the
+`@description` value as its description), and a `text/cql-identifier` criteria referencing the
+definition — display or key on `stratifier.component.code`/`.description`, not on the container.
 The definition's return value is the stratum value for the subject; any CQL type is allowed.
+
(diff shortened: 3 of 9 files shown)


Cql/Cql.Packaging/FhirMeasureExtensions.cs
```diff
@@ -231,11 +231,12 @@ private static FhirMeasure.StratifierComponent GetOrCreateStratifier(FhirMeasure
         if (existing != null)
             return existing;
 
+        // Only the element id is set: the FHIR invariant on Measure.group.stratifier
+        // ((code | description | criteria).exists() xor component.exists()) forbids
+        // code/description/criteria on a stratifier that holds components.
         var container = new FhirMeasure.StratifierComponent
         {
             ElementId = id,
```
baseTwo: **Required (finding 1): an empty container is now invalid, and only call-site ordering prevents it.**

Before this change a container with no components still satisfied `mea-1` — `code`/`description` existed, `component` did not, xor held. Now a componentless container has *neither* side and violates the very invariant this PR fixes.

It is unreachable today only because the single caller adds a component immediately after, and the duplicate-component `throw` sits between the two — but this helper adds the container to `group.Stratifier` before the caller can populate it, so the invariant depends on that ordering rather than on anything structural. Your comment warns the next reader off re-adding `code`; it does not tell them the container must never be left empty.

Please make it enforced rather than incidental — e.g. have the helper receive the first `ComponentComponent` (or the component id) and add it, so no code path can produce a container without one — and extend the comment to say *why* (a container with no components satisfies neither side of the xor). Add a test that pins it.

Cql/CoreTests/Packaging/FhirMeasureExtensionsTests.cs
```diff
@@ -269,6 +269,32 @@ public void DefinitionWithPopulationAndStratifierTags_ProducesBoth()
...
+                CreateTag("description", "Stratifies by age band"))));
+
+        var stratifiers = measure.Group.SelectMany(g => g.Stratifier).ToList();
+        stratifiers.Should().NotBeEmpty();
+        foreach (var stratifier in stratifiers)
+        {
+            var hasOwnContent = stratifier.Code != null || stratifier.Description != null || stratifier.Criteria != null;
+            var hasComponents = stratifier.Component.Count > 0;
+            (hasOwnContent ^ hasComponents).Should().BeTrue(
+                "stratifier {0} must have either code/description/criteria or components, never both", stratifier.ElementId);
```
baseTwo: **Required (finding 2): assert the invariant, not a C# paraphrase of it.**

`hasOwnContent`/`hasComponents` hand-translates `(code | description | criteria).exists() xor component.exists()`. If the translation is ever wrong in the same direction as a regression, this test agrees with the bug — and a hand-rolled invariant is exactly what a golden-invariant test should avoid.

The FhirPath engine ships **inside `Hl7.Fhir.Base`**, which this project already gets transitively via its `Hl7.Fhir.R4` reference, so no new `PackageReference` is needed:

```csharp
const string Mea1 = "group.stratifier.all((code | description | criteria).exists() xor component.exists())";
new FhirPathCompiler().Compile(Mea1).Predicate(measure).Should().BeTrue();
```

`Hl7.FhirPath.FhirPathCompiler.Compile(string)` and `Hl7.FhirPath.CompiledExpressionExtensions.Predicate(CompiledExpression, Base, EvaluationContext)` are both present in `Hl7.Fhir.Base` 6.3.0 (the pinned `FirelyNetVersion`) — confirm the exact overload and whether an `EvaluationContext` argument is required when you write it.

Please keep a targeted assertion alongside it (container has no `code`/`description`, components do), so a failure still identifies which half broke instead of only reporting that `mea-1` failed.

docs/releases/vnext/1499-measure-stratifier-invariant.md
```diff
@@ -0,0 +1,9 @@
+## Fixes
```
baseTwo, line 10: **Required (finding 4): this needs a `## Breaking Changes` section too.**

The root `CLAUDE.md` rule is that a change to Packager behaviour/output is a breaking change recorded in the PR that introduces it, and the versioned notes model that explicitly — `docs/releases/release-notes-2.10.0.md` has `### CQL Packager` → `#### Breaking`. Re-packaging an unchanged library now produces different `Measure` JSON than 2.11.0 did, which anyone diffing or asserting on generated resources will see.

Please add a `## Breaking Changes` entry noting that generated `Measure.group.stratifier` entries no longer carry `code` or `description`, and keep this `## Fixes` entry as it stands. `docs/releases/vnext/1461-library-set-invoker-pool.md` shows the multi-section fragment shape.

docs/cql-packager.md
```diff
@@ -216,7 +216,10 @@ ELM `annotation` element. See `LibrarySets/Demo/Cql/MeasureExample.cql` for a co
 
 All stratifier-tagged definitions of a group collapse into a single container stratifier
 (`<group>-Stratifier`) whose dimensions are `stratifier.component` entries — including when a group
-has only one dimension. Each component gets id `<group>-StratifierComponent-<value>`, a text-only
+has only one dimension. The container carries only its element id — no `code`, `description`, or
+`criteria` — because the FHIR invariant on `Measure.group.stratifier`
+(`(code | description | criteria).exists() xor component.exists()`) forbids combining them with
+components. Each component gets id `<group>-StratifierComponent-<value>`, a text-only
```
baseTwo: **Required (finding 5): say where the labels went, and what components mean for reporting.**

This is accurate, but the container is now the one element in the generated Measure with no human-readable label at all, and the docs stop at that fact. Please also state:

1. the labels live on the components (`stratifier.component.code` / `.description`), which is what a consumer should display or key on; and
2. because the dimensions are modelled as `component` entries, the stratum value in a `MeasureReport` is the **combination** of all of the group's dimension values, not one stratification per `@stratifier` tag.

Both the R4 `component` element comment ("Stratifiers are defined either as a single criteria, or as a set of component criteria") and CQF Measures IG conformance requirement 3.17 read components as a single composite stratification. The current wording — "dimensions", one container per group — leaves a reader free to assume N independent stratifications, which is the opposite of what gets reported.

Docs only; not a request to change the output shape.

baseTwo (changes requested):
## Review: FHIR stratifier invariant fix

Reviewed the full diff (+44/−5, 4 files) against the FHIR R4 spec mirror in `spec/fhir/condensed/` and the CQF Measures IG. **The premise and the fix are correct**, but there are five required changes before this should merge.

### Verified

- **The invariant is quoted exactly right.** `spec/fhir/condensed/fhir-r4-measure-definitions.md` records it as `mea-1` (Rule, defined on `Measure`): *"Stratifier SHALL be either a single criteria or a set of criteria components"* — `group.stratifier.all((code | description | criteria).exists() xor component.exists())`. The old output set `Code`, `Description` **and** `component` on the same container, so it violated `mea-1` on every generated Measure that used `@stratifier`. Real bug, correctly diagnosed.
- **Nothing of value is lost.** The removed `Code.Text` and `Description` were both just the element id (`"RateA-Stratifier"`) — no author-supplied content, nothing correlatable across measures. The per-dimension `code`/`description`/`criteria` on the components are untouched, which is where the real information lives (`Measure.group.stratifier.component.criteria` is `1..1` in R4, and it is still always set).
- **No stale artifacts.** The only generated `Measure` resources in the tree (`LibrarySets/Demo/Resources/Measure-MeasureExample-0.0.1.json` and the `410` example's `output/fhir/` copy) are untracked/gitignored build output, and `LibrarySets/RR23/TestData/RR23_QM_Umbrella_Measure.json` has no stratifiers at all — so there is nothing checked in carrying the old shape. `docs/releases/release-notes-2.10.0.md` describes the components' codes, not the container's, so it is not stale either.
- **`GetOrCreateStratifier` has exactly one call site** (`AnnotateMeasureStratifiers`), which adds a component immediately after — see finding 1 for why that now matters.
- All 8 Azure checks are green and Copilot's automated pass generated no comments.

### Required changes

**1. The change makes an empty container *invalid*, and nothing enforces that it can't happen.** Before this PR, a container with no components still satisfied `mea-1` — `code`/`description` existed, `component` did not, so the xor held. After it, a container with zero components has **neither** side of the xor and violates the same invariant this PR is fixing. Today that state is unreachable only by accident of call-site ordering: `GetOrCreateStratifier` adds the container to `group.Stratifier` *before* the caller's duplicate-component check, and it is only the `throw` in between that stops an empty container escaping. The comment you added tells the next reader not to add `code` back, but nothing tells them the container must never be left componentless. Make it structural — have the helper take the first component (or the component id) and add it, so a container cannot exist without one — and add a test that pins it.

**2. The new test asserts a C# paraphrase of the invariant, not the invariant.** `hasOwnContent`/`hasComponents` re-implements `(code | description | criteria).exists() xor component.exists()` by hand; if the paraphrase is wrong in the same direction as a future bug, the test agrees with the bug. The FhirPath engine ships **inside `Hl7.Fhir.Base`**, which `CoreTests` already gets transitively via its `Hl7.Fhir.R4` reference — no new package needed. Evaluate the literal expression instead:

```csharp
const string Mea1 = "group.stratifier.all((code | description | criteria).exists() xor component.exists())";
new FhirPathCompiler().Compile(Mea1).Predicate(measure).Should().BeTrue();
```

(`Hl7.FhirPath.FhirPathCompiler.Compile(string)` and `Hl7.FhirPath.CompiledExpressionExtensions.Predicate(CompiledExpression, Base, EvaluationContext)` are both present in `Hl7.Fhir.Base` 6.3.0 — confirm the exact overload/`EvaluationContext` argument when you write it.) Keep a targeted assertion that the container has no `code`/`description` while its components do, so a failure still says *which* half broke.

**3. No end-to-end coverage of the changed output.** The fix is only exercised through unit tests on the extension method. `LibrarySets/Demo/Cql/MeasureExample.cql` — which `docs/cql-packager.md` cites as *"a complete example"* of measure annotations, in the same section that documents stratifiers — carries `@measure`, `@year`, `@group`, `@population` and `@description` but **no `@stratifier` tag at all**, so no packaged artifact anywhere in the repo goes through this code path. Add a `@stratifier`-tagged definition to `MeasureExample.cql` and regenerate its ELM (`LibrarySets/Demo/Elm/MeasureExample.json` is checked in — use the `generate-elm-from-cql` skill; also check whether `Demo/Measures.Demo/CSharp/MeasureExample-0.0.1.g.cs` changes and commit it only if it actually does). That makes the doc reference true and gives the fix real packaging coverage.

**4. The release-note fragment needs a Breaking Changes entry.** `docs/releases/vnext/1499-measure-stratifier-invariant.md` is `## Fixes` only. Per the root `CLAUDE.md`, a change in Packager behavior/output is a breaking change and must be recorded as one in the PR that introduces it — and the versioned notes model this explicitly (`docs/releases/release-notes-2.10.0.md` has a `### CQL Packager` → `#### Breaking` section). Anyone re-packaging existing libraries gets different `Measure` JSON than 2.11.0 produced, and anyone diffing or asserting on that output will see it. Add a `## Breaking Changes` section stating that generated `Measure.group.stratifier` entries no longer carry `code`/`description`, keeping the `## Fixes` entry as-is (`1461-library-set-invoker-pool.md` shows the multi-section shape).

**5. Document how a consumer now identifies a stratification.** The docs change is accurate but stops at "the container carries only its element id". That container is now the only element in the generated Measure with no human-readable label at all, so `docs/cql-packager.md` should say where the labels live instead: on the components (`stratifier.component.code`/`.description`), and that because the dimensions are modelled as components, the stratum value in a `MeasureReport` is the **combination** of all the group's dimension values — the base spec's `component` comment ("Stratifiers are defined either as a single criteria, or as a set of component criteria") and the CQF Measures IG conformance requirement 3.17 both read components as one composite stratification, not N independent ones. A consumer reading only the current wording could reasonably expect one stratification per `@stratifier` tag.

---

@copilot please implement all five changes above on this branch. Findings 1–3 are code/test changes, 4–5 are docs; none of them are optional. Keep the existing behaviour of the fix itself — the container losing `code`/`description` is correct and should stay.

Two notes that are **not** asks on this PR: the composite-stratum modelling itself is pre-existing (introduced in #1358, released in 2.10.0), so finding 5 is a documentation ask only, not a request to restructure the output. And `DefinitionWithMultipleStratifierTags_AddsOneComponentEach` pins behaviour where one define with two `@stratifier` tags yields two components sharing one identical `criteria` expression, which reads oddly under the composite interpretation — also pre-existing, and worth its own issue rather than a change here.

Acceptance criteria on #1500: all four verified against the diff and left ticked — container is id-only, components unchanged, tests assert the invariant (finding 2 asks for a stronger form, not a missing one), and docs plus a release-note fragment are present (finding 4 asks for an added section).

Cql/Cql.Packaging/FhirMeasureExtensions.cs
alexzautke: Fixed in d22a467 — `GetOrCreateStratifier` is gone. The helper is now `AddStratifierComponent(group, component)`: it runs the duplicate-id check *first*, then creates the container (if needed) and adds the component in the same call, so no code path can put a container into `group.Stratifier` without at least one component — the duplicate `throw` now happens before the container ever exists. The comment now states the why: a componentless container satisfies neither side of the xor.

Pinned in `GeneratedStratifiers_SatisfyFhirStratifierInvariant` by the `OnlyContain(s => s.Component.Count > 0)` assertion, plus the FhirPath evaluation of mea-1 itself (an empty container fails the xor outright). Verified end-to-end: the demo's packaged `Measure-MeasureExample-0.0.1.json` (finding 3) has the id-only container with its component.

---

Cql/CoreTests/Packaging/FhirMeasureExtensionsTests.cs
alexzautke: Fixed in d22a467 — the test now compiles the literal invariant string with `new FhirPathCompiler().Compile(...)` and evaluates it via `.Predicate(measure, new EvaluationContext())`. Your API prediction held: the `Predicate(CompiledExpression, Base, EvaluationContext)` overload exists in `Hl7.Fhir.Base` 6.3.0 (verified by reflection over the pinned package; the `ITypedElement` route is `[Experimental]` and blocked by SDK0001, so the `Base` overload is also the only viable one). The targeted per-half assertions are kept alongside: container has no code/description/criteria, every container has ≥1 component, every component has code+description+criteria. All 22 tests in `FhirMeasureExtensionsTests` pass.

---

docs/releases/vnext/1499-measure-stratifier-invariant.md
alexzautke: Fixed in d22a467 — the fragment now opens with a `## Breaking Changes` section (following the `1461-library-set-invoker-pool.md` multi-section shape) stating that generated `Measure.group.stratifier` entries no longer carry `code`/`description`, why re-packagers will see different JSON, and that the labels live on `stratifier.component`. The `## Fixes` entry stands as before, extended only to note the container can no longer be emitted componentless (finding 1).

---

docs/cql-packager.md
alexzautke: Fixed in d22a467 — the Stratifiers section now says both things explicitly: the human-readable labels live on `stratifier.component.code`/`.description` (display/key on those, not the container), and a new paragraph explains that the components form a single **composite** stratification — a `MeasureReport` stratum is the combination of all the group's dimension values (one `stratum.component` per dimension), not one independent stratification per `@stratifier` tag — citing the R4 component comment and the CQF Measures IG conformance page, with links to both (and to the local spec mirror).

---

alexzautke:
@baseTwo all five findings are addressed directly on this branch in d22a467 (findings 1, 2, 4, 5 answered in their threads, now resolved). Finding 3, which had no inline thread:

- `LibrarySets/Demo/Cql/MeasureExample.cql` now has a `@stratifier`-tagged definition (`Gender Stratifier`, group 1, with a `@description`), so the doc's "complete example" claim is true and a packaged artifact exercises the stratifier path.
- Its ELM and `Demo/Measures.Demo/CSharp/MeasureExample-0.0.1.g.cs` are regenerated and committed; the other 33 `.g.cs` files differed only by the `GeneratedCodeAttribute` header and were reverted per repo convention, and the other Demo ELM files had no content changes.
- Verified end-to-end: the packaged `Measure-MeasureExample-0.0.1.json` (build output, not checked in) contains `stratifier` `1-Stratifier` with no `code`/`description`/`criteria` and one fully-labelled component — mea-1 satisfied.
- One wrinkle worth knowing: the first attempt used `Patient.gender` directly, and Library packaging failed because `CqlTypeToFhirTypeMapper` can't map the `AdministrativeGender` enum type to a FHIR parameter type (`ElmDefinitionToParameter` throws). The committed definition uses `Patient.gender.value` (a plain string). That mapper gap may deserve its own issue — along with the double-`@stratifier`-tag oddity you flagged as pre-existing.

Also fixed en route: `CqlToolingEnabled` ELM regeneration was broken on Linux/macOS — the unquoted Java classpath wildcard in `Demo/Cql/Build/CqlTooling.Targets.xml` was shell-expanded so the first dependency jar was misread as the main class. It's now quoted (noted in the release fragment).

Since the changes are on the branch itself, #1501 is redundant and can be closed.

---

baseTwo (approved):
## Approving — all five findings addressed, verified at `d22a467b0`

Re-reviewed the branch tip and confirmed every one of the five required changes landed, in each case in a stronger form than I asked for. All 8 Azure check runs report `success` against `d22a467b0` specifically (not an earlier commit), and all four of my inline threads are resolved.

**1. Empty container — fixed structurally.** `GetOrCreateStratifier` is gone, replaced by `AddStratifierComponent(group, component)`: the container id is derived in one place, the duplicate check runs before any mutation, and `container.Component.Add(component)` is unconditional at the end, so a container cannot be added to a group without a component. The empty state is now unrepresentable rather than merely unreached, which is exactly the point. `git grep` confirms no leftover references to the old name.

**2. Invariant asserted literally.** The `mea-1` expression is a documented `const` evaluated through `FhirPathCompiler`/`Predicate` against the generated `Measure`, with three targeted assertions kept for diagnosability — including one nothing else in the file made: that every *component* carries `code`, `description` and `criteria`.

**3. End-to-end coverage.** `MeasureExample.cql` gains a `Gender Stratifier` returning `Patient.gender.value` — the unwrapped `String` stratum value, which is the right thing to demonstrate given the docs define the return value as the stratum value. ELM and `.g.cs` regenerated; the ELM locators (`58:1-59:21`, `59:2-59:22`) line up exactly with the new CQL line numbers, and the `.g.cs` content change is real, so committing it with its refreshed `GeneratedCode` header is correct per `CLAUDE.md`. No `GeneratorToolVersion` bump is needed — nothing changed about *how* C# is emitted.

**4 and 5.** The fragment now carries a `## Breaking Changes` entry describing the output shape change and where labels moved to, distinct from the `## Fixes` entry describing the bug. The docs say to key on `stratifier.component.code`/`.description` and explain the composite-stratum consequence, with real Markdown links to both spec pages per the repo's docs convention.

**Bonus fix, correctly scoped.** Regenerating the ELM surfaced a genuine cross-platform bug — the Java classpath wildcard in `Demo/Cql/Build/CqlTooling.Targets.xml` was unquoted, so Unix shells expanded it and `java` misread the first dependency jar as the main class. Quoting it is safe on Windows and satisfies the `CLAUDE.md` rule that build scripts work on both, and it is recorded under `## Fixes` rather than slipped in silently.

Acceptance criteria on #1500: all four verified against the final diff and correctly ticked. #1501, @copilot's parallel implementation of this same review, is closed as superseded.

Issue #1506, opened 2026-08-05
FHIR dateTime values converted from CQL lack the required timezone offset


alexzautke:
## Summary

[FHIR-55977](https://jira.hl7.org/browse/FHIR-55977) (the CQL-to-FHIR type-mapping guidance for partial-precision `System.Time`/`System.DateTime` values, implemented here via #1456 / PR #1458) describes the resulting FHIR value **without a timezone offset when the CQL value has none**. That description is wrong on the HL7 ticket itself: the FHIR [`dateTime`](https://hl7.org/fhir/R4/datatypes.html#dateTime) datatype requires that **"if hours and minutes are specified, a timezone offset SHALL be populated"** — the lexical grammar only admits `Z` or `±hh:mm` once a time component is present. Our converter now faithfully reproduces the ticket's mistake, so every `dateTime` we emit from an offset-less CQL value is lexically invalid FHIR.

## Impact

Resources built from converted CQL values (e.g. `Period` boundaries in evaluated results, or any `FhirDateTime` produced by `FhirTypeConverter`) fail FHIR validation and can be rejected by servers. This affects both the new padded partial-precision output of PR #1458 and the second-or-finer precision path that has been on `develop` since PR #1408.

## Observed Behavior

All in `Cql/Cql.Firely/FhirTypeConverter.cs` (as of PR #1458's branch, `claude/cql-time-precision-extension-txyir9`):

1. `CqlDateTimeToFhirDateTime` — a `CqlDateTime` with a time component but no offset produces an offset-less `dateTime`:
   - padded path: `@2014-02-01T10` → `"2014-02-01T10:00:00"` (`FormatPaddedTime` renders `""` for a null offset);
   - second-or-finer path (pre-existing on `develop`): `@2014-02-01T10:30:00` → `"2014-02-01T10:30:00"`.
2. `CqlInterval<CqlDateTime>` → `M.Period` boundaries go through the same helper — same invalid output.
3. `CqlTimeToFhirDateTime` (`CqlInterval<CqlTime>` → `M.Period`) — a `CqlTime` without an offset yields e.g. `"0001-01-01T10:30:00"`, with no offset. It additionally rewrites a zero offset's `Z` into `+00:00`, which is valid but inconsistent with `FormatPaddedTime` (which renders a zero offset as `Z`), so the same zero-offset value renders differently depending on the conversion path.

[...]

Issue #1432, opened 2026-07-29
Split vnext-release-notes.md into per-PR fragment files to avoid merge conflicts


baseTwo:
## Summary

`docs/releases/vnext-release-notes.md` is a single shared file that every PR introducing a breaking change, feature, or fix is required to append a bullet to (per `CLAUDE.md` §"Release notes" and copilot-instructions §4.5.1). With several PRs landing in parallel (including stacked chains), this causes recurring merge conflicts on that one file — we've hit it twice in a single review pass this week (PR #1421 vs. `develop`, and again against its own stacked base #1414).

## Proposed Design

Replace the single shared file with **one fragment file per PR**, consolidated into a versioned release note only when a release is cut.

### 1. New fragment directory: `docs/releases/vnext/`

- Each PR that needs a release-note entry adds **one new file** here instead of editing a shared file — so two PRs developed in parallel practically never touch the same file.
- **Filename**: `<issue-number>-<short-slug>.md` (e.g. `1413-expand-overshoot.md`), using the tracking **issue** number rather than the PR number, since the issue number is already known when the branch is created (this repo's branch-naming convention already embeds it, e.g. `claude/fix-expand-overshoot-1413`) whereas the PR number isn't assigned until the PR is opened. A PR with no tracking issue falls back to a slug-only name unlikely to collide (e.g. `add-benchmark-skill.md`).
- **Content**: exactly the same structure as today's `vnext-release-notes.md`, just scoped to only the section(s) this PR needs:

  ```md
  ## Fixes

  - `Expand` no longer emits a trailing interval that overshoots the upper
    boundary when `per` does not divide the interval width. ... (#1413)
  ```

  A PR that has both a breaking change and a fix writes both `##` sections in its one fragment file. No new metadata format (frontmatter, filename-encoded category, etc.) to invent or parse incorrectly — it's the same markdown contributors already write today, just split per-file.
- `docs/releases/vnext/README.md` documents the convention and is explicitly excluded from consolidation by name (not treated as a fragment).

### 2. `docs/releases/vnext-release-notes.md` becomes a static pointer (end state, see Transitional Phase below)

Once the transition is complete, replaced with a short doc: "Pending release notes are tracked as individual files under `docs/releases/vnext/`, one per PR. See that folder for everything accumulated since the last release." This avoids breaking existing hyperlinks into this path from `CLAUDE.md`, copilot-instructions, and the `cut-release-notes` skill.

[...]

PR #1509, opened 2026-08-06
Validate emitted FHIR in tests, not just its text

### LINKED ISSUE #1508: Tests pin emitted FHIR by string only, so invalid output passes
## Summary

Tests covering the CQL→FHIR converter assert the emitted value with an exact-string comparison and nothing else. A string assertion pins whatever the converter currently produces, so it cannot distinguish a correct value from a lexically invalid one — it only detects *change*. As a result, invalid FHIR reached `develop` with a full green suite, and in one case a test was written that asserted the invalid output was correct.

## Impact

Two concrete instances, both found by review rather than by the test suite (see #1506 and PR #1458):

1. **`FHIR.time` carrying a timezone offset.** `Convert<Time>` emitted `"10:30:00Z"` and `"10:30:00+02:00"`. FHIR R4 `time` states *"A time zone SHALL NOT be present"*; `Hl7.Fhir.Model.Time.IsValidValue` returns `false` for both. A test — `ConvertCqlTime_Time_PassesAVestigialOffsetThrough` — pinned exactly these two strings, so the suite actively defended the invalid output.
2. **Time-bearing `dateTime` with no offset.** `"2014-02-01T10:30:00"` was emitted for an offset-less CQL value. FHIR R4 requires an offset once hours and minutes are present; `FhirDateTime.IsValidValue` returns `false`. Several tests pinned the offset-less strings.

Both were lexically invalid FHIR that a receiving server can reject, and in both cases the suite was green.

## Root Cause

`Hl7.Fhir.Model.<Type>.IsValidValue` is a public static on `Hl7.Fhir.Base`, which the test projects already reference, but it is not used anywhere in the test suite. Nothing in `CLAUDE.md` or `.github/copilot-instructions/` asks for it, so each new converter test reproduces the pattern: assert the string, assume validity.

A string pin is a *regression* oracle. It answers "did the output change?", never "is the output correct?" — and when the first-written expectation is wrong, it converts a bug into a defended invariant.

## Expected Behavior

Every test pinning the text of an emitted FHIR primitive also asserts `IsValidValue` on that text, so a malformed value fails on the validity assertion regardless of what the string assertion says. Documented as a convention in both instruction files, since Copilot and Claude both author converter tests in this repo.

Separately: "pre-existing" and "out of scope" are legitimate reasons not to *fix* invalid output in a given change, but never reasons to pin it as correct or to document an invariant the code does not hold. Instance 1 above was initially triaged that way before being corrected.

## Acceptance Criteria

- [ ] `CLAUDE.md` states that a test pinning emitted FHIR primitive text must also assert `Hl7.Fhir.Model.<Type>.IsValidValue`.
- [ ] The equivalent rule exists in `.github/copilot-instructions/`, per the sync requirement that a universal convention not live in only one of the two.
- [ ] Both state that invalid output may be left unfixed but must never be pinned as correct, nor described by a comment asserting an invariant the code does not hold.
- [ ] The existing converter tests already carrying `IsValidValue` assertions (added in PR #1458) match the documented convention.

alexzautke:
## Description

**Primary Work**

Records a testing convention that two invalid-FHIR defects on #1458 showed was missing: **a test pinning the text of an emitted FHIR primitive must also assert `Hl7.Fhir.Model.<Type>.IsValidValue`.**

A string assertion is a regression oracle — it answers "did the output change?", never "is the output correct?". When the first-written expectation is itself wrong, the assertion converts a bug into a defended invariant. That is what happened twice:

- `Convert<Time>` emitted `"10:30:00Z"` and `"10:30:00+02:00"`, which FHIR R4 forbids (`time`: *"A time zone SHALL NOT be present"*). A test pinned exactly those two strings, so the suite defended the invalid output.
- Time-bearing `dateTime` values were emitted with no offset (`"2014-02-01T10:30:00"`), which FHIR requires once hours and minutes are present. Several tests pinned the offset-less strings.

`IsValidValue` returns `false` for all of these and is a public static on `Hl7.Fhir.Base`, which the test projects already reference — one line per emission point. It was not used anywhere in the suite before #1458.

The same rule carries a second sentence: *"pre-existing"* and *"out of scope"* justify not **fixing** invalid output in the change at hand, but never pinning it as correct or documenting an invariant the code does not hold. The first defect above was initially triaged as out-of-scope pass-through and given a comment defending it, before being corrected.

Added to `CLAUDE.md` and mirrored in `.github/copilot-instructions/08-cql-specification-conformance.md` (new §8.4, with the TOC entry), per this repo's rule that a universal convention must not live in only one of the two instruction sets. Both agents author converter tests here, so the Copilot side is load-bearing rather than a formality.

---

**Auxiliary Work**

None.

## Related issues

Closes #1508.

Follows from #1506 and #1458, where both defects were found.

## Testing

Documentation only — no code, no test changes, no public API surface. CI skips the build for this PR because every changed path matches an `ignorePatterns` entry (`*.md`), which is the intended behaviour for a documentation-only change.

The convention the rule describes is already satisfied in the code base: the `IsValidValue` assertions it mandates were added to `FhirTypeConverterTests` and `FhirTypeConverterHostConversionTests` in #1458, which is green.

## FirelyTeam Checklist


baseTwo (changes requested):
## Review of `d6c8a6f9b` — changes requested

**Note on staleness:** this branch is still **52 commits behind `develop`** (the others in this batch were updated). Per repo convention the base merge is your call, not mine — flagging it so it is a decision. CI green here predates most of what has landed.

The documented rule is sound and the assertions it describes genuinely work — I proved they fail on invalid FHIR — but AC #4 is false: I reintroduced the issue's own defect #1 and `ConvertCqlTime_Time_DropsVestigialOffset` (the direct successor of the test the issue names) passed green while emitting `"10:30:00Z"` and `"10:30:00+02:00"`, because it and 13 other emission tests carry no validity assertion at all.

### Verified

- The rule the PR documents is correct and the mechanism works: `FhirDateTime.IsValidValue("2014-02-01T10:30:00")` and `Time.IsValidValue("10:00:00Z")` both return false, and the existing helper assertions surface that as a test failure — proven by breaking the converter and watching the validity assertion (not the string assertion) fire, at FhirTypeConverterTests.cs:606 and :612.
- The diff is exactly 2 files, +21/-0, both markdown: CLAUDE.md and .github/copilot-instructions/08-cql-specification-conformance.md. No code, no generated files, so no GeneratorToolVersion bump, no LibraryInstanceInvoker range change, no ReflectionUtility / #nullable enable / copyright-header / InternalsVisibleTo concerns apply.
- AC #2 is met: the equivalent rule exists in .github/copilot-instructions/08-cql-specification-conformance.md as a new §8.4 with a matching TOC entry, satisfying the CLAUDE.md sync requirement that a universal convention not live in only one of the two instruction sets.
- AC #3 is met on both sides: CLAUDE.md's third paragraph and copilot §8.4.5 both state that 'pre-existing'/'out of scope' justify not fixing invalid output but never pinning it as correct or documenting an invariant the code does not hold.
- The §8.4 numbering is now consecutive (8.4.1-8.4.5) after commit d6c8a6f9b closed the gap left when commit 7863c7a9b removed the old 8.4.5; every other section in that document numbers consecutively, so this matches.
- CI legitimately skipped the build: both changed paths match the `^.*\.md$` ignorePatterns entry at build/azure-pipelines.yml:35, and both check-runs report success. The PR description's statement about this is accurate.
- No docs/releases/vnext fragment is present, and none is required — the folder holds only 1467-shared-bundle-index.md and README.md, and an AI-instruction-file convention is not a public API, generated C# output, Packager CLI, or build/tooling behaviour change.
- The branch merges into develop cleanly despite being 52 behind — verified by an actual `git merge --no-commit --no-ff origin/develop` trial in my worktree, which auto-merged CLAUDE.md with no unmerged paths.

```
I could build and run everything. Because this PR is documentation-only and its branch predates #1458 (the branch head has zero `IsValidValue` occurrences in any `.cs` file), I ran all code experiments against `origin/develop`, checked out onto a throwaway branch inside my own worktree, then restored to the PR head (worktree verified clean, `git status --short` empty, HEAD at d6c8a6f9b).

BASELINE: `dotnet build Cql/CoreTests/CoreTests.csproj -c Release` -> 0 Errors, 412 warnings (pre-existing nullable noise, not reported). `dotnet test ... -f net10.0 --filter "FullyQualifiedName~FhirTypeConverter"` -> "Passed! - Failed: 0, Passed: 143, Skipped: 8, Total: 151".

EXPERIMENT A — does the dateTime validity assertion actually fire? I corrupted `Cql/Cql.Firely/FhirTypeConverter.cs` (`private const string Utc = "Z"` -> `""`), reproducing the issue's defect #2, AND updated the expected string in `ConvertCqlDateTime_MinutePrecisionWithoutOffset_...` from `"2014-02-01T10:30:00Z"` to `"2014-02-01T10:30:00"` so the string assertion could no longer fire and only the validity assertion could. Result: FAILED with `Assert.IsTrue failed. '2014-02-01T10:30:00' is not a valid FHIR dateTime value` at FhirTypeConverterTests.cs:612 (`AssertIsValidFhirDateTime`). The guard is real and load-bearing.

EXPERIMENT B — same for `time`. I corrupted `CqlTimeToFhirTime` to pass `offsetWhenAbsent: Utc` and updated the expected string in `ConvertCqlTime_HourPrecision_...` to `"10:00:00Z"`. Result: FAILED with `Assert.IsTrue failed. '10:00:00Z' is not a valid FHIR time value` at FhirTypeConverterTests.cs:606. Confirms `Time.IsValidValue` rejects a timezone-bearing time exactly as the issue claims.

EXPER
```
</details>

**Acceptance criteria:** All four checkboxes on issue #1508 are currently unticked. My assessment:

- AC #1 — "`CLAUDE.md` states that a test pinning emitted FHIR primitive text must also assert `Hl7.Fhir.Model.<Type>.IsValidValue`, with the two real failures as motivation." — PARTIALLY MET. The rule is stated clearly and well. The motivation clause is NOT met: neither failure is named and there is no link to #1508/#1506/#1458 in either file. Commit 7863c7a9b removed the only text that carried them. Needs either a restored sentence or an amended AC (see findings).

- AC #2 — "The equivalent rule exists in `.github/copilot-instructions/`, per the sync requirement." — MET. New §8.4 in 08-cql-specification-conformance.md with a matching TOC entry, content equivalent to the CLAUDE.md section. (Placement and routing ar

**Release note:** No fragment under `docs/releases/vnext/` — and correctly so. That folder currently contains only `1467-shared-bundle-index.md` and `README.md`. CLAUDE.md scopes fragments to breaking changes in public API, generated C# output, Packager CLI behaviour, or build/tooling behaviour. This PR changes only AI-instruction markdown (`CLAUDE.md` and `.github/copilot-instructions/`), which is none of those and has no consumer-visible effect. The PR description's "Documentation only — no code, no test change

### Required — @alexzautke please fix these

#### 1. AC #1 not met — the two real failures were deleted as motivation, leaving the rule with no example and no issue link anywhere — `CLAUDE.md:59`

AC #1 requires CLAUDE.md to state the rule "with the two real failures as motivation". The new `## Validating emitted output` section never mentions them: no `Time` carrying an offset, no offset-less time-bearing `dateTime`, and no link to #1508, #1506 or #1458. The Copilot side did carry them in the original 8.4.5, but commit 7863c7a9b ("Remove invalid values example from CQL specification", @alexzautke) deleted that item outright — which also removed the only issue link in either file.

That deletion appears to be a reaction to Copilot's two review comments, both of which only asked to repoint the link from #1506 to #1508; deleting the line satisfied neither the comments nor the AC. Note the irony that surviving §8.4.5 instructs the reader to "quarantine it with a comment pointing at the tracking issue" while the section itself points at no issue.

This needs a human decision because the removal was a deliberate maintainer edit. Either (a) restore one sentence to CLAUDE.md and to copilot §8.4 naming the two failures with a `[#1508](https://github.com/FirelyTeam/firely-cql-sdk/issues/1508)` link, or (b) amend AC #1 on the issue to drop the motivation requirement. Option (a) also brings the section in line with CLAUDE.md's own house style — neighbouring sections do link out (e.g. the release-notes section links `[#1432]`, the errors section links `[Cql/Cql.Abstractions/CLAUDE.md]`).

*Evidence:* `git diff origin/develop...HEAD -- CLAUDE.md` shows the added section contains zero Markdown links and no mention of either failure. `git show 7863c7a9b` shows the removed line was the only one naming `Time("10:30:00Z")`, the offset-less dateTime, and issue #1508. `gh api .../pulls/1509/comments` shows both Copilot comments asked to change the link target to #1508, not to remove the line.

1. **A FHIR output-testing convention is filed under "CQL Specification Conformance"** — Section 8 is titled "CQL Specification Conformance" and 8.1-8.3 are entirely about the local CQL spec mirror at `/spec/cql/condensed/`. The new §8.4 is about asserting FHIR primitive validity in tests — a testing convention, not CQL spec conformance. `05-build-and-test.md` is the closer home; alternatively `04-development-guidelines.md`. If §8 is kept, retitle it to something like "Specification Conformance and Outpu

### Notes — no action needed

- **The FHIR datatypes page the new rule relies on is not in the local spec mirror** — §8.4.4 and the CLAUDE.md text make lexical claims about FHIR `time` and `dateTime` ("A time zone SHALL NOT be present"; offset required once hours and minutes appear), but `spec/fhir/condensed/` holds only 4 pages — library definitions, measure definitions, CQL conformance IG, and CQF measures IG — none of them the datatypes page. §8.1 of this very document tells readers to con
- **Copilot's review comment on .claude/settings.json is stale and should not block** — The bot review flags a hard-coded `/home/user/firely-cql-sdk` path in `.claude/settings.json`, but that file is not part of this PR — the diff against the merge-base is exactly two markdown files. The comment is left over from an earlier state of the branch. Dismiss it rather than acting on it here; if the settings.json path issue is real it belongs in its own ticket.

### Considered and dropped

- **AC #4 is false: 14 emission tests pin FHIR primitive text with no validity assertion, and the issue's own defect #1 still passes green** — All three load-bearing claims fail, though the raw observation (14 emission tests on develop lack IsValidValue) is factually accurate.

(1) AC #4 is misread. Its text is "The existing converter tests **already carrying** IsValidValue assertions (added in PR #1458) match the documented convention." T
- **The new rule is unreachable through the routing index that decides which instruction doc gets read** — The facts are right but the inference fails on four counts. (1) The evidence omits half the diff: it claims the PR "touches only the doc's own internal TOC", but `git diff origin/develop...HEAD --stat` shows `CLAUDE.md | 8 ++++++++` — the PR adds the rule as a new top-level `## Validating emitted ou

baseTwo:
@alexzautke — `f6996a402` is a `develop` merge only; the branch is now **0 behind**, which is useful, but the PR's
own diff is unchanged at 2 files / +21 lines and the outstanding item is untouched.

To restate it precisely so it is actionable without re-reading the thread — **AC #1 on
[#1508](https://github.com/FirelyTeam/firely-cql-sdk/issues/1508)** asks that the rule be stated *"with the two real
failures as motivation"*. Neither instruction file mentions either failure, and neither carries an issue link:

```
grep -iE "#1506|#1508|#1458|10:30|issues/" <the added sections>   ->  no matches
```

The two failures the rule exists because of are:

- `Time("10:30:00Z")` — a FHIR `time` carrying a timezone offset, which `Time.IsValidValue` rejects
- an offset-less time-bearing `dateTime`, which `FhirDateTime.IsValidValue` rejects

Both were verified by reflection over `Hl7.Fhir.Base` 6.3.0 during the #1458 review, and both are exactly the shape
§8.4.4 now tells the reader to guard with `IsValidValue`.

They were in the original §8.4.5 and were removed by `7863c7a9b` ("Remove invalid values example from CQL
specification"), which also removed the only issue link in either file. That commit appears to have been a response
to two Copilot comments that asked only to **repoint the link** from #1506 to #1508 — deleting the line satisfied
neither the comments nor the AC.

Worth noting the irony that the surviving §8.4.5 tells the reader to "quarantine it with a comment pointing at the
tracking issue" while the section itself points at no issue.

**What I am asking for:** restore one sentence of motivation naming both failures, with a link to
[#1508](https://github.com/FirelyTeam/firely-cql-sdk/issues/1508). Something like:

> This rule exists because two emission tests pinned invalid FHIR as correct: a `time` carrying an offset
> (`"10:30:00Z"`) and a time-bearing `dateTime` with none, both rejected by their own `IsValidValue`
> ([#1508](https://github.com/FirelyTeam/firely-cql-sdk/issues/1508)).

In `CLAUDE.md`'s `## Validating emitted output` and in `08-cql-specification-conformance.md` §8.4, so the two
instruction systems stay in sync per the repo's own rule.

That is the only thing outstanding. Everything else about this change I verified and it stands — the rule is sound,
the `IsValidValue` mechanism is correct, and I confirmed the assertions it describes genuinely fail on invalid FHIR.

baseTwo (approved):
## Approving `f6996a402` — all four acceptance criteria met

@alexzautke — my outstanding item was AC #1's *"with the two real failures as motivation"*. That criterion has been
rewritten, so the request is moot and I am withdrawing it. Re-verified all four against the branch:

1. **`CLAUDE.md` states the rule** — `## Validating emitted output` (lines 64-70) requires a test pinning emitted
   FHIR primitive text to also assert `Hl7.Fhir.Model.<Type>.IsValidValue`.
2. **The Copilot side carries the equivalent** — §8.4.1-8.4.5 in `08-cql-specification-conformance.md`, so the
   convention lives in both instruction systems per the sync rule rather than only one.
3. **Both state the never-pin rule** — `CLAUDE.md` line 70 and §8.4.5 both say "pre-existing" and "out of scope"
   justify not *fixing* invalid output, never pinning it as correct, and both direct the reader to quarantine with
   a comment pointing at the tracking issue.
4. **The #1458 converter tests match the convention** — 7 `IsValidValue` assertions across
   `FhirTypeConverterHostConversionTests` and `FhirTypeConverterTests`, including the `FhirDateTime` start/end pairs
   and the `Time` check.

Branch is 0 behind `develop`; checks green.

Standing behind the original review: the rule is sound, and the assertions it prescribes genuinely work — I
confirmed they fail on invalid FHIR rather than passing regardless, which is the whole point of the convention.

One note, no action: §8.4 sits under a section titled "CQL Specification Conformance" while 8.1-8.3 are about the
local CQL spec mirror, and this is a FHIR output-testing convention. Filing is a judgement call and not worth
another round — worth remembering if that section is ever reorganised.

Issue #1483, opened 2026-08-01
Decide: drop FhirTypeConverter's LRUCache<CqlDateTime>? Measured: −5.4% allocation, no CPU effect


alexzautke:
## Summary

`FhirTypeConverter` keeps a process-wide `static LRUCache(10_000)` mapping FHIR date strings to `CqlDateTime` wrappers, plus supporting machinery (`LRUCacheCache`, a `cacheSize` knob surfaced through `FhirCqlContext`/`ElmToolkitConfig`, reflection-based test helpers). It is value-keyed global mutable state shared across patients and evaluations — the kind of cache worth removing *if* it no longer pays for itself.

The cache's justification was avoiding expensive `CqlDateTime` construction. #1478 (PR #1482) makes construction cheap by formatting ISO-8601 strings lazily (1232 ns / 2864 B → 545 ns / 792 B per construction). This ticket carries the post-#1478 measurement of what the cache still buys, so the keep/drop decision can be made on numbers.

## Measurement

**Setup:** commit `abc2866` (head of `1478-lazy-iso8601-formatting`), in-process harness running 900 previously-passing CMS test cases per rep. Two builds differing only in `Hl7.Cql.Fhir.dll` (verified by hash; `Hl7.Cql.Iso8601.dll` identical): **A** = cache active (default 10,000 entries, confirmed live on the `FhirCqlContext.Create` path), **B** = cache disabled via a one-line probe (`dateTimesCache = null`). 24 reps per variant, interleaved and order-balanced over 6 rounds (`A B / B A / …`) to cancel machine drift; no drift observed over the session. 4-core machine, workstation GC, sequential (dop=1).

**Results (B − A, i.e. the cost of running without the cache):**

| Metric | Paired mean diff | 95% CI (n=24 pairs) | Pairs positive | Verdict |
|---|---|---|---|---|
| Allocation | **+39.5 MB (+5.4%)** | [+38.4, +40.6] | 24/24 | Real, tight |
| CPU | −0.17% | ±~3% (straddles zero) | 11/24 | No measurable effect |
| Wall | −0.64% | straddles zero | 7/24 | No measurable effect |

[...]

Issue #1470, opened 2026-08-01
TypeConverter re-resolves every conversion, and CanConvert formats a diagnostics string per call


alexzautke:
## Summary

`TypeConverter` does two pieces of per-call work whose result never varies for a given type pair:

1. **`Convert` re-resolves which converter applies, every time.** It scans `_customConverters` with `SingleOrDefault`
   before consulting the `_converters` dictionary. The lambda captures, so each call also allocates a closure.
2. **`CanConvert` builds a diagnostics string on every call** — `TypesToString((from, to))` formats both type names —
   and adds it to a `HashSet<string>`. That set feeds `LogFinalConverters`, which only reports anything when a
   logger is attached. At evaluation time there is none, so the string is built and discarded.

Conversions run per value during an evaluation, so both are paid per element of a query.

## Impact

The clearest case is a coded element compared against a string literal — `Encounter.status = 'finished'` — which
the code generator emits as a conversion followed by a string comparison:

```csharp
Code<EventStatus> ar_ = ThrombolyticProcedure?.StatusElement;
EventStatus? as_ = ar_?.Value;
string at_ = context.Operators.Convert<string>(as_);   // <- the expensive step
bool? au_ = context.Operators.Equal(at_, "completed"); // ~40 ns, not the problem
```

Measured over 200 elements on .NET 10:

| | Time | Allocated |
| --- | --- | --- |
| Convert + Equal, per element | 2.60 µs | 856 B |
| of which `Equal(string, string)` | ~0.04 µs | 0 B |

So essentially all of it is the conversion, and nearly all of the allocation with it.

The allocation comes from `EnumToStringTypeConverterEntry.Handles`, the third of the three registered custom
converters, which ends in `nonNullableType.GetCustomAttribute<FhirEnumerationAttribute>()`. `GetCustomAttribute`
constructs a fresh attribute instance on every call, and `SingleOrDefault` cannot stop at the first match — it has
to keep scanning to detect a second one — so that converter is always reached.

This is not specific to coded elements. The same scan sits in front of **every** conversion the engine performs:
`FhirDateTime` → `CqlDateTime`, `Quantity` → `CqlQuantity`, `Period` → `CqlInterval`, and the rest.

## Observed Behavior

`TypeConverter.Convert(object?, Type)`, per call:

```csharp
if (_customConverters.SingleOrDefault(converter => converter.Handles(fromType, to)) is {} subConverter)
    return subConverter.Convert(from, to);
```

[...]

PR #1525, opened 2026-08-08
Remove FhirTypeConverter's CqlDateTime LRU cache

### LINKED ISSUE #1487: Remove FhirTypeConverter's LRUCache&lt;CqlDateTime&gt; (implements the #1483 decision)
## Summary

Implements the decision on #1483: **drop** the process-wide `LRUCache<CqlDateTime>` from `FhirTypeConverter`. The measured basis (see #1483 for full protocol): with lazy ISO-8601 formatting (#1482) in place, the cache buys ~39.5 MB allocation per 900 cases and **no measurable CPU or wall benefit**; the fixture-free trace additionally shows the cache-consult path itself (`FhirDateTimeToCqlDateTimeViaCaching`) at **8.6% of active CPU (~57 µs/bundle)** — string hashing and lookup on every conversion. Removal trades a modest, gen0-only allocation increase for less CPU on the conversion path, elimination of process-wide mutable state keyed by patient data, and the deletion of a tuning knob.

## Mechanics

- Delete `Cql/Cql.Firely/LRUCache.cs`, `DefaultDateTimesCache`, `LRUCacheCache`, and the cache wiring in `FhirTypeConverter.CreateImpl`/`ConvertFhirToCqlPrimitives`.
- **Remove the `cacheSize` parameter surface outright** from `FhirTypeConverter.Create` and the `ElmToolkitConfig` knob.
  _(Amended 2026-08-10. This originally read "keep the surface but mark it `[Obsolete]`(accepted-and-ignored) — non-breaking now, removal at the next major". The decision was taken to remove now: an accepted-and-ignored obsolete parameter is a maintenance burden and misleads callers, who keep passing a cache size while nothing caches. This is a **breaking change** — three of the removed members are in the released public API, so consumer code stops compiling on a minor upgrade. It must be recorded in the fragment with a migration table.)_
- Update the reflection-based test helper (`CoreTests/Shared.cs` `GetCapacity`), `FhirCqlContextCreateTests`, and `RecreateCqlContextBenchmarks` — remove or repoint whatever asserts cache behavior.
- `ConditionalWeakTable` replacement is explicitly **not** part of this ticket (it remains #1483's documented fallback if a future GC-pause measurement on server-GC/high-dop hosts shows the allocation matters).

## Acceptance Criteria

- [x] No `LRUCache` type or usage remains in the runtime path; the `cacheSize`/`ElmToolkitConfig` knob is **removed** (amended 2026-08-10 from "obsoleted").
- [x] Behavior locks: FhirDateTime→CqlDateTime conversion results identical (including partial precisions and offsets), passing against the current implementation too.
- [x] Corpus A/B (steady-state, fixture-free): CPU and allocation deltas reported; the expectation from #1483's data is CPU flat-or-better and allocation up by roughly the cache's measured value — the fragment states what was measured, both directions.
- [x] Full CoreTests green; **public API removals are expected and documented** (amended 2026-08-10 from "obsoletions only") — each removed member tabled with its replacement, and the PR carries the `breaking change` label; no `GeneratorToolVersion` impact.
- [x] Release fragment records the removal and the measured trade with the #1483 reference, and carries a `## Potentially Breaking` section tabling every removed member against its replacement.

alexzautke:
## Description
Implements #1487 (maintainer-approved): `FhirTypeConverter`'s process-wide `LRUCache<CqlDateTime>` is removed entirely, together with every knob that fed it — per the follow-up direction, **removed outright, no deprecation**:

- `Cql/Cql.Firely/LRUCache.cs` deleted (no other users repo-wide); `DefaultDateTimesCache`, `DefaultWithCache`, `LRUCacheCache` and the `FhirDateTimeToCqlDateTimeViaCaching` path deleted. Conversions take the path the cached code already used for extension-adorned values, so **results are identical**.
- **Public API removals** (recorded as `*REMOVED*` in the PublicAPI ledgers; `breaking change` label applies): `FhirTypeConverter.Create(ModelInspector, int? cacheSize = null)` → `Create(ModelInspector)`; the `(ModelInspector, int?, TimeSpan?)` overload → `Create(ModelInspector, TimeSpan?)`; `FhirCqlContextOptions.OverrideFhirTypeConverterCacheSize` and `ElmToolkitConfig.LRUCacheSize` gone. Migration is dropping the argument/initializer. Splitting `Create` into two overloads keeps `Create(model)` source-compatible and retires the `RS0027` suppression the old shape needed.
- Converter-instance memoization survives, keyed by **(model, default timezone offset)** only — two callers that previously differed solely in cache size now share a converter. Default model + no offset still short-circuits to the `Default` singleton.
- Nothing depends on the cache's only observable effect (reference identity of returned `CqlDateTime`s): `Equals`/`GetHashCode` are value-based, and the `ReferenceEquals` uses in the comparer bridges are fast paths whose fall-through yields the same answer.
- `ConditionalWeakTable` replacement is explicitly out of scope (#1483's documented fallback).
- No `GeneratorToolVersion` impact: `ICqlOperators`, `_CODE GENERATOR VERSION_.cs` and all `*.g.cs` untouched.
- `docs/caching-strategies.md` §4 rewritten, runtime-services dependency diagram regenerated via the export script.

> **Merge-order note:** the removal's performance rationale assumes #1482 (lazy ISO 8601 formatting, open PR) — the A/B below was measured with #1482 in both arms. The code itself is independent and merges cleanly in either order.

## Related issues
Closes #1487. Implements the decision on #1483.

## Testing
- **Behavior locks:** new `FhirDateTimeConversionTests` (34 cases: year→millisecond precisions × Z/positive/negative offsets, the four time-precision-extension codes, `.DateOnly` projection, repeated-conversion equality, null value), pulled through a default `FhirCqlContext` so on develop they exercise the live 10,000-entry cache and here the uncached path — **34/34 on both** (verified by restoring develop's implementation, rebuilding, re-running).
- **Corpus A/B for this exact change** (2,851 cases, 62 measures, #1482 in both arms, cache presence the only delta, 48 order-balanced runs incl. mirrored sequences):
  - **Equivalence:** per-case oracle fingerprints **byte-identical** across the corpus (`ok 2851 fail 0` both arms).
  - **Allocation +83.1 MB/run (+11.3 %, +29 KB/case)** — unanimous across all 24 rounds, and **entirely gen0** (+5 gen0 collections, gen1/gen2 unchanged at zero, +5 ms GC pause on a ~1.9 s run). Direction as #1483 predicted; magnitude ~⅔ of the naive per-case extrapolation.
  - **CPU/wall lean 1.5–6 % slower without the cache** — direction consistent across order-balanced and mirrored sequences, magnitude noise-sensitive (19/24 rounds positive). ⚠️ **This does not reproduce #1483's "no measurable CPU benefit"** on this corpus: construction is measurably more expensive than the string lookup even with lazy formatting. The release fragment states this plainly — the removal stands on eliminating process-wide mutable state keyed by patient data, deleting an un-tunable knob, and the allocation being strictly gen0, **not** on a CPU win. If the CPU lean changes the decision, say so and I'll close this in favor of keeping the cache.
- **Suites:** full CoreTests net10.0 Release **1198 passed / 8 skipped / 0 failed** (baseline 1167: +34 locks, −3 cache-observation tests); `Cql-Sdk.slnf` Release build clean incl. PublicAPI analyzer; `CqlToElmTests` 2497 passed / 33 skipped.

## FirelyTeam Checklist

---


alexzautke:
## Parallel A/B — the missing dimension of the dop-1 measurement

The sequential A/B above left an open question: the removed cache is process-wide shared state behind a coarse `lock` (`TryGetValue` and `Insert` both lock the same monitor, on reads too), so dop 1 could only ever see its benefit, never its contention cost. Re-ran the identical corpus A/B on the same arms at **dop 4** (primary, matches vCPUs) and **dop 8** (oversubscription probe), 4 order-balanced rounds each, rebuild per arm switch, every run confirmed to reach its requested concurrency.

**The mechanism is real and directly observable.** Monitor-contention counts per repetition (2,851 cases):

| | A (cache) median `cont` | A reps > 0 | B (no cache) median `cont` | B reps > 0 |
| --- | --- | --- | --- | --- |
| dop 4 | 2.5 | 20/24 | 0 | 3/24 |
| dop 8 | 12.0 | 23/24 | 0 | 4/24 |

At dop 1 both arms were 0 everywhere. The cache's lock is a genuine serialization point under concurrent evaluation; the uncached arm is contention-free.

**But it does not invert the result — it dissolves it.** B never becomes absolutely faster; its sequential penalty just evaporates:

| B vs A | dop 1 | dop 4 | dop 8 |
| --- | --- | --- | --- |
| wall | +5.70 % | +1.15 % | +0.39 % |
| cpu | +5.66 % | +0.45 % | +0.43 % |
| alloc | +11.32 % | +11.64 % | +10.77 % |

The parallel wall/CPU deltas fail the round-agreement test (signs mixed, rounds swinging ±10–25 % on a ~90 %-saturated 4-vCPU box), so the honest reading is **"at dop ≥ 4 the arms are indistinguishable in wall and CPU"** — not "B closed a measured gap". Scaling from each arm's own dop-1 baseline: B ×2.71 vs A ×2.59 at dop 4 (×2.53 vs ×2.40 at dop 8) — directionally the lock hypothesis, within noise. A also pays ~90–95 ms more parallelization CPU overhead than B at both dops, the expected footprint of a contended monitor, though that's a cross-session comparison and only suggestive.

**Allocation is the metric that survives every cut**: +10.8…+11.6 % at all three dops, dop-independent per arm within 1 % (which doubles as the run-validity check), all gen0.

**Net for the decision:** the dop-1 "cache is 1.5–6 % faster" reading was an artifact of measuring a shared lock without concurrency. Under the concurrent evaluation this SDK actually serves, removal costs nothing measurable in time, removes a demonstrably contended lock, and costs ~11 % allocation as short-lived gen0 garbage. The release fragment now states all three dops (`aebc41845`).

---

alexzautke:
All three review comments addressed in `1aefb2eb0` (plus `536910b38` for the CI break):

1. **"closes #1483" wording** — valid. The fragment cites `(#1487, #1525)` per the `(#issue, #PR)` convention the released notes use; #1483 stays linked in prose where the measurement is discussed. The same citation defect at the end of the Potentially Breaking section was fixed too.
2. **Never-shipped overload in the breaking table** — valid, confirmed against the PublicAPI ledgers: the 3-arg `Create` overload exists only in `Unshipped.txt` (its line was replaced, not `*REMOVED*`-prefixed), while the other three rows are all in `Shipped.txt` with `*REMOVED*` entries. The row is kept — code built against unreleased `develop` may call it — with a `†` footnote stating it never appeared in a released version, so release-to-release upgrades cannot break on it. One provenance correction: it was added by #1458 (alongside `OverrideConverterTimezoneOffset`), not by the #1482 work.
3. **Diagram omitting `OverrideConverterTimezoneOffset`** — valid by the diagram's own selection rule (it lists options used for runtime operator construction, and this one feeds `FhirTypeConverter.Create` — now the only thing distinguishing converter instances). Added in the slot the removed cache-size option vacated; `EvaluationCache` stays correctly excluded (read in `CreateContext`, not `CreateOperators`). SVG regenerated via `tools/mermaid/export-mermaid-svg.sh`, not hand-edited.

Separately, `536910b38` fixes the CI failure on `Cql-Sdk-All.sln`: `Benchmarks.csproj` compiled the deleted `CoreTests/Shared.cs` in by link (it existed only for the cache-inspection helpers, and no benchmark uses it) — invisible to the `Cql-Sdk.slnf` local validation, which doesn't carry the Benchmarks project.

---

baseTwo (changes requested):
## Review of `1aefb2eb0` — the removal is right and well evidenced; it deviates from the agreed plan on one point that needs a decision

@alexzautke — picked this up off the board. 28 behind `develop`, all 8 checks green.

### Verified

**The behaviour lock does what AC #2 asks.** I checked out `FhirDateTimeConversionTests.cs` onto **`develop`'s**
implementation — cache and all, `LRUCache.cs` still present — and ran it there: **34/34 pass**. So the tests pin
conversion behaviour that already exists rather than describing the post-removal code, which is exactly what makes
them able to catch a regression from the removal. They pass on this branch too.

**The disclosure is the most thorough I have seen in this repo.** Every removed member is tabled with its
replacement, migration guidance covers the positional-record subtlety on `ElmToolkitConfig`, the disappearing
`ArgumentOutOfRangeException` is called out, and the † footnote correctly separates the one overload that only ever
existed inside this unreleased window from the three that are in the released surface. The PR carries the
`breaking change` label. None of that is in question.

**`LRUCache` is genuinely gone** from the runtime path, `docs/caching-strategies.md` is updated, and the dependency
diagram was regenerated as both `.mmd` and committed `.svg` per the repo's Mermaid convention.

### 1. AC #1 and AC #4 are unmet as written — obsoletion vs removal needs an explicit decision

[#1487](https://github.com/FirelyTeam/firely-cql-sdk/issues/1487) says, in the Mechanics section:

> **Keep the `cacheSize` parameter surface but mark it `[Obsolete]`** (accepted-and-ignored) on
> `FhirTypeConverter.Create` and the `ElmToolkitConfig` knob — non-breaking now, removal at the next major.

and AC #4 says "no public API removals (obsoletions only)". This PR removes them instead — `grep Obsolete` over the
whole diff returns nothing, and `PublicAPI.Unshipped.txt` carries five `*REMOVED*` entries:

```
*REMOVED*ElmToolkitConfig.LRUCacheSize.get / .init
*REMOVED*ElmToolkitConfig..ctor(...) and .Deconstruct(...)   (positional record arity change)
*REMOVED*FhirCqlContextOptions.OverrideFhirTypeConverterCacheSize.get / .init
*REMOVED*FhirTypeConverter.Create(ModelInspector, int? cacheSize = null)
```

Three of those are in the **released** public API by your own footnote's reckoning, so consumer code stops
compiling on a minor upgrade rather than at the next major — which is the specific outcome the agreed plan was
written to avoid.

I am not asserting the removal is wrong. It is cleaner, and the migration is mechanical. But it is a different
decision from the one recorded on the issue, and it should not land by absorption. Either:

- **restore the obsoletion** — keep the parameter and the two properties, `[Obsolete]` with a message pointing at
  #1487, accepted-and-ignored, exactly as the issue specifies; or
- **record the deviation on #1487** — amend Mechanics and AC #4 to say "removed now, not obsoleted", with whoever
  owns the release cadence agreeing that a compile break inside 2.x is acceptable here.

I lean toward the second only if someone with release authority signs it off; absent that, the first is what was
agreed. Your call which, but the issue and the PR should not disagree once this merges.

### 2. Question: `## Potentially Breaking` or `## Breaking`?

The fragment files this under `## Potentially Breaking`. Removing released public members is not potential — code
using them will fail to compile. House convention is genuinely mixed (2.7.0 filed an `ICqlOperators.Contains<T>`
signature change under *Potentially* Breaking), so I am raising it as a question rather than a defect. Given the
table of removals reads as a migration guide, `## Breaking` seems the more honest heading.

---

Nothing else. The measurement story, the tests and the documentation are all in good shape — item 1 is purely about
the plan of record matching what ships.

alexzautke:
Suggest that @ewoutkramer does the final sign off. I would simply kill this otherwise it will become a maintenance burden.

baseTwo:
@alexzautke — agreed, and your reasoning is the stronger one. An accepted-and-ignored `[Obsolete]` parameter is a
lie the compiler helps you tell: callers keep passing a cache size, nothing caches, and we carry the shim plus its
tests until someone remembers to delete it. Killing it now is cleaner.

To be clear about my changes-request: it is **not** asking you to revert to obsoletion. It is asking that #1487 and
the PR stop disagreeing, because the issue's Mechanics and AC #4 explicitly say "obsoleted, not removed" and
"no public API removals". Either the code follows the issue or the issue follows the code — and given you are
proposing the latter, @ewoutkramer signing off is exactly the right gate. Once that lands I will amend #1487's
Mechanics and AC #4 to record the decision, and approve.

@ewoutkramer — the decision in one paragraph, so you do not have to reconstruct it:

**What is being decided.** Whether to delete the FHIR date/time cache-size knob from the public API now, or keep it
as an accepted-and-ignored `[Obsolete]` parameter until the next major. #1487 planned the second; this PR does the
first; alexzautke argues the shim is a maintenance burden.

**What actually breaks.** Three released members — `FhirTypeConverter.Create(ModelInspector, int? cacheSize = null)`
(`PublicAPI.Shipped.txt:40`), `FhirCqlContextOptions.OverrideFhirTypeConverterCacheSize`, and
`ElmToolkitConfig.LRUCacheSize`. Consumer code using any of them stops compiling on a minor upgrade. The migration
is mechanical — drop the argument, drop the initializer; the value was only ever a cache bound, so nothing needs a
behavioural substitute. `ElmToolkitConfig` is a positional record, so anyone constructing it positionally past
`AllowInvalidCSharp` also has to drop that argument.

**What is not in question.** The removal itself is well evidenced (#1483's measurements: no CPU benefit, and the
cache-consult path was 8.6% of active CPU), the behaviour locks are real — I ran the new
`FhirDateTimeConversionTests` against **develop's cached implementation** and got 34/34 — and the fragment already
documents every removal with migration guidance and a `breaking change` label.

---

Separately, the three automated review comments do not hold, so please do not spend time on them:

1. **"The release note says 'closes #1483'"** — it does not. The word "closes" does not appear anywhere in the
   fragment; it says *"The trade was measured in #1483"* and cites `(#1487, #1525)`.
2. **"The 3-arg `Create(model, cacheSize, defaultTimezoneOffset)` overload is unshipped, so listing it as breaking
   is misleading"** — correct that it is unshipped, and the fragment already says so: that row carries the †
   footnote explaining it never appeared in a released version and only code built against unreleased `develop`
   is affected. Already addressed.
3. **"The diagram omits `OverrideConverterTimezoneOffset`"** — it does not.
   `dependency-diagrams.runtime-services.mmd:16` lists it explicitly in the `FhirCqlContextOptions` node, between
   `OverrideModelInspector` and `ResourceIdComparer`.

baseTwo:

baseTwo:
## ⚠️ The conflict resolution in `9805d1e62` silently reverted #1471 and introduced a new hot-path regression

@alexzautke @copilot — please do not merge this until the `TypeConverter.cs` resolution is redone. This PR does not
touch `TypeConverter.cs` at all, so its diff against `develop` for that file should be empty. It is not.

`git diff origin/develop origin/1487-remove-lru-cache -- Cql/Cql.Runtime/Conversion/TypeConverter.cs` shows the
merge **added back** everything [#1471](https://github.com/FirelyTeam/firely-cql-sdk/pull/1471) deleted an hour ago
(`376ad5393`): `_conversionsUsed`, `_conversionsAvailable`, `_logger`, `UseLogger`, `CaptureAvailableConverters`,
`LogFinalConverters`, `TypesToString` — plus the `.UseLogger(logger)` / `CaptureAvailableConverters()` wiring in
`ElmToolkitServices.cs`.

**And the result is not simply "pre-#1471". It is worse than any state that has ever existed on either branch.**

| | `CanConvert` | `Convert` (hot path) |
| --- | --- | --- |
| Before #1471 (`8fa74ccdc`) | adds to `_conversionsUsed` | **does not** |
| `develop` today | nothing — deleted | nothing — deleted |
| **This branch after `9805d1e62`** | adds, **unguarded** | **adds — new** |

Two specific problems:

1. **`Convert` now formats a diagnostics string on every conversion.** Line 146:
   `_conversionsUsed.Add(TypesToString((fromType, to)));` inside `Convert`. `TypesToString` runs four
   `string.Replace` calls over each type's namespace. `Convert` is the per-value hot path — this is precisely the
   cost [#1470](https://github.com/FirelyTeam/firely-cql-sdk/issues/1470) was filed to remove, now applied to a
   path that never carried it. It would eat a visible share of the 47× win #1471 just landed.
2. **Both adds are unsynchronized.** No `lock`, and no `if (_logger is null) return;` guard — so a `HashSet<string>`
   is mutated concurrently on every conversion, under the concurrent-evaluation contract documented on
   `CqlContext`. #1471's own intermediate revision had a lock and a logger guard; this has neither.

Nothing reads `_conversionsUsed` either — `LogFinalConverters` is still only reachable from the commented-out line
in `Dispose()`, which is why #1471 deleted the whole apparatus.

### The fix

`TypeConverter.cs` and `ElmToolkitServices.cs` should match `develop` exactly. This PR's own change to
`ElmToolkitServices` is only dropping the `cacheSize` argument, so the correct resolution is develop's file with
`FhirTypeConverter.Create(modelInspector, lruCacheSize)` → `FhirTypeConverter.Create(modelInspector)` and nothing
else:

```bash
git checkout origin/develop -- Cql/Cql.Runtime/Conversion/TypeConverter.cs
git checkout origin/develop -- Cql/CodeGeneration.NET/Toolkit/Internal/ElmToolkitServices.cs
# then re-apply only the cacheSize removal in ElmToolkitServices
```

Afterwards `git diff origin/develop...HEAD -- Cql/Cql.Runtime/Conversion/TypeConverter.cs` must be **empty**.

Worth a general note: this is the failure mode of resolving a conflict by picking a side wholesale. The two PRs
touched adjacent lines for unrelated reasons, and taking this branch's side restored code whose deletion was the
entire point of the other one.

The remove-vs-obsolete question for @ewoutkramer is unaffected and still open.

alexzautke:
We deceived to remove this to reduce maintenance.

baseTwo:
## Regression fixed in `d57627a7d` — verified

The decisive check now passes:

```
git diff origin/develop origin/1487-remove-lru-cache -- Cql/Cql.Runtime/Conversion/TypeConverter.cs
(empty)
```

#1471's deletion is intact — no `_conversionsUsed`, no `_logger`, no per-call `TypesToString` on `Convert`, and the
unsynchronized `HashSet` mutation is gone. `ElmToolkitServices.cs` now diffs against `develop` by exactly this PR's
intended change and nothing else: `lruCacheSize` dropped from `AddCqlCompilerServices` and from the
`FhirTypeConverter.Create` call, with no logger wiring resurrected.

This PR's own change is unaffected — `LRUCache.cs` still deleted, same 19 files, +328/−432.

Thanks for the quick turnaround.

---

Still open and unchanged: the **remove vs. obsolete** decision. @ewoutkramer has not weighed in yet, and that is the
only thing blocking from my side. To restate it compactly so it needs no reconstruction:

#1487's Mechanics and AC #4 say to `[Obsolete]` the `cacheSize` surface and defer removal to the next major. This PR
removes it now — three of the removed members are in the **released** public API, so consumer code stops compiling
on a minor upgrade. @alexzautke argues, persuasively, that an accepted-and-ignored `[Obsolete]` parameter is a
maintenance burden and a lie the compiler helps you tell. Either position is defensible; what should not happen is
the issue and the PR disagreeing after merge. Once there is a decision I will amend #1487 to match and approve.

baseTwo:
@copilot, based on a discussion I had between @alexzautke and @ewoutkramer in our daily meeting, it was decided to  delete the FHIR date/time cache-size knob from the public API now. It's a really small surface for a niche feature on the API, and since it doesn't benefit performance anymore, it should be removed. The Release Notes must clearly indicate this

baseTwo (approved):
## Approving `7f80a9288` — the removal decision is recorded, and the merge regression is verified gone

The remove-vs-obsolete question is settled: **remove now**. I have amended
[#1487](https://github.com/FirelyTeam/firely-cql-sdk/issues/1487) so the issue and the PR agree — its Mechanics
section no longer says "keep the surface but mark it `[Obsolete]`", AC #1 now says the knob is *removed*, and AC #4
now says public API removals are expected and documented rather than "obsoletions only". Each amendment is dated
and states what it replaced, so the history is not rewritten silently. That was the only thing my review was
blocking on.

### Verified at this head, not carried over

**The merge regression is gone and stays gone.** The check that matters:

```
git diff origin/develop origin/1487-remove-lru-cache -- Cql/Cql.Runtime/Conversion/TypeConverter.cs
(empty)
```

A file this PR does not touch no longer differs from its base. `ElmToolkitServices.cs` differs by exactly the
intended change — `lruCacheSize` dropped from `AddCqlCompilerServices` and from the `FhirTypeConverter.Create`
call — with no logger wiring resurrected. I re-ran this at `7f80a9288` rather than trusting the check I did at
`d57627a7d`, since the head moved in between.

**The behaviour lock genuinely holds.** `FhirDateTimeConversionTests` passes 34/34 against **develop's cached
implementation**, so it pins conversion behaviour that already exists rather than describing the post-removal code.

**The `## Breaking` question I raised is answered** — the fragment moved off `Potentially Breaking` and now leads
with the rationale for removing now rather than deferring. Right call: nothing about a removed released member is
potential.

### One thing to fix at release-cut time, not now

The new heading is `## Breaking changes`. Every shipped release note uses `#### Breaking` (20 occurrences, against
11 of `#### Potentially Breaking`), and `docs/releases/vnext/README.md` says fragments "use the same section
headings as the versioned release notes". This is the only fragment using `Breaking changes`, so whoever runs
`cut-release-notes` should fold it under `Breaking`. Not worth another round — recording it here so it is not
missed.

PR #1565, opened 2026-08-13
Short-circuit CQL and/or/implies/xor natively, over a new CqlBoolean type

baseTwo:
Fixes #1514. Also closes #1571 (native `implies`). Two stacked PRs have been merged into this branch and are described here rather than separately: #1572 (`CqlBoolean`) and #1574 (`&&`/`||` lowering).

CQL's logical operators no longer compile to `ICqlOperators` calls in generated C#. They compile to C#'s own operators, and `and`/`or`/`implies`/`xor` now **short-circuit**: the right operand is evaluated only when the left has not already decided the result. Getting from "no short-circuit" to "short-circuit as an ordinary expression" required a new primitive type, and that is the part worth reviewing closely.

### 1. Native lowering of the logical operators

| CQL | Lowers to | Right operand skipped when left is |
|---|---|---|
| `and` | `&&` over `CqlBoolean` | `false` |
| `or` | `\|\|` over `CqlBoolean` | `true` |
| `implies` | `!left \|\| right` (exactly Kleene implication) | `false` — and §9.B permits this skip *explicitly* |
| `xor` | lifted `^` behind a guard statement | `null` — the odd one out |
| `not` | lifted `!` | n/a |
| `is true` / `is false` | the `is true` / `is false` constant patterns | n/a — total predicates, nothing to skip |

C#'s lifted `&`/`\|`/`!`/`^` on `bool?` **are** CQL's three-valued logic, so the values are unchanged by construction; the truth tables are still checked against the runtime operators (see Tests).

`xor` is the exception on two counts: its deciding value is `null` rather than a bool (its null row is constant — "if either or both arguments are null, the result is null"), and C# has no `^^`, so it keeps the guard-statement shape instead of becoming an expression.

On top of that, a **deciding constant** operand on either side folds at build time and cascades through nested chains — `X and false` → `false`, `X or true` → `true`, `X xor null` → `null` — erasing the other operand entirely.

Every guard and operator carries a traceability comment naming the CQL operator and its source span, e.g. `/* CQL 'or' (35:14-35:91) */`.

### 2. `CqlBoolean` — why `bool?` was not enough (#1572)

`bool?` cannot short-circuit. C# synthesises `&&`/`||` from a type's own `operator true`/`operator false`, and those **cannot be lifted over `Nullable<T>`**; the lifted `&`/`|` always evaluate both operands. The first increment therefore had to express short-circuiting as branching control flow — guard statements around the right operand.

[...]

Issue #1514, opened 2026-08-07
Short-circuit And/Or in generated code


alexzautke:
> **Update 2026-08-13** — implemented in #1565, with these decisions on top of the original
> proposal:
>
> 1. **`and`/`or`/`not` lower to C#'s lifted `&`/`|`/`!` over `bool?`**, not to guarded
>    `ICqlOperators` calls: the lifted operators' semantics are exactly CQL's three-valued
>    logic (§9.B), so for these three operators the C# language itself is the reference and no
>    runtime call remains. `not` (no short-circuit dimension) was included in the scope.
> 2. **The guard** is the proposed branching form, from new statement-shaped IR (`CodeLet`
>    binding the left operand once + a guarded `CodeConditional`); in expression contexts it
>    prints as a lazy ternary. Guard tests print as constant patterns (`is false`/`is true`).
>    Deciding/absorbing constant operands fold at build time (also a correctness requirement —
>    see the CS0037 discussion in #1565).
> 3. **Traceability comments** on every guard, e.g.
>    `// CQL 'and' (33:5-33:57): right operand skipped when left is false`, from the ELM locator.
> 4. **`CSharpGeneratingConfig.PreferNoElseBlocks`** (on `ElmToolkitConfig.CSharpGeneratingConfig`) (default `false`): opt-in guard-clause
>    formatting; the default keeps `if`/`else` blocks.
> 5. **The `Lazy<bool?>` overloads and `And`/`Or`/`Not` stay public** — previously generated
>    libraries still call them, and they are now the reference implementation the lowering is
>    tested against. Lifecycle recorded in `docs/generated-code-operator-ledger.md` and in
>    `<remarks>` on the members.
> 6. `GeneratorToolVersion` 5.2.1.0 → 5.2.2.0 (patch; no new invoker). Runner-resource
>    regeneration (5.3.2.6) follows after #1563 merges.

## Problem

`Cql.Compiler` flattens every boolean expression into straight-line SSA locals, so **both operands
of `and`/`or` are always evaluated** before the operator is called. From a generated HEDIS AIS-E
library:

```csharp
private bool? Denominator_1_Compute(CqlContext context)
{
    bool? a_ = this.Initial_population_1(context);
    bool? b_ = this.Exclusion(context);              // evaluated even when a_ is already false
    bool? c_ = context.Operators.Not(b_);
    bool? d_ = context.Operators.And(a_, c_);
    return d_;
}
```

[...]

Issue #1477, opened 2026-08-01
Retrieve results are lazy and re-run their filters on every enumeration


alexzautke:
## Summary

`BundleDataSource.Retrieve<T>` (and the `IndexedBundle` machinery behind it) returns lazily evaluated LINQ sequences. Retrieve results get walked repeatedly during an evaluation — a definition's memoized value is read from several expressions, `exists` checks enumerate it, cross joins re-enumerate their inner source — and every walk re-runs the entire deciding chain, even though the bundle is immutable for the duration of the evaluation. The results should be materialized once per retrieve.

A working implementation exists on the `claude/sdk-cql-performance-j3uz16` branch; this ticket tracks landing it.

## Observed Behavior

Three lazy layers, each re-executed per enumeration of a retrieve result:

1. **`IndexedBundle.FilterByType<T>()`** (no code filter) returns `resources.Cast<T>()` — a bundle-sized cast per walk.
2. **`IndexedBundle.Filter<T>`** (code-filtered retrieves) is a `yield return` iterator that re-tests every candidate's codings against the filter predicate per walk — for value-set retrieves that means re-asking `IValueSetDictionary.IsCodeInValueSet` for every candidate coding, every time.
3. **`BundleDataSource.ApplyProfileFilter`** stacks a `Where` on top, re-running the profile predicate per walk.

Since the answers cannot change mid-evaluation (the bundle is immutable while it is being evaluated), every re-run reproduces the same membership decision.

## Root Cause

Retrieve plumbing returns deferred `IEnumerable<T>` chains where the natural contract for an immutable snapshot is a materialized list.

## Expected Behavior

- `Retrieve<T>` returns a fully materialized `IReadOnlyList<T>` (still satisfying the `IEnumerable<T>` that `IDataSource.Retrieve<T>` promises — the interface is unchanged).
- `FilterByType<T>()` caches the per-type cast once per bundle instance (a `ConcurrentDictionary` keyed by resource type — bounded by the number of resource types in the bundle, lifetime tied to the bundle, holding resources the bundle already owns).
- Code filters and the profile filter run eagerly, once, at retrieve time.
- **Identical elements in identical order** for every retrieve shape — this is a behavior-preserving change. The trade is memory (result lists live as long as the evaluation) for not re-deciding membership per walk; bounded by bundle size.
[...]

PR #1578, opened 2026-08-18
Evaluate non-inlinable short-circuit operands eagerly instead of in a local function

baseTwo:
**Draft, for review of the approach rather than for merge** — the integration suite has not been run against this output yet, and three of the four corpora are not regenerated, so CI will fail on the golden tests. See "Outstanding" below.

This is a sub-PR of #1565. It answers @ewoutkramer's review question there — *"if there are more retrieve sites, the overall performance might be slower"* — with measurements, and then changes the PR in response to them.

---

### The measurement that prompted this

`IDataSource` has exactly one member, so a counting decorator was injected by reflection into `CqlOperators.DataSource` after `FhirCqlContext.ForBundle` had bound everything. Case sets verified identical by ID; counts deterministic (a 3-iteration run returns exactly 3× the 1-iteration count); a provenance probe confirmed the generator stamp on every run.

Two traps worth recording, because both silently produce wrong comparisons: `CQL_CMS_TEST_RUNNER_CmsMeasuresDllFile` is only validated and never loads anything, and the runner loads a **pre-compiled assembly embedded in each vendored FHIR `Library` resource** — so the comparison variable is the checkout, not the DLL.

**Runtime `Retrieve` invocations, 231 test cases:**

| measure | cases | develop | #1565 |
|---|---|---|---|
| CMS22FHIRPCSBPScreeningFollowUp | 40 | 1,675 | 2,144 (+28.0%) |
| CMS0334FHIRPCCesareanBirth | 136 | 6,652 | 9,412 (+41.5%) |
| CMS156FHIRHighRiskMedsElderly | 55 | 5,815 | 5,795 (−0.3%) |
| **total** | **231** | **14,142** | **17,351 (+22.7%)** |

**Wall clock** (CMS0334, 136 cases × 3 iterations, two interleaved passes, median ms/iter): branch 1.460 / 1.506, develop 1.505 / 1.476. The sides swap rank between passes — no measurable difference either way. Since #1477/#1480 `BundleDataSource.Retrieve` returns a materialized result, so a retrieve is a cheap filtered enumeration over an in-memory bundle. Against a real data source the +22.7% would matter; this harness cannot show it.

### Root cause

Attribution by calling frame located it exactly. In `PCMaternal.lastTimeOfDelivery`, develop computes a call once and uses it twice; #1565 splits that define body into sibling operand functions and each re-emits the call, taking its call sites from 3 to 4. That function performs 2 retrieves per call, so **that single duplicate accounted for +2,802 retrieves** — essentially all of CMS0334's regression.

The cause is **not** `&&` and **not** `CqlBoolean`. It is that hoisting an operand into a local function gives it its own scope and therefore its own dedup table, so subexpressions the enclosing scope already holds get emitted again inside it. Corpus-wide, 145 of 453 sibling groups (32%) share a subexpression, giving **371 removable duplicate evaluations** — of which 75 are calls into other CQL functions, each dragging a whole body along.

### The change

One method in the emitter. A short-circuit operand that cannot print inline is now **evaluated eagerly into the enclosing scope**, and the operator position becomes a reference to the resulting local. One scope, one dedup table.

| | develop | #1565 | this PR |
|---|---|---|---|
| retrieve invocations (231 cases) | 14,142 | 17,351 (+22.7%) | **15,182 (+7.4%)** |
| CMS22 | 1,675 | 2,144 | **1,675** — exact parity |
| CMS0334 | 6,652 | 9,412 | **7,732** |
| CMS156 | 5,815 | 5,795 | **5,775** — better than develop |
| duplicated `hospitalization…` call sites | 3 | 4 | **3** — parity |
| zero-arg operand functions | 0 | ~1,900 | **0** |
| generated lines | 111,742 | ~118,900 | **106,537** |

**68% of the retrieve regression removed.** The corpus is now ~5,200 lines *shorter* than develop, where #1565 added ~7,100.

Nothing else about #1565 changes: `and`/`or`/`not`/`xor`/`implies`/`IsTrue`/`IsFalse` still lower to C# operators rather than `ICqlOperators` calls (−3,835 call sites), the `ToValue` dedup fix stands (−2,844), `CqlBoolean` stays, and the **565 short-circuit operands that print inline keep their skip** — those never leave the scope, so they cost nothing.

### The trade, stated plainly

Operands too large to inline **no longer skip**. On the one call site we could attribute, both operands were reached in ~98% of evaluations, so the skip was worth very little while the duplication it bought was expensive. The CQL specification only *permits* short-circuiting for `and`/`or` — it mandates nothing, and calls it out explicitly only for `implies` — so this is conformant. `GeneratorToolVersion` 5.3.7.0 → 5.3.8.0 (patch, inside the invoker's `[5.1.0.0, 5.4.0.0)` window).

---

## Auxiliary Work

None. The commit is the emitter change, the version bump, and the regenerated `dqm-content-qicore-2025` corpus.

---

## Also tried, and rejected

Recorded so they are not retried:

- **Inherit dedup into operand bodies / pass shared values as parameters** — reaches only duplicates of locals the enclosing scope still holds (~15–23% of the regression). Implemented and measured at 16,617 retrieves.
- **`static` local functions with parameters** — needs an *exact* read set, since one miss is CS8421. Tracking references during linearization both misses and over-reports; deriving them from the printed text fixes that (115 errors → 16) but cannot tell a variable read from an identifier in tuple **type syntax**, and the heuristic for that broke genuine reads (1051 errors). There is no single point where a local's name becomes text — atoms carry pre-rendered strings and a `let` rebinds a name to another atom's code — so this needs `Atom` to carry the locals it reads. A code comment records this.
- **`LazyInitializer` for a shared value** — throws on a null value (`ValueFactory returned null`; null is an ordinary CQL value), and its `Func<T>` forces a heap closure: 40 bytes allocated per invocation *even when the chain short-circuits and nothing is needed*, versus 0 for a hand-rolled flag.
- **A runtime retrieve cache** — blocked on `RetrieveParameters` being a record whose `Codes` member compares by reference (so identical retrieves miss), on `CqlContext.GetOrCompute`'s `long` key being codegen-assigned rather than hashed, and on a materialization semantics decision. Would also show ~0 gain on this harness.

---

## Outstanding

- [ ] Integration suite run against this output — **the real gate**. Compilation and counts are verified; evaluation parity is not, and #1514 previously shipped a silent mis-grouping regression that compiled and passed three local suites.
- [ ] Regenerate `LibrarySets/RR23/CSharp`, `LibrarySets/dqm-content-qicore-2025/Extracted/CSharp`, `Cql/CoreTests/CSharp`, and the runner's vendored resources. Until then the golden tests fail on text.
- [ ] Rewrite #1565's release-note fragments: the skipped-side-effect disclosure now applies only to inline operands, and "generated code grows by 7,126 lines" is now a reduction.
- [ ] Attribute the residual +7.4%, which is entirely CMS0334.

## Questions for review

1. Is **+7.4% retrieve invocations** acceptable for **−6,020 emitted operator calls and ~5,200 fewer generated lines**, given no measurable wall-clock difference on this harness?
2. Do we accept short-circuiting only where it is free (565 sites), or is full short-circuiting worth the memoisation machinery — knowing we currently cannot measure its benefit?
3. Should a benchmark whose `IDataSource` is not a pre-loaded bundle be a prerequisite before claiming any performance result in either direction?

Refs #1514

### DIFF
--- a/Cql/CodeGeneration.NET/CSharpEmitter.Scope.cs
+++ b/Cql/CodeGeneration.NET/CSharpEmitter.Scope.cs
@@ -195,7 +195,7 @@ internal partial class CSharpEmitter
                 // cases must not depend on their position in the switch.
                 case CodeBinary shortCircuit when CodeBinary.ShortCircuits(shortCircuit.Op)
                                                  && !PrintsInlineAsShortCircuitOperand(shortCircuit.Right):
-                    return LinearizeShortCircuitCallingRightOperand(shortCircuit);
+                    return LinearizeShortCircuitEvaluatingRightOperandEagerly(shortCircuit);
 
                 // Pass-through composites: printed inline over their (spine-linearized)
                 // children instead of being hoisted into a local. This mirrors the old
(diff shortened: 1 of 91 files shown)


baseTwo:
Won't do

Issue #1588, opened 2026-08-20
Adopt and document EffVer as the SDK versioning scheme


baseTwo:
## Summary

The repository has no written versioning policy, and the one document that does say something says the wrong thing. The [Creating Tags and Releases](https://github.com/FirelyTeam/firely-cql-sdk/wiki/Creating-Tags-and-Releases) wiki page — the authoritative release-process page, linked from [`cql-sdk.props`](https://github.com/FirelyTeam/firely-cql-sdk/blob/f6bfacf78e008740622024458740c4b4781bb9bb/cql-sdk.props#L26) and the [`cut-release-notes`](https://github.com/FirelyTeam/firely-cql-sdk/blob/f6bfacf78e008740622024458740c4b4781bb9bb/.claude/skills/cut-release-notes/SKILL.md) skill — opens with:

> ## Semantic Versioning Basics
> Most C# projects follow **Semantic Versioning (SemVer)**: `MAJOR.MINOR.PATCH[-PRERELEASE][+BUILD]`
> - **MAJOR**: Breaking changes
> - **MINOR**: New features, backward-compatible
> - **PATCH**: Bug fixes, backward-compatible

Practice has never matched this. Adopt and document [EffVer](https://jacobtomlinson.dev/effver/) (Intended Effort Versioning) instead, which is the scheme the release history actually follows.

Surfaced while reviewing #1587, whose first draft justified `2.14.0` by appealing to "the repository's existing convention of shipping breaking changes under minor releases, while patch releases remain non-breaking bugfix updates" — the second half of which is contradicted by `2.9.1`. A stated policy the tree contradicts is worse than no policy, because it gets cited.

## Observed Behavior

Every release from `2.7.0` to `2.13.0`, taken from the committed release-notes documents and `git rev-list` between tags:

| Release | Level | Breaking | Potentially Breaking | New API | Window |
| --- | --- | --- | --- | --- | --- |
| 2.7.0 | minor | 4 | 2 | 5 | 26 commits |
| 2.8.0 | minor | none | none | none | 26 |
| 2.8.1 | patch | none | none | none | 6 |
| 2.8.2 | patch | none | none | none | 36 |
| 2.9.0 | minor | none | 2 | none | 30 |
| 2.9.1 | **patch** | **1** | none | none | 37 |
| 2.9.2 | patch | none | none | none | 9 |
| 2.10.0 | minor | none | 2 | none | 111 |
[...]

Issue #1445, opened 2026-07-30
Phase 2: retire vnext-release-notes.md as a directly-editable file


baseTwo:
## Summary

Phase 2 follow-up to #1432 (release-note fragment-file convention). Phase 1 (fragment directory, updated instructions/skill, transitional grandfathering) landed via #1435. Phase 2 finishes the transition by removing `vnext-release-notes.md` as a directly-editable file, once a release cut confirms no PR-authored content remains directly in it.

## Status as of the v2.11.0 cut (2026-07-29/30)

The v2.11.0 release (#1441 / PR #1442) was the first cut to happen since the fragment convention landed, but `vnext-release-notes.md` still had transitional bullets directly in it at cut time (predating the convention), alongside the `docs/releases/vnext/` fragments. Per the `cut-release-notes` skill's existing branch for this case, both sources were folded in and `vnext-release-notes.md` was reset to the **empty template** (not replaced with the pointer doc) — so Phase 2's precondition ("no PR-authored content remains directly in `vnext-release-notes.md`") wasn't yet met at that cut.

This ticket tracks doing Phase 2 at the **next** release cut, assuming that cut finds `vnext-release-notes.md` empty of direct content (only fragments under `docs/releases/vnext/` contributing).

## Acceptance Criteria

(copied from #1432's Phase 2 batch)

- [ ] Replace `docs/releases/vnext-release-notes.md` with the static pointer doc: "Pending release notes are tracked as individual files under `docs/releases/vnext/`, one per PR. See that folder for everything accumulated since the last release." — instead of resetting it to the empty template.
- [ ] Drop the "also sweep `vnext-release-notes.md`" step from the `cut-release-notes` skill — fragments become the sole source from here on.
- [ ] Drop the transitional/grandfathering wording from `CLAUDE.md` and `.github/copilot-instructions/04-development-guidelines.md` §4.5.1 (bump its changelog for the removal).

## Notes

- Do this as part of a future `cut-release-notes` pass, not standalone — the skill should be updated in the same PR that does the cut, per #1432's own framing ("tracked as a follow-up checklist in that release's cutting work rather than blocking this issue").
- Before starting, re-verify the precondition: confirm `vnext-release-notes.md` has no directly-authored bullets left (only fragments contributing) — if new direct content has appeared again in the meantime, hold off and keep grandfathering.

Refs #1432

PR #1589, opened 2026-08-20
Document EffVer as the versioning scheme

baseTwo:
## Description

Documents the SDK's versioning scheme, which until now had no written statement anywhere in the repository — the only one was the release-process wiki page's claim that this project follows Semantic Versioning, which practice has never matched.

Surfaced while reviewing #1587, whose first draft justified `2.14.0` by appealing to "the repository's existing convention of shipping breaking changes under minor releases, while patch releases remain non-breaking bugfix updates". The second half is contradicted by `2.9.1`, a patch that shipped a `#### Breaking` entry. A stated policy the tree contradicts is worse than no policy, because it gets cited.

### What the history shows

Taken from the committed release-notes documents and `git rev-list` between tags:

| Release | Level | Breaking | Potentially Breaking | New API | Window |
| --- | --- | --- | --- | --- | --- |
| 2.7.0 | minor | 4 | 2 | 5 | 26 commits |
| 2.8.0 | minor | none | none | none | 26 |
| 2.8.1 | patch | none | none | none | 6 |
| 2.8.2 | patch | none | none | none | 36 |
| 2.9.0 | minor | none | 2 | none | 30 |
| 2.9.1 | **patch** | **1** | none | none | 37 |
| 2.9.2 | patch | none | none | none | 9 |
| 2.10.0 | minor | none | 2 | none | 111 |
| 2.11.0 | minor | none | 6 | 8 | 273 |
| 2.12.0 | minor | 1 | 1 | 9 | 55 |
| 2.13.0 | minor | 1 | 7 | 5 | 190 |

The major digit has never incremented for a breaking change; a patch shipped one; and a minor shipped no features. Window size does not separate the levels either — a 36-commit patch sits beside a 26-commit minor.

There is a consistent signal, just not the SemVer one. **Every patch except `2.9.1` has both "Breaking changes: None" and "Required migrations: None" in its `tl;dr`.** The digit has been tracking migration effort all along.

### The scheme

[EffVer](https://jacobtomlinson.dev/effver/) — Intended Effort Versioning — names exactly that axis: `MACRO.MESO.MICRO`, where each digit says how much work adoption costs rather than what category the change belongs to. `docs/versioning.md` is the canonical policy.

Two consequences are called out prominently because both cut against SemVer intuition:

- **A bug fix can force MESO** — when it changes CQL evaluation results or `GeneratorToolVersion`, consumers have real work to do. `2.13.0`'s `Median`/`GeometricMean` correction is the worked example: a measure whose output moves is a migration even when the old output was a defect.
- **New public API is MICRO** — additions cost existing consumers nothing.

Pre-release identifiers keep SemVer 2.0 syntax (`alpha.N`/`beta.N`/`rc.N`), because NuGet's ordering and opt-in resolution depend on it. The level is decided *before* the suffix is appended, so `VersionPrefix` always names the version intended to ship and a pre-release never defers the decision.

The `breaking change` label is documented as a **hint, not a trigger**. It has been applied to roughly half the changes that shipped in a `Breaking` or `Potentially Breaking` section — #1312, #1490, #1532 and #1568 all went unlabelled, and #1312's miss is the direct cause of `2.9.1` being cut as a patch — and once to a release-notes PR (#1560), which would have triggered a bump on the very PR that sets the version. The `docs/releases/vnext/` fragment heading decides; the label prompts you to check it.

### Files

**New**

- `docs/versioning.md` — **the only place the policy is written**: digit meanings, the MESO trigger list, what never affects the version, the two independent scales, pre-release handling, the tie-breaker (when unclear, go up), a "why not SemVer" section, and an explicit `## Deviations from EffVer` section.

There is deliberately **no** `11-versioning.md` Copilot sub-document. An earlier revision had one, restating the whole policy in 76 lines, and it had already drifted while this PR sat in review — its trigger list was complete while the `cut-release-notes` skill's copy of the same list silently omitted the `[Obsolete]` and dependency-major triggers, so a release carrying either could have been cut as MICRO. `CLAUDE.md`'s own sync rule says a task-specific procedure gets exactly one copy with the other side linking to it. What the instruction files carry now is the two universal gotchas that rule does require in both, plus a pointer.

**Changed**

- `README.md` — a `## Versioning` section with the consumer-facing summary and the pre-release note.
- `CLAUDE.md` — a `## Versioning` section mirroring the rules.
- `docs/releases/release-notes-template.md` — a **required** `### Version Level` section, guidance that the trigger must be named and not just the level, and a `Common Items To Check` entry. Also forbids justifying a level by precedent.
- `docs/releases/vnext/README.md` — a table mapping fragment `##` heading to version effect, with #1568 as the worked example of a `## Performance` fragment that still forced MESO.
- `docs/technical-readme.md` — "follow semantic versioning" scoped to `GeneratorToolVersion` so it no longer reads as an SDK-wide claim, plus a note that a generator change makes the next release at least MESO.
- `build/README.md` — "SemVer Compliance" scoped to NuGet's pre-release suffix grammar, which is genuinely SemVer-valid.
- `.claude/skills/cut-release-notes/SKILL.md` — a new first step: decide the level before writing anything, and record the trigger.
- `.github/copilot-instructions/04-development-guidelines.md` — 4.6.3 cross-referencing fragment heading to version level.
- `.github/copilot-instructions.md` — decision tree, navigation index, and version history `3.17.0`.
- `cql-base.props` — the Vonk sync comment named `FirelyNetVersion` in `src/Vonk.props`; it is `FhirNetApiVersion` in `Directory.Packages.props`. Same error as the wiki's, which is why it is fixed here rather than filed separately.

### Not in scope

- **The wiki.** `Creating-Tags-and-Releases.md` is where the wrong claim actually lives, but it is a separate git repository and cannot be changed by a PR here. The full proposed diff is in a comment below, for review alongside this PR; it needs someone with wiki write access to apply once this merges. Tracked on #1588.
- **`docs/releases/release-notes-2.14.0.md`.** That file only exists on #1587's branch, so its `### Versioning Decision` section cannot be aligned with the new template here. Tracked as an acceptance criterion on #1588, to be done once both have merged — `2.14.0` is the first release cut under this scheme, so its notes need to introduce the vocabulary rather than assume it, and its current text ends "the level remains a per-cycle judgement about consumer impact rather than a written rule", which goes stale the moment this merges.
- **No release-note fragment.** Under the policy this PR introduces, documentation changes never affect the version.
- **`CLAUDE.md`'s heading-numbering rule.** It states that `docs/` uses hierarchical heading numbers, but only 2 of 11 files there do. `docs/versioning.md` follows the local majority (unnumbered, matching `getting-started.md`). Noted on #1588 rather than settled here.

## Related issues

- Refs #1588 — deliberately **not** `Fixes`. Several acceptance criteria on that issue are
  handed off rather than done here: the wiki edits (separate repository, needs wiki write
  access), aligning `release-notes-2.14.0.md` once #1587 merges, and the `docs/` heading-number
  reconciliation. Auto-closing the only tracker would lose them.

## Testing

Documentation only apart from one comment line in `cql-base.props`, so there is nothing behavioral to test. `cql-base.props` is not in `build/azure-pipelines.yml`'s `ignorePatterns`, so CI runs the full build rather than skipping it.

Verified by hand:

- Every fact in `docs/versioning.md` was checked against the tree rather than carried over from the wiki: `FhirNetApiVersion` `6.3.0` in Vonk's `Directory.Packages.props` on `develop`; `GeneratorToolVersion` `5.2.1.0`; `LibraryInstanceInvoker_5_0`'s supported range `[5.1.0.0, 5.3.0.0)`; `VersionPrefix` declared only in `cql-sdk.props` and `Demo/cql-demo.props`; the `breaking change` label applied to 11 items total, with the four unlabelled misses confirmed against each release's `Breaking`/`Potentially Breaking` sections.
- The per-release table was generated from the committed release-notes documents and `git rev-list` between tags.
- All relative links resolve from their containing file.
- Confirmed this branch carries no part of #1587: no `VersionPrefix` change, no `README.md` release-version line change, no `docs/releases/release-notes-2.14.0.md`, no deleted `vnext/` fragments.

## FirelyTeam Checklist


baseTwo:
## Proposed wiki changes — [`Creating-Tags-and-Releases`](https://github.com/FirelyTeam/firely-cql-sdk/wiki/Creating-Tags-and-Releases)

The wiki is a separate git repository, so these cannot ride along in this PR. Posting the full proposed diff here so it can be reviewed alongside the policy it points at. **Someone with wiki write access applies this once this PR merges** — not before, or the page will link a `docs/versioning.md` that does not exist yet. Tracked as acceptance criteria on #1588.

Four changes. The first is the point of the exercise; the other three are stale items on the same page, worth fixing while someone is in there.

---

### 1. Replace the opening block — this is where the wrong claim lives

**Remove** (the `Introducation` typo goes with it):

```md
# Versioning Introducation

## Semantic Versioning Basics

Most C# projects follow **Semantic Versioning (SemVer)**: `MAJOR.MINOR.PATCH[-PRERELEASE][+BUILD]`
-   **MAJOR**: Breaking changes
-   **MINOR**: New features, backward-compatible
-   **PATCH**: Bug fixes, backward-compatible
-   **PRERELEASE**: Optional tag like `-rc`, `-beta`, `-alpha`
-   **BUILD**: Optional metadata (e.g., commit hash)
```

**Replace with:**

```md
# Versioning

The Firely CQL SDK does **not** use Semantic Versioning. It uses
[EffVer](https://jacobtomlinson.dev/effver/) — Intended Effort Versioning — where each digit
answers *how much work will this cost me to adopt?* rather than *what kind of change is it?*

`MACRO.MESO.MICRO`

| Digit | Signals to a consumer |
| --- | --- |
| **MACRO** | "You will need to dedicate significant time to upgrading." |
| **MESO** | "Some small effort may be required to make sure this version works for you." |
| **MICRO** | "This change doesn't intend for you to need to do anything." |

The canonical rules — what triggers each digit, what never affects the version, how pre-release
suffixes fit, and why not SemVer — live in the repository at
[docs/versioning.md](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/versioning.md).
**That document is authoritative; this page covers only the mechanics of tagging and releasing.**

Two things differ from SemVer intuition and catch people out: a **bug fix can force a MESO bump**
when it changes CQL evaluation results or the code generator version, and **new public API arrives
in MICRO releases**, since additions cost existing consumers nothing.

Pre-release identifiers keep SemVer 2.0 syntax (`alpha.N`, `beta.N`, `rc.N`), because NuGet's
ordering and opt-in resolution depend on it. Decide the EffVer level first, then append the suffix:
`VersionPrefix` always names the version you intend to ship.

Note that `GeneratorToolVersion` is a **separate scale** and *does* follow SemVer. Do not conflate
the two.
```

The `## Versioning Strategy Example` table (alpha/beta/rc/final) and the whole `## NuGet Considerations` section — including the `VersionPrefix`/`VersionSuffix` explanation — **stay unchanged**. They are still correct, and `docs/versioning.md` now cross-references this page for them rather than duplicating them.

---

### 2. `## Before You Start` — two stale bullets

**Before:**

```md
* Capture the Firely .NET SDK version used in the VONK CQL Plugin. These need to be in sync.
  See property `FhirNetApiVersion` in [Vonk.props](https://github.com/FirelyTeam/Vonk/blob/develop/src/Vonk.props)
...
* **[NEW] Empty out the vnext-release-notes.md file, those must be copied and manually pasted
  in github during the Release Notes creation (see below)**
```

**After:**

```md
* Capture the Firely .NET SDK version used in the VONK CQL Plugin. These need to be in sync.
  See property `FhirNetApiVersion` in [Directory.Packages.props](https://github.com/FirelyTeam/Vonk/blob/develop/Directory.Packages.props)
...
* Consolidate the release-note fragments under
  [docs/releases/vnext/](https://github.com/FirelyTeam/firely-cql-sdk/tree/develop/docs/releases/vnext)
  into the new versioned release note, then delete them. `vnext-release-notes.md` has been a static
  pointer doc since #1445 and is not edited.
```

Two separate defects here:

- **The Vonk path is wrong, and the page contradicts itself.** `## Before You Start` says `src/Vonk.props`; step 4 of `## Update Version` correctly says `Directory.Packages.props`. I verified `FhirNetApiVersion` (`6.3.0`) is in `Directory.Packages.props`. The same error was in `cql-base.props`, fixed in this PR.
- **The vnext instruction has been wrong since #1445.** Fragments live in `docs/releases/vnext/` as one file per PR; there is nothing to "empty out".

---

### 3. `## Update Version` — add the level decision, and name the props files

Insert as a new first step:

```md
1. **Decide the version level first.** Apply
   [docs/versioning.md](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/versioning.md)
   and record which trigger fired. This goes in the release-notes document's `### Version Level`
   section and in the version-bump PR's description. Do not justify the level by what previous
   releases did — the history is inconsistent, so precedent proves nothing.
```

And tighten the existing "Open all the `*.props` files" step, which overstates the work:

```md
1. Update `VersionPrefix` in **every** `*.props` file that declares it — currently
   `cql-sdk.props` and `Demo/cql-demo.props` — plus `VersionSuffix` for a pre-release.
```

---

### 4. `## Releasing` — stale note

**Before:**

```md
Note: For the CQL repo, there is no `Release Notes.md` in the repo yet. Instead, this can be
added at the step where the Release is created mentioned later on this page.
```

**After:**

```md
Note: versioned release notes live in
[docs/releases/](https://github.com/FirelyTeam/firely-cql-sdk/tree/develop/docs/releases). Paste
the new document's content into the GitHub Release body.
```

There are 11 versioned documents plus a template in `docs/releases/` — this note predates all of them.

baseTwo:
@copilot All six findings were valid and are addressed in `dd78e1674`. Two of them turned out to share a root cause, so the fix for those is structural rather than local.

**Findings 5 and 6 were both duplication artifacts.** The missing `[Obsolete]` and dependency-major triggers in `SKILL.md` were not an oversight in that file — they were drift between two copies of the same list. `11-versioning.md` carried the complete list, the skill carried an incomplete restatement, and nothing kept them in step. The version-header/history mismatch existed only because that sub-document had to be registered in three places.

So `.github/copilot-instructions/11-versioning.md` is **deleted** — 76 lines restating the whole policy. `CLAUDE.md`'s own sync rule says a task-specific procedure gets exactly one canonical copy with the other side linking to it rather than restating it, and this is the failure that rule exists to prevent. `docs/versioning.md` is now the only place the trigger list appears; the instruction files, the vnext fragment README and the skill step carry a pointer plus the two universal gotchas the sync rule does require in both. The skill step now says to walk every trigger in the canonical list and explains why it isn't copied.

Individually:

1. **`README.md:32`** — fixed. "new API is added in MICRO releases" became "**API additions on their own never raise the digit** … a release that adds API may still be MESO for some other reason." You were right that the original could be read as "API only appears in MICRO releases".

2. **`docs/versioning.md:101`** — fixed, and this was the sharpest catch. The "breaking but cheap" bullet cited "fixing an option that never worked", which is #1570 — the change we ruled **MESO** precisely because consumers passing `--cs-namespace` must regenerate. The example argued against its own section. Replaced with genuinely unreachable API (an `internal` type's accidentally-public member, or a surface added and withdrawn inside one release window so no released version ever exposed it), plus an explicit note that a defect being the *cause* of a change says nothing about the *cost* of adopting the fix.

3. **`Fixes #1588`** — fixed. Changed to `Refs #1588`, with the PR description now naming the three deferred criteria and why each is handed off (wiki edits need a different repository and write access; the `release-notes-2.14.0.md` alignment needs #1587 merged first; the `docs/` heading-number reconciliation is a separate judgement). You were right that merging would have closed the only tracker while all three were outstanding.

4. **`cql-base.props:14` twin** — fixed. `Demo/cql-demo.props:14` had the same two errors (`FirelyNetVersion` for `FhirNetApiVersion`, `src/Vonk.props` for `Directory.Packages.props`) and now matches. Both comments say "keep this in sync with the other", which is exactly why fixing one was worse than fixing neither.

5. **`copilot-instructions.md` header** — fixed. `**Version:** 3.16.1` → `3.17.0`, matching the newest history entry.

6. **`SKILL.md:10`** — fixed by removing the duplicated list, per the structural change above.

Two further changes not from your review, for context since they touch the same sections:

- `docs/versioning.md` gains a **`## Deviations from EffVer`** section. There is exactly one: MACRO may be *declared* for a new engine generation as well as measured by adoption effort, borrowing the PROJECT digit from [RomVer](https://romversioning.github.io/romver/). Enumerating it matters because the failure this policy exists to fix was a document claiming a scheme the project did not follow — an undocumented deviation would repeat that in miniature.
- That deviation closes a hole in the pre-release rule. EffVer's macro is a claim about cost, but a new engine generation gets declared *before* the cost is known, which made "never ship a pre-release whose prefix you expect to change" unfollowable in that case. Our own major digit is the worked example: `v2.0.0-alpha` was tagged 2024-06-13, fourteen months before the first 2.x final (`v2.1.0`) and while 1.x was still being tagged, so there was no released version to measure against — and no `v2.0.0` final has ever existed. A declared MACRO now fixes `VersionPrefix` up front, which no measured level may do.

Net effect of the review round: **125 lines deleted, 86 added** — the policy is shorter and lives in one file.

ewoutkramer (approved):

PR #1592, opened 2026-08-21
Expand resolved valuesets on a private copy

### LINKED ISSUE #1591: ValueSetSource.Add expands a resolver-owned ValueSet in place, mutating the host's shared instance
## Summary

When `ValueSetSource.Add(ValueSet)` receives a valueset **without** an expansion, it runs `ValueSetExpander.ExpandAsync` directly on that instance ([ValueSetSource.cs#L201-L216](https://github.com/FirelyTeam/firely-cql-sdk/blob/012dce8ce3d33f7349cbbd79ca1e574854ca2d79/Cql/Cql.Firely/ValueSetSource.cs#L201-L216)). The FHIR SDK's expander works in place — its first statement is `source.Expansion = ExpansionComponent.Create()`, and it sets `source.Expansion = null` on failure — so the SDK writes into (or wipes the expansion from) an object it does not own. On the `Load(canonical)` path that object came from the caller-supplied `IAsyncResourceResolver`, and in a host like Firely Server that resolver is backed by a process-wide conformance cache that hands **the same POCO instance to every consumer** with an explicit do-not-modify contract.

The class's own remarks acknowledge the behavior as "the one edit to a resolved valueset the SDK performs on its own" ([ValueSetSource.cs#L81-L90](https://github.com/FirelyTeam/firely-cql-sdk/blob/012dce8ce3d33f7349cbbd79ca1e574854ca2d79/Cql/Cql.Firely/ValueSetSource.cs#L81-L90)) — this issue proposes removing that one edit rather than documenting it, since the facade `ValueSetSource` keeps is the only thing the expansion is needed for.

## Impact

For any host whose resolver serves shared instances (`CachedResolver`, a server-side conformance cache):

- Other consumers of the cached instance — including code that never touches Hl7.Cql, e.g. a REST `$validate-code` walking `vs.FindInExpansion` — can observe a **half-built expansion** while the fill is in progress; the expander fills `Expansion.Contains` incrementally per include-group and nothing excludes readers.
- Two sources resolving the same canonical concurrently can both take the `!HasExpansion` branch and expand the **same instance simultaneously**, tearing the result (each `handleInclude` iteration re-reads `source.Expansion.Contains` while the other thread may have just replaced `source.Expansion`).
- A **failed** expansion nulls `Expansion` out, so a cached instance that previously carried a good (statically authored) expansion loses it for every consumer until cache eviction or process restart.

Bounding fact that keeps this latent in practice: content whose valuesets ship with static expansions never takes the mutating branch — it arms only once something first strips an expansion. It surfaced while investigating a Firely Server incident (persistent wrong `$validate-code` answers after a terminology race) as a verified standing hazard adjacent to the root cause.

## Observed Behavior

`Load(canonical)` → `FindValueSetAsync` returns the host cache's shared instance → `Add(vs)` → `build`: `!vs.HasExpansion` → `BuildExpander().ExpandAsync(vs)` mutates the shared instance (writes `Expansion`, or nulls it on failure).

## Root Cause

`build` expands the borrowed instance instead of a private copy:

```csharp
// Cql/Cql.Firely/ValueSetSource.cs (develop @ 012dce8c)
var expander = BuildExpander();
await expander.ExpandAsync(vs).ConfigureAwait(false);
return BuildFromExpansion(vs);
```

## Repro Steps

1. Construct a `ValueSetSource` over a resolver that returns a retained `ValueSet` instance without an expansion (the host-cache shape).
2. `await source.Load(canonical)` (or `Add(vs)` directly).
3. Observe the caller-retained instance now has `HasExpansion == true` — the SDK wrote into the resolver's object. Make the expander fail instead (e.g. an unresolvable include) and observe the instance's prior state is not restored / `Expansion` is nulled.

## Expected Behavior

`Add`/`Load` never mutate a resolved valueset. The expansion the source needs is computed on a private copy (`(ValueSet)vs.DeepCopy()`), the facade is built from that copy, and the caller's instance is left byte-for-byte untouched — success or failure.

Trade-off to accept: today the in-place write means a *second* source handed the same instance sees `HasExpansion` and takes the memo path; with the copy, each source expands its own copy (once per source, since the per-source `_valueSets` layer still caches by URL). Correctness over that incidental reuse; hosts that want cross-source reuse can ship static expansions or seed via `Add(string, IEnumerable<CqlCode>)`.

## Acceptance Criteria

- [x] `build` in `ValueSetSource.Add(ValueSet)` expands a `DeepCopy` and builds the facade from the copy; the caller's instance is never written to, on success or failure.
- [x] The class remarks paragraph describing "the one edit to a resolved valueset the SDK performs on its own" ([L81-L90](https://github.com/FirelyTeam/firely-cql-sdk/blob/012dce8ce3d33f7349cbbd79ca1e574854ca2d79/Cql/Cql.Firely/ValueSetSource.cs#L81-L90)) is rewritten to state the no-mutation contract.
- [x] A test adds a valueset without an expansion through a resolver and asserts the resolved instance is unchanged afterwards (still `!HasExpansion`), while the returned facade answers membership correctly.
- [x] Release-note fragment under `docs/releases/vnext/` (`## Fixes`).

Related: the in-place behavior of `ValueSetExpander.ExpandAsync` itself (and its `Task.WhenAll` fan-out over `Settings.ValueSetSource`) is an Hl7.Fhir SDK matter — a companion report is being filed against firely-net-sdk. This issue is only about `ValueSetSource` choosing to run that expander on an instance it does not own, which is fixable here regardless.

alexzautke:
## Description

**Primary Work**

`ValueSetSource.Add(ValueSet)` expanded an expansion-less valueset by running `ValueSetExpander.ExpandAsync` on the instance itself. The expander writes the computed expansion into its argument (and clears it on failure), while on the `Load` path that instance is on loan from the caller-supplied resolver — which a host conformance cache (e.g. Firely Server's) may be serving to every consumer in the process. The in-place write therefore corrupted shared state: consumers could observe a half-built expansion during the fill, two sources could expand the same instance concurrently and tear it, and a failed expansion destroyed a previously good (statically authored) expansion for everyone until cache eviction or restart.

The expansion is now computed on a `DeepCopy` and the facade built from the copy, so the caller's instance stays byte-for-byte untouched on success and failure alike. The `ValueSetSource` class remarks are rewritten to state the no-mutation contract (they previously documented the write as "the one edit to a resolved valueset the SDK performs on its own"), including the accepted trade-off: a second source handed the same expansion-less instance expands its own copy instead of finding an expansion already written — each source still expands a given canonical at most once through its per-canonical facade layer, and hosts wanting cross-source reuse can ship static expansions or seed via `Add(string, IEnumerable<CqlCode>)`.

A release-note fragment (`## Fixes`) is included.

**Auxiliary Work**

None. (The in-place behavior of `ValueSetExpander.ExpandAsync` itself, and its `Task.WhenAll` fan-out over `Settings.ValueSetSource`, are Hl7.Fhir SDK matters — a companion report is being filed against firely-net-sdk. This PR only stops `ValueSetSource` from running that expander on an instance it does not own.)

## Related issues

Fixes #1591.

## Testing

- The test that pinned the old in-place write (`ComputedExpansion_IsWrittenIntoTheInstance_SoLaterSourcesShareIt`) is rewritten as `ComputedExpansion_LeavesTheResolvedInstanceUntouched`: adding an expansion-less valueset through a resolver leaves the instance `!HasExpansion` while the returned facade answers membership correctly.
- New `FailedExpansion_LeavesTheResolvedInstanceUntouched`: a failing expansion (unresolvable include) throws without altering the caller's instance.
- Full `CoreTests` suite: 1374 passed, 0 failed, 8 pre-existing skips.

## FirelyTeam Checklist

---


Cql/Cql.Firely/ValueSetSource.cs
```diff
@@ -81,12 +81,16 @@ namespace Hl7.Cql.Fhir;
...
-/// on its own; a later <see cref="Add(ValueSet)"/> of that same instance therefore does see an
-/// expansion and does take the memo path, keyed on the component just written. That adds no staleness
-/// beyond the in-place mutation itself, which already leaves every source handed that instance looking
-/// at the same frozen expansion.
+/// not determine - so its facade stays in the per-source layer only. The expansion is computed on a
+/// private copy of the valueset: the instance handed in is on loan - on the <see cref="Load"/> path it
+/// belongs to the resolver, which may be serving the same object to every consumer in the process -
+/// and the underlying expander writes into its argument (and clears the expansion on failure), so
+/// running it on the original would edit, or on failure damage, an object this class does not own.
+/// The SDK therefore never modifies a resolved valueset. The price is that another source handed the
```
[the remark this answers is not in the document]
alexzautke: Narrowed in a041f551 (both occurrences): the remarks now state the contract as "never modifies the valueset handed to `Add` or resolved by `Load`" and name the residual explicitly — the expander still expands an expansion-less *included* valueset from the resolver in place, which is the expander's own behavior and is tracked at FirelyTeam/firely-net-sdk#3582. Protecting transitive resolver results here was considered and dropped as too complex for a shim around an upstream fix.

---

Cql/CoreTests/ValueSetSourceFacadeMemoTests.cs
```diff
@@ -162,26 +163,38 @@ public async Task ComputedExpansion_StaysPrivateToItsSource()
...
+        // is what keeps that damage away from the caller's instance.
+        const string composedUrl = "http://example.org/ValueSet/failing-composed";
+
+        var composed = ComposedValueSet(composedUrl, "http://example.org/ValueSet/does-not-exist");
+        var resolver = new InMemoryResourceResolver(composed);
+
+        await Assert.ThrowsExceptionAsync<ValueSetUnknownException>(() => new ValueSetSource(resolver).Add(composed));
 
-        Assert.AreSame(facadeB, facadeC);
+        Assert.IsFalse(composed.HasExpansion, "a failed expansion must not alter the caller's instance either.");
```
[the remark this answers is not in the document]
alexzautke: Right — the old failure path wrote and then cleared, so the end state couldn't discriminate. Fixed in a041f551: the test now subscribes to the instance's `PropertyChanged` (the generated `Expansion` setter raises it) and asserts it never fires. Against the old implementation that fails with two `Expansion` change events; the end-state `HasExpansion` assertion stays as a belt-and-braces check.

---

docs/releases/vnext/1592-valuesetsource-no-mutation.md
```diff
@@ -0,0 +1,10 @@
+## Fixes
+
+- `ValueSetSource` no longer writes into a resolved `ValueSet`: a valueset that arrives without an
+  expansion is expanded on a private copy, so the resolver's instance — which a host cache may be
+  serving to every consumer in the process — is left untouched, on success and on failure alike.
```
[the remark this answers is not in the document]
alexzautke: Scoped in a041f551: the note now claims only the instance handed to `Add`/resolved by `Load`, and explicitly says the underlying expander still expands an expansion-less included valueset in place, tracked at FirelyTeam/firely-net-sdk#3582.

---

docs/releases/vnext/1592-valuesetsource-no-mutation.md
```diff
@@ -0,0 +1,13 @@
+## Fixes
```
baseTwo, line 10: **Required: this fragment gives the release cut nothing to set the version level with.**

The fragment is `## Fixes` only, and the cross-source sentence on L11-13 frames the consequence as lost reuse ("hosts that want cross-source reuse of computed expansions"). It is more than reuse - it is an **answer change**:

- **Old:** source A (resolver RA) expands `V` in place, so `V.Expansion = E_A`. Source B, with a *different* resolver RB, then calls `Add(V)`, sees `HasExpansion`, and answers from `E_A`. RB is never consulted. The old remarks admitted exactly this ("already leaves every source handed that instance looking at the same frozen expansion").
- **New:** B expands its own copy against RB. Two sources with different resolvers handed the same expansion-less instance now give **different membership answers** - the correct ones, but different.

Second host-visible change: a host that read `vs.Expansion` back after `Add`/`Load`, relying on what the class docs called "the one edit to a resolved valueset the SDK performs on its own", no longer finds one.

[docs/versioning.md](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/versioning.md) lists as a MESO trigger: *"CQL evaluation results change - including bug fixes that correct previously wrong answers. A measure whose output moves is a migration even when the old output was a defect."* Plus the tie-breaker: *"When the level is genuinely unclear, go up."* And [the vnext README](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/releases/vnext/README.md) warns about precisely this shape - its cautionary example, #1568, was recorded in 2.14.0 under `#### Potentially Breaking` for the same reason ("`ValueSetSource` no longer observing in-place edits ... is a behavior change").

**Ask:** add a `## Potentially Breaking` section to this fragment (or, if you read the level differently, state both consequences in the `## Fixes` text) so the release cut can decide rather than infer. Suggested text:

```md
## Potentially Breaking

- `ValueSetSource` no longer writes a computed expansion into the valueset handed to `Add` or
  resolved by `Load`. Two consequences for hosts: a host that read `ValueSet.Expansion` back after
  `Add`/`Load` no longer finds one there and must expand for itself; and two sources with
  *different* resolvers handed the same expansion-less instance no longer share the first source's
  expansion - each now answers from its own resolver's world, so membership answers can move (to
  the correct ones). (#1591)
```

While here: the `breaking change` checkbox in the PR body is unticked and the label is absent. `versioning.md` calls the label a hint rather than a trigger, so the fragment decides - but if the section above goes in, the label is the consistent follow-up.

docs/releases/vnext/1592-valuesetsource-no-mutation.md
```diff
@@ -0,0 +1,13 @@
...
+
+- `ValueSetSource` no longer writes into the valueset handed to `Add` or resolved by `Load`: a
+  valueset that arrives without an expansion is expanded on a private copy, so that instance — which
+  a host cache may be serving to every consumer in the process — is left untouched, on success and
+  on failure alike. Previously the computed expansion was written into the caller's instance (and a
+  failed expansion cleared it), corrupting host-side conformance caches that share instances. Note
+  that the underlying `ValueSetExpander` still expands an expansion-less *included* valueset
+  (`compose.include`) from the resolver in place — that is the expander's own behavior, tracked at
```
baseTwo: Nit: this citation will drift. firely-net-sdk#3582 is titled *"Terminology stack issues concurrent calls to the caller-supplied resolver via internal `Task.WhenAll`"* - the in-place mutation is **"Adjacent defect B"** inside it, and the issue body itself closes with *"Happy to split adjacent defect B (in-place mutation of resolver-owned instances) into its own issue if you prefer to track it separately."*

So a consumer clicking through from a shipped release note lands on a title about something else, and if that split happens the reference points at the wrong ticket. Suggest citing it as `firely-net-sdk#3582 (adjacent defect B)` both here and in the class remarks at [ValueSetSource.cs L92-93](https://github.com/FirelyTeam/firely-cql-sdk/blob/1591-valuesetsource-expand-copy/Cql/Cql.Firely/ValueSetSource.cs#L92-L93), and updating both if B is ever split out.

Cql/Cql.Firely/ValueSetSource.cs
```diff
@@ -81,12 +81,21 @@ namespace Hl7.Cql.Fhir;
...
-/// beyond the in-place mutation itself, which already leaves every source handed that instance looking
-/// at the same frozen expansion.
+/// not determine - so its facade stays in the per-source layer only. The expansion is computed on a
+/// private copy of the valueset: the instance handed in is on loan - on the <see cref="Load"/> path it
+/// belongs to the resolver, which may be serving the same object to every consumer in the process -
+/// and the underlying expander writes into its argument (and clears the expansion on failure), so
+/// running it on the original would edit, or on failure damage, an object this class does not own.
+/// This class therefore never modifies the valueset handed to <see cref="Add(ValueSet)"/> or resolved
+/// by <see cref="Load"/>. The copy does not reach further: the expander pulls <c>compose.include</c>
+/// dependencies from the resolver on its own and still expands an expansion-less <em>included</em>
```
baseTwo, line 91: **The hazard #1591 was filed for is still reachable through `compose.include` - this wants your ruling, not necessarily a change.**

I verified the residual: outer -> include(inner, expansion-less) -> include(leaf, expanded). After `Add(outer)`, `inner.HasExpansion` is `true` with one `Expansion` PropertyChanged event on the caller's instance. So in the motivating scenario - Firely Server's shared conformance cache - a cached expansion-less valueset that some *other* valueset includes is still written to, and a failed expansion still nulls it out for every consumer. That is the same corruption #1591 describes, one level down.

The resolved thread says protecting transitive results was "dropped as too complex for a shim around an upstream fix". I tested that judgement, and it is smaller than it sounds: a copy-on-resolve decorator installed **only** into the expander's settings (`ValueSetExpanderSettings.ValueSetSource`, set in `BuildExpander`) closes it completely - `inner.HasExpansion=False`, zero PropertyChanged events, membership unchanged (`111=True`, `999=False`):

```csharp
#pragma warning disable CS0618  // the two IAsyncResourceResolver members are obsolete
private sealed class CopyOnResolve(IAsyncResourceResolver inner) : IAsyncResourceResolver
{
    public async Task<Resource?> ResolveByUriAsync(string uri) => Copy(await inner.ResolveByUriAsync(uri));
    public async Task<Resource?> ResolveByCanonicalUriAsync(string uri) => Copy(await inner.ResolveByCanonicalUriAsync(uri));
    private static Resource? Copy(Resource? r) => r is ValueSet vs ? (ValueSet)vs.DeepCopy() : r;
}
#pragma warning restore CS0618
```

Copying only `ValueSet` results is deliberate: those are what the expander writes to, and it keeps large `CodeSystem` resources (`getAllConceptsFromCodeSystem`) out of the copy.

The framing point behind it: the remarks call this "the expander's own behavior", but it is *this class* that hands the expander the resolver it mutates through, so the mutation of included valuesets is this class's choice on exactly the argument the PR already makes for the top-level instance.

Real trade-offs, which is why this is your call and not a demand:

1. A deep copy per resolved dependency on the cold expansion path.
2. Loss of the incidental reuse where an already-in-place-expanded include is found expanded. That bites `ToValueSetDictionaryAsync`, where every valueset sits in one `InMemoryResourceResolver` and grouping-heavy HEDIS content would re-expand shared includes.

Deferring to upstream is defensible. The part I would not leave as-is is shipping a release note that says the fix protects host caches while this path still corrupts them: either close it here, or say in the note that a host cache is *still* corrupted through an expansion-less included valueset. The note currently says the expander mutates it; it does not say what that costs the host the fix is for.

baseTwo (commented):
## Review

Checked the branch out into a scratch worktree and verified the claims rather than reading them:

- **Both new tests actually discriminate.** With the fix reverted (`ExpandAsync(vs)` / `BuildFromExpansion(vs)`), `ComputedExpansion_LeavesTheResolvedInstanceUntouched` fails on the `HasExpansion` assert and `FailedExpansion_LeavesTheResolvedInstanceUntouched` fails with exactly `Expected:<0>. Actual:<2> ... Expansion, Expansion` - the write-then-clear pattern its comment describes. 12/12 pass with the fix. These are real regression tests, not shape tests.
- **`DeepCopy` is the only route available.** `ValueSetExpander` in `Hl7.Fhir.Conformance 6.3.0` exposes exactly one mutating `Task ExpandAsync(ValueSet)` and nothing else, so there is no non-mutating overload to prefer.
- **The documented residual is real.** outer -> include(inner, expansion-less) -> include(leaf, expanded): after `Add(outer)`, `inner.HasExpansion` flips to `true` with one `Expansion` PropertyChanged event. The remarks and the release note are accurate on this.

The approach is right and the core change is a clean two-line fix. Five findings inline; one more below, because it lands outside the diff hunks and cannot be anchored.

### Not anchorable: `Add(ValueSet)` and `Load` do not state the contract themselves

The no-mutation guarantee is now the single most consumer-relevant fact about `Add`, and it lives only as a paragraph inside an 80-line class `<remarks>`. `Add(ValueSet)` (L194-196) still says only *"Adds a `ValueSet` to the cache, so it will not be retrieved using the resolver."*, and `Load` (L308-311) likewise. Please add one clause to each `<summary>` - for `Add`, *"The instance is not modified: an expansion this source has to compute is computed on a private copy."*, and for `Load` the same in one line. That is where IntelliSense will actually show it, and a consumer checking whether their cached instance is safe should not have to read the class remarks to find out.

Cql/Cql.Firely/ValueSetSource.cs
```diff
@@ -81,12 +81,21 @@ namespace Hl7.Cql.Fhir;
...
+/// valueset in place - that is the expander's own behavior, tracked at
+/// <see href="https://github.com/FirelyTeam/firely-net-sdk/issues/3582"/>. The price of the copy is
+/// that another source handed the
```
baseTwo: Nit: ragged wrap. This line is 34 characters and L92 is 71, while the rest of this paragraph - and the whole file - sits at 98-104. Re-flow L92-L98 as one paragraph at the file's width; the orphan reads as a leftover from an edit rather than a deliberate break.

Cql/CoreTests/ValueSetSourceFacadeMemoTests.cs
```diff
@@ -162,26 +163,43 @@ public async Task ComputedExpansion_StaysPrivateToItsSource()
...
+        var resolver = new InMemoryResourceResolver(composed);
+        var changedProperties = new List<string>();
+        composed.PropertyChanged += (_, e) => changedProperties.Add(e.PropertyName ?? "?");
+
+        await Assert.ThrowsExceptionAsync<ValueSetUnknownException>(() => new ValueSetSource(resolver).Add(composed));
 
-        Assert.AreSame(facadeB, facadeC);
+        Assert.IsFalse(composed.HasExpansion, "a failed expansion must not alter the caller's instance.");
+        Assert.AreEqual(0, changedProperties.Count,
+            $"a failed expansion must not write to the caller's instance at all, but these properties changed: {string.Join(", ", changedProperties)}");
```
baseTwo, line 202: **Required: nothing pins the residual these two tests deliberately stop short of.**

Both the class remarks (L90-93) and the shipped release note assert that an expansion-less *included* valueset is still expanded in place. No test holds either statement to account. When firely-net-sdk#3582 lands, or `FirelyNetVersion` moves off 6.3.0, both go stale silently - and the release note keeps warning consumers about a hazard that no longer exists.

`CLAUDE.md` asks for exactly this case to be quarantined with a comment pointing at the tracking issue, so please add a characterization test named as pinning a *known upstream defect*. Verified to pass on the branch as it stands:

```csharp
[TestMethod]
public async Task IncludedExpansionlessValueSet_IsStillExpandedInPlace_ByTheExpander()
{
    // Quarantine, not an invariant we want: the copy protects the instance handed to Add, but the
    // expander resolves compose.include dependencies itself and expands an expansion-less one in
    // place. Tracked upstream at FirelyTeam/firely-net-sdk#3582 (adjacent defect B). When that is
    // fixed - or FirelyNetVersion moves - this test fails, which is the prompt to update the class
    // remarks and the release note that both currently document this residual.
    var leaf = ExpandedValueSet("http://example.org/ValueSet/residual-leaf", "111");
    var inner = ComposedValueSet("http://example.org/ValueSet/residual-inner", leaf.Url);
    var outer = ComposedValueSet("http://example.org/ValueSet/residual-outer", inner.Url);
    var resolver = new InMemoryResourceResolver(leaf, inner, outer);

    var facade = await new ValueSetSource(resolver).Add(outer);

    Assert.IsFalse(outer.HasExpansion, "the instance handed to Add is protected by the copy.");
    Assert.IsTrue(inner.HasExpansion, "known residual: the expander expands an included valueset in place.");
    Assert.IsTrue(facade.IsCodeInValueSet("111", CodeSystem));
}
```

If you take the copy-on-resolve decorator from the other comment instead, this flips to `Assert.IsFalse(inner.HasExpansion, ...)` and stops being a quarantine - either way the documented behavior ends up pinned rather than asserted only in prose.

baseTwo:
@copilot Please apply all six changes below to this branch (`1591-valuesetsource-expand-copy`, head `a041f551`). Each is stated in full — file, exact before/after, and why — so nothing needs to be looked up in the review threads. They come from [this review](https://github.com/FirelyTeam/firely-cql-sdk/pull/1592#pullrequestreview-5027743233).

Items 1 and 3 both edit `docs/releases/vnext/1592-valuesetsource-no-mutation.md`; apply them together so the file ends up as shown at the end of item 3.

---

## 1. Add a `## Potentially Breaking` section to the release-note fragment

**File:** `docs/releases/vnext/1592-valuesetsource-no-mutation.md`

**Why:** the fragment is `## Fixes` only, which gives the release cut nothing to set the version level with. The cross-source consequence is not just lost reuse, it is an *answer change*: previously source A (resolver RA) expanded `V` in place, so source B — with a **different** resolver RB — called `Add(V)`, saw `HasExpansion`, and answered from A's expansion without ever consulting RB. Now B expands its own copy against RB, so the two sources give different membership answers. [`docs/versioning.md`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/versioning.md) lists "CQL evaluation results change — including bug fixes that correct previously wrong answers" as a MESO trigger, and its tie-breaker says to go up when the level is unclear. Second host-visible change: a host that read `vs.Expansion` back after `Add`/`Load` — behaviour the class docs called "the one edit to a resolved valueset the SDK performs on its own" — no longer finds one.

**Change:** insert this section **above** the existing `## Fixes` heading, so the file starts with it:

```md
## Potentially Breaking

- `ValueSetSource` no longer writes a computed expansion into the valueset handed to `Add` or
  resolved by `Load`. Two consequences for hosts: a host that read `ValueSet.Expansion` back after
  `Add`/`Load` no longer finds one there and must expand for itself; and two sources with
  *different* resolvers handed the same expansion-less instance no longer share the first source's
  expansion — each now answers from its own resolver's world, so membership answers can move (to
  the correct ones). (#1591)
```

Keep the existing `## Fixes` section and its bullet as they are apart from the edits in items 2 and 3.

---

## 2. Make the upstream citation survive a ticket split — two places

`FirelyTeam/firely-net-sdk#3582` is titled *"Terminology stack issues concurrent calls to the caller-supplied resolver via internal `Task.WhenAll`"*. The in-place mutation we are citing it for is **"Adjacent defect B"** inside that issue, and its body ends with *"Happy to split adjacent defect B (in-place mutation of resolver-owned instances) into its own issue if you prefer to track it separately."* So a consumer clicking through from a shipped release note lands on a title about something else, and the reference breaks outright if that split happens.

**2a. File:** `docs/releases/vnext/1592-valuesetsource-no-mutation.md`, line 10.

Before:
```md
  [firely-net-sdk#3582](https://github.com/FirelyTeam/firely-net-sdk/issues/3582). A source still
```
After:
```md
  [firely-net-sdk#3582 (adjacent defect B)](https://github.com/FirelyTeam/firely-net-sdk/issues/3582).
  A source still
```
(re-wrap the rest of that bullet to the file's width as needed).

**2b. File:** `Cql/Cql.Firely/ValueSetSource.cs`, in the class `<remarks>` — currently lines 92-93:

Before:
```csharp
/// valueset in place - that is the expander's own behavior, tracked at
/// <see href="https://github.com/FirelyTeam/firely-net-sdk/issues/3582"/>. The price of the copy is
```
After: the same sentence, but with the link labelled so the reader knows which part of that issue is meant — e.g. `tracked upstream as adjacent defect B of <see href="https://github.com/FirelyTeam/firely-net-sdk/issues/3582"/>`. Fold this into the re-flow in item 4 rather than making it a separate edit.

---

## 3. Say what the residual costs the host the fix is for

**File:** `docs/releases/vnext/1592-valuesetsource-no-mutation.md`, the `## Fixes` bullet.

**Why:** the note already says the expander still expands an expansion-less *included* valueset in place. What it does not say is the consequence: a host-side conformance cache is **still corrupted through that path** — the same corruption the first half of the bullet says this change fixes. Verified on this branch: outer → include(inner, expansion-less) → include(leaf, expanded), and after `Add(outer)` the caller's `inner.HasExpansion` is `true` with one `Expansion` PropertyChanged event. A consumer reading "so that instance … is left untouched" should not have to infer that a *different* cached instance is not.

**Change:** extend the sentence about the expander so it states the cost. After the `(adjacent defect B)` citation from item 2a, add:

```md
  A host cache serving an expansion-less valueset that another valueset includes is therefore still
  written to through that path, including having its expansion cleared if the expansion fails.
```

The whole fragment should end up reading in this order: `## Potentially Breaking` (item 1), then `## Fixes` with the corrected citation (item 2a) and this added sentence.

**Not part of this ask:** the review also sketched a copy-on-resolve resolver decorator that closes the residual entirely. That is a design decision for @alexzautke to rule on — please do **not** implement it. If it is later adopted, this disclosure sentence comes back out.

---

## 4. Re-flow the ragged paragraph in the class remarks

**File:** `Cql/Cql.Firely/ValueSetSource.cs`, lines 92-95 inside the class `<remarks>`.

**Why:** line 92 is 71 characters and line 94 is 34 (`/// that another source handed the`), while the rest of that paragraph — and the whole file — wraps at 98-104. The orphan reads as a leftover from an edit rather than a deliberate break.

**Change:** re-flow lines 82-98 as one paragraph at the file's existing width (~100 columns), preserving the wording exactly except for the label added in item 2b. No sentence should change meaning; this is purely a re-wrap.

---

## 5. Pin the residual with a quarantine test

**File:** `Cql/CoreTests/ValueSetSourceFacadeMemoTests.cs`

**Why:** both the class remarks (L90-93) and the shipped release note assert that an expansion-less *included* valueset is still expanded in place, and no test holds either statement to account. When `FirelyTeam/firely-net-sdk#3582` lands, or `FirelyNetVersion` moves off `6.3.0`, both go stale silently — and the release note keeps warning consumers about a hazard that no longer exists. `CLAUDE.md` asks for exactly this case to be quarantined with a comment pointing at the tracking issue.

**Change:** add this test method after `FailedExpansion_LeavesTheResolvedInstanceUntouched`. It is verified to pass on this branch as it stands:

```csharp
    [TestMethod]
    public async Task IncludedExpansionlessValueSet_IsStillExpandedInPlace_ByTheExpander()
    {
        // Quarantine, not an invariant we want: the copy protects the instance handed to Add, but the
        // expander resolves compose.include dependencies itself and expands an expansion-less one in
        // place. Tracked upstream as adjacent defect B of FirelyTeam/firely-net-sdk#3582. When that is
        // fixed - or FirelyNetVersion moves - this test fails, which is the prompt to update the class
        // remarks and the release note that both currently document this residual.
        var leaf = ExpandedValueSet("http://example.org/ValueSet/residual-leaf", "111");
        var inner = ComposedValueSet("http://example.org/ValueSet/residual-inner", leaf.Url);
        var outer = ComposedValueSet("http://example.org/ValueSet/residual-outer", inner.Url);
        var resolver = new InMemoryResourceResolver(leaf, inner, outer);

        var facade = await new ValueSetSource(resolver).Add(outer);

        Assert.IsFalse(outer.HasExpansion, "the instance handed to Add is protected by the copy.");
        Assert.IsTrue(inner.HasExpansion, "known residual: the expander expands an included valueset in place.");
        Assert.IsTrue(facade.IsCodeInValueSet("111", CodeSystem));
    }
```

`ExpandedValueSet`, `ComposedValueSet` and `CodeSystem` are the existing private helpers in that class; no new `using` is needed.

---

## 6. State the no-mutation contract on `Add` and `Load` themselves

**File:** `Cql/Cql.Firely/ValueSetSource.cs`

**Why:** the guarantee is now the single most consumer-relevant fact about `Add`, and it lives only as a paragraph inside an 80-line class `<remarks>`. A consumer checking whether their cached instance is safe should see it in IntelliSense.

**6a.** `Add(ValueSet)` — currently lines 194-196:

Before:
```csharp
    /// <summary>
    /// Adds a <see cref="ValueSet"/> to the cache, so it will not be retrieved using the resolver.
    /// </summary>
```
After:
```csharp
    /// <summary>
    /// Adds a <see cref="ValueSet"/> to the cache, so it will not be retrieved using the resolver.
    /// The instance is not modified: an expansion this source has to compute is computed on a private
    /// copy. See the remarks on this class.
    /// </summary>
```

**6b.** `Load(string)` — currently lines 308-311:

Before:
```csharp
    /// <summary>
    /// Given a canonical, returns the <see cref="IValueSetFacade"/> for that canonical from the
    /// cache, or uses the resolver to load it.
    /// </summary>
```
After:
```csharp
    /// <summary>
    /// Given a canonical, returns the <see cref="IValueSetFacade"/> for that canonical from the
    /// cache, or uses the resolver to load it. The resolved instance is not modified; see the remarks
    /// on this class.
    /// </summary>
```

---

## Before you push

- `dotnet test Cql/CoreTests/CoreTests.csproj --filter "FullyQualifiedName~ValueSetSourceFacadeMemoTests"` must pass — 13 tests once item 5 is in.
- Build `Cql-Sdk.slnf` in **`Release`** (`dotnet build Cql-Sdk.slnf -c Release`), not `Debug`: the PublicAPI analyzer's RS0016/RS0017 are suppressed in `Debug`, so a green `Debug` build proves nothing about CI.
- Do not touch `Cql/Cql.Firely/ValueSetSource.cs`'s existing copyright header, and do not change the `DeepCopy` fix itself — items 1-6 are documentation, one release-note section, and one test.

Then please reply on this PR listing what you changed per item, so the review threads can be resolved against a commit SHA.

docs/releases/vnext/1592-valuesetsource-no-mutation.md
baseTwo: Addressed in e1ab98d75: the fragment now opens with a `## Potentially Breaking` section stating both host-visible consequences - a host that read `ValueSet.Expansion` back after `Add`/`Load` no longer finds one, and two sources with different resolvers no longer share the first source's expansion, so membership answers can move. The release cut now has the trigger it needs to set the level under `versioning.md` rather than inferring it from a `## Fixes`-only fragment.

Also flagged separately: I have added the `breaking change` label and ticked that checkbox in the PR body, since the fragment now declares a potentially-breaking section.

Resolving.

docs/releases/vnext/1592-valuesetsource-no-mutation.md
baseTwo: Addressed in e1ab98d75: both places now name which part of the upstream issue is meant - the fragment reads `firely-net-sdk#3582 (adjacent defect B)` and the class remarks read "tracked upstream as adjacent defect B of ...". If that defect is ever split into its own ticket, the label makes the stale reference obvious instead of silently pointing at an issue about concurrent resolver calls.

Resolving.

Cql/Cql.Firely/ValueSetSource.cs
baseTwo: Partly addressed in e1ab98d75, and deliberately left open for your ruling.

**Done:** the release-note fragment now states what the residual costs the host the fix is for - "A host cache serving an expansion-less valueset that another valueset includes is therefore still written to through that path, including having its expansion cleared if the expansion fails." A consumer no longer has to infer that from the sentence about the expander's behaviour.

**Not done, and not asked for:** the copy-on-resolve decorator. I explicitly told @copilot not to implement it, because whether to close the residual here or wait for upstream is your call, and the trade-off I listed (a deep copy per resolved dependency, and `ToValueSetDictionaryAsync` re-expanding shared includes for grouping-heavy content) is unmeasured. The 8-line decorator in my comment above is verified to close the hole if you want it; if you take it, the disclosure sentence added in e1ab98d75 comes back out.

Leaving this thread open so the decision is visible rather than buried in a resolved thread. Nothing blocks the merge on it - the residual is disclosed and, as of e1ab98d75, pinned by a test.

Cql/Cql.Firely/ValueSetSource.cs
baseTwo: Addressed in e1ab98d75: the paragraph is re-flowed and the 34-character orphan is gone. Verified - lines 82-98 now run 94-103 characters with the tail at 52-56, matching the rest of the file, and the wording is unchanged apart from the `adjacent defect B` label from the other thread.

Resolving.

Cql/CoreTests/ValueSetSourceFacadeMemoTests.cs
baseTwo: Addressed in e1ab98d75: `IncludedExpansionlessValueSet_IsStillExpandedInPlace_ByTheExpander` is in, with the comment naming `FirelyTeam/firely-net-sdk#3582` as the tracking issue and saying in as many words that this is a quarantine rather than an invariant we want. @copilot added `!` to the `leaf.Url` / `inner.Url` arguments, which is right - `Url` is `string?`.

Verified locally at e1ab98d75: `ValueSetSourceFacadeMemoTests` is 13/13 on `net10.0`. Both statements that documented the residual - the class remarks and the release note - are now pinned, so the upstream fix landing will fail this test instead of silently staling the docs.

Resolving.

baseTwo (commented):
## Review of `e1ab98d75`

All six items from [the earlier ask](https://github.com/FirelyTeam/firely-cql-sdk/pull/1592#issuecomment-5438888663) landed, and I verified each rather than reading the summary:

- **Re-flow is faithful.** Diffed the paragraph word by word: wording is identical to `a041f551` apart from the `adjacent defect B` label, the 34-character orphan is gone, and lines 82-98 now run 94-103 characters against a file that sits at 98-104.
- **The quarantine test passes and discriminates.** `ValueSetSourceFacadeMemoTests` is 13/13 on `net10.0` at this head. It asserts `inner.HasExpansion` is `true`, so the upstream fix landing will fail it rather than silently staling the docs — which is the point. The `!` added to `leaf.Url` / `inner.Url` is correct, `Url` is `string?`.
- **No flake risk from adding to that class.** Checked for `[assembly: Parallelize]` in `CoreTests` — there is none, so MSTest runs the class sequentially and the new test cannot perturb `ConcurrentSources_ConvergeOnOneFacade`'s `BuildFromExpansionCount` delta (which snapshots a baseline anyway).
- **The heading spelling makes the trigger fire mechanically.** `## Potentially Breaking` is one of the two spellings [`versioning.md`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/versioning.md) names under MESO, so the release cut reads it without interpretation. Fragment filename also matches the `<PR-number>-<slug>.md` convention.
- Nothing else was touched: three files, no header edits, and the `DeepCopy` fix itself is unchanged.

Two documentation findings inline. Neither is a code defect and neither blocks the merge; the first is a mess my own ask created, so I would rather fix it here than leave it for whoever cuts the release.

For the record on thread state: the `compose.include` design thread is the one I left open, because the copy-on-resolve decorator is still @alexzautke's ruling to make. Every other thread on this PR is replied to and resolved against a commit SHA.

docs/releases/vnext/1592-valuesetsource-no-mutation.md
```diff
@@ -0,0 +1,24 @@
...
+- `ValueSetSource` no longer writes a computed expansion into the valueset handed to `Add` or
+  resolved by `Load`. Two consequences for hosts: a host that read `ValueSet.Expansion` back after
+  `Add`/`Load` no longer finds one there and must expand for itself; and two sources with
+  *different* resolvers handed the same expansion-less instance no longer share the first source's
+  expansion — each now answers from its own resolver's world, so membership answers can move (to
+  the correct ones). (#1591)
+
+## Fixes
+
+- `ValueSetSource` no longer writes into the valueset handed to `Add` or resolved by `Load`: a
```
baseTwo: **The two bullets now restate the same change — and that is my ask's fault, not @copilot's.**

I asked for a `## Potentially Breaking` section without saying what should then move out of `## Fixes`, so both bullets now open with almost the same clause:

- L3: "`ValueSetSource` no longer writes a computed expansion into the valueset handed to `Add` or resolved by `Load`."
- L12: "`ValueSetSource` no longer writes into the valueset handed to `Add` or resolved by `Load`: ..."

When the release cut folds this fragment in, the versioned notes carry that sentence twice, in two different sections — exactly the kind of tangle [`cut-release-notes`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/.claude/skills/cut-release-notes/SKILL.md) then has to unpick, and the template asks for concision. Worse, the `## Fixes` bullet as it stands opens by claiming host-cache corruption is fixed and closes by saying host caches are still written to, which reads as a retraction rather than a scope statement.

Cleanest split: let `## Potentially Breaking` own *what changed for you*, and let `## Fixes` own *the defect, its scope, and what is left*. Concretely, replace the `## Fixes` bullet (L12-24) with:

```md
## Fixes

- A valueset that arrives without an expansion is now expanded on a private copy, so a host cache
  serving that instance to every consumer in the process is no longer corrupted by the SDK: the
  computed expansion used to be written into the caller's instance, and a failed expansion used to
  clear it. What is protected is the instance handed to `Add` or resolved by `Load`, and only that:
  the underlying `ValueSetExpander` still expands an expansion-less *included* valueset
  (`compose.include`) resolved from this source's resolver in place — the expander's own behavior,
  tracked as
  [firely-net-sdk#3582 (adjacent defect B)](https://github.com/FirelyTeam/firely-net-sdk/issues/3582)
  — so a cache serving such an included valueset is still written to, and still has its expansion
  cleared if that expansion fails. A source expands a given canonical at most once through its
  per-canonical facade layer; hosts wanting cross-source reuse of computed expansions can serve
  valuesets with static expansions or seed sources via `Add(string, IEnumerable<CqlCode>)`. (#1591)
```

Leave the `## Potentially Breaking` section (L1-8) exactly as it is — it says the right thing.

Cql/Cql.Firely/ValueSetSource.cs
```diff
@@ -81,12 +81,21 @@ namespace Hl7.Cql.Fhir;
...
+/// not determine - so its facade stays in the per-source layer only. The expansion is computed on a
+/// private copy of the valueset: the instance handed in is on loan - on the <see cref="Load"/> path it
+/// belongs to the resolver, which may be serving the same object to every consumer in the process -
+/// and the underlying expander writes into its argument (and clears the expansion on failure), so
+/// running it on the original would edit, or on failure damage, an object this class does not own.
+/// This class therefore never modifies the valueset handed to <see cref="Add(ValueSet)"/> or resolved
+/// by <see cref="Load"/>. The copy does not reach further: the expander pulls <c>compose.include</c>
+/// dependencies from the resolver on its own and still expands an expansion-less <em>included</em>
+/// valueset in place - that is the expander's own behavior, tracked upstream as adjacent defect B of
+/// <see href="https://github.com/FirelyTeam/firely-net-sdk/issues/3582"/>. The price of the copy is
```
baseTwo: Minor, for symmetry with the release note: these remarks give the residual's **mechanism** but not its **consequence**, while the fragment now gives both.

A maintainer reading this class learns that the expander "still expands an expansion-less *included* valueset in place" and has to work out for themselves that this means a shared host cache is still being corrupted — which was the whole point of the original finding. The release note should not be the only place that says so.

Suggest one sentence after the `adjacent defect B` citation, before "The price of the copy is":

```
/// A cache serving an expansion-less valueset that another valueset includes is therefore still
/// written to through that path, and still loses its expansion if that expansion fails.
```

Re-flow the paragraph to the file's width afterwards, as before.

baseTwo:
@copilot Two follow-up edits, both documentation, from [this review](https://github.com/FirelyTeam/firely-cql-sdk/pull/1592#pullrequestreview-5040799350) of your commit `e1ab98d75`. Everything from the previous ask verified correct — these are consequences of how I worded that ask, not mistakes in what you applied.

---

## 1. De-duplicate the release-note fragment's two bullets

**File:** `docs/releases/vnext/1592-valuesetsource-no-mutation.md`

**Why:** the previous ask added `## Potentially Breaking` without saying what should move out of `## Fixes`, so both bullets now open with nearly the same clause — L3 "`ValueSetSource` no longer writes a computed expansion into the valueset handed to `Add` or resolved by `Load`" and L12 "`ValueSetSource` no longer writes into the valueset handed to `Add` or resolved by `Load`". When the release cut folds this fragment into the versioned notes, that sentence appears twice in two sections. The `## Fixes` bullet also currently opens by claiming host-cache corruption is fixed and closes by saying host caches are still written to, which reads as a retraction rather than a scope statement.

**Change:** replace the entire `## Fixes` section (lines 10-24, i.e. the heading and its single bullet) with this. Leave the `## Potentially Breaking` section on lines 1-8 **completely unchanged**:

```md
## Fixes

- A valueset that arrives without an expansion is now expanded on a private copy, so a host cache
  serving that instance to every consumer in the process is no longer corrupted by the SDK: the
  computed expansion used to be written into the caller's instance, and a failed expansion used to
  clear it. What is protected is the instance handed to `Add` or resolved by `Load`, and only that:
  the underlying `ValueSetExpander` still expands an expansion-less *included* valueset
  (`compose.include`) resolved from this source's resolver in place — the expander's own behavior,
  tracked as
  [firely-net-sdk#3582 (adjacent defect B)](https://github.com/FirelyTeam/firely-net-sdk/issues/3582)
  — so a cache serving such an included valueset is still written to, and still has its expansion
  cleared if that expansion fails. A source expands a given canonical at most once through its
  per-canonical facade layer; hosts wanting cross-source reuse of computed expansions can serve
  valuesets with static expansions or seed sources via `Add(string, IEnumerable<CqlCode>)`. (#1591)
```

The em dashes and the `(#1591)` suffix are intentional; keep them. Lines should stay at or under ~103 characters, matching the rest of the file.

---

## 2. State the residual's consequence in the class remarks too

**File:** `Cql/Cql.Firely/ValueSetSource.cs`, the class `<remarks>` paragraph (currently lines 82-98).

**Why:** the remarks give the residual's *mechanism* — "still expands an expansion-less *included* valueset in place" — but not its *consequence*. A maintainer reading the class has to work out for themselves that this means a shared host cache is still being corrupted, which was the entire point of the finding. The release-note fragment now states both; the class that owns the behaviour should not be the vaguer of the two.

**Change:** insert this sentence immediately after the `<see href="https://github.com/FirelyTeam/firely-net-sdk/issues/3582"/>.` citation and immediately before the existing sentence that begins "The price of the copy is":

```
/// A cache serving an expansion-less valueset that another valueset includes is therefore still
/// written to through that path, and still loses its expansion if that expansion fails.
```

Then re-flow the whole paragraph to the file's existing width (~100 columns, nothing over 104) as you did in `e1ab98d75`. Do not change any other wording in that paragraph.

---

## Before you push

- `dotnet test Cql/CoreTests/CoreTests.csproj --filter "FullyQualifiedName~ValueSetSourceFacadeMemoTests"` must still pass at 13/13.
- Build `Cql-Sdk.slnf` in **`Release`** (`dotnet build Cql-Sdk.slnf -c Release`), not `Debug` — RS0016/RS0017 from the PublicAPI analyzer are suppressed in `Debug`.
- Item 2 edits a `.cs` file, so the full CI pipeline will run rather than being skipped by the docs-only `ignorePatterns` filter. That is expected.
- Do not touch the copyright header, the `DeepCopy` fix, the quarantine test, or the XML summaries on `Add`/`Load` — all four are verified correct and final.

Then reply on this PR saying what you changed per item, so both threads can be resolved against the new SHA.

baseTwo:
@copilot Two follow-up edits, both documentation, from [this review](https://github.com/FirelyTeam/firely-cql-sdk/pull/1592#pullrequestreview-5040799350) of your commit `e1ab98d75`. Everything from the previous ask verified correct — these are consequences of how I worded that ask, not mistakes in what you applied.

---

## 1. De-duplicate the release-note fragment's two bullets

**File:** `docs/releases/vnext/1592-valuesetsource-no-mutation.md`

**Why:** the previous ask added `## Potentially Breaking` without saying what should move out of `## Fixes`, so both bullets now open with nearly the same clause — L3 "`ValueSetSource` no longer writes a computed expansion into the valueset handed to `Add` or resolved by `Load`" and L12 "`ValueSetSource` no longer writes into the valueset handed to `Add` or resolved by `Load`". When the release cut folds this fragment into the versioned notes, that sentence appears twice in two sections. The `## Fixes` bullet also currently opens by claiming host-cache corruption is fixed and closes by saying host caches are still written to, which reads as a retraction rather than a scope statement.

**Change:** replace the entire `## Fixes` section (lines 10-24, i.e. the heading and its single bullet) with this. Leave the `## Potentially Breaking` section on lines 1-8 **completely unchanged**:

```md
## Fixes

- A valueset that arrives without an expansion is now expanded on a private copy, so a host cache
  serving that instance to every consumer in the process is no longer corrupted by the SDK: the
  computed expansion used to be written into the caller's instance, and a failed expansion used to
  clear it. What is protected is the instance handed to `Add` or resolved by `Load`, and only that:
  the underlying `ValueSetExpander` still expands an expansion-less *included* valueset
  (`compose.include`) resolved from this source's resolver in place — the expander's own behavior,
  tracked as
  [firely-net-sdk#3582 (adjacent defect B)](https://github.com/FirelyTeam/firely-net-sdk/issues/3582)
  — so a cache serving such an included valueset is still written to, and still has its expansion
  cleared if that expansion fails. A source expands a given canonical at most once through its
  per-canonical facade layer; hosts wanting cross-source reuse of computed expansions can serve
  valuesets with static expansions or seed sources via `Add(string, IEnumerable<CqlCode>)`. (#1591)
```

The em dashes and the `(#1591)` suffix are intentional; keep them. Lines should stay at or under ~103 characters, matching the rest of the file.

---

## 2. State the residual's consequence in the class remarks too

**File:** `Cql/Cql.Firely/ValueSetSource.cs`, the class `<remarks>` paragraph (currently lines 82-98).

**Why:** the remarks give the residual's *mechanism* — "still expands an expansion-less *included* valueset in place" — but not its *consequence*. A maintainer reading the class has to work out for themselves that this means a shared host cache is still being corrupted, which was the entire point of the finding. The release-note fragment now states both; the class that owns the behaviour should not be the vaguer of the two.

**Change:** insert this sentence immediately after the `<see href="https://github.com/FirelyTeam/firely-net-sdk/issues/3582"/>.` citation and immediately before the existing sentence that begins "The price of the copy is":

```
/// A cache serving an expansion-less valueset that another valueset includes is therefore still
/// written to through that path, and still loses its expansion if that expansion fails.
```

Then re-flow the whole paragraph to the file's existing width (~100 columns, nothing over 104) as you did in `e1ab98d75`. Do not change any other wording in that paragraph.

---

## Before you push

- `dotnet test Cql/CoreTests/CoreTests.csproj --filter "FullyQualifiedName~ValueSetSourceFacadeMemoTests"` must still pass at 13/13.
- Build `Cql-Sdk.slnf` in **`Release`** (`dotnet build Cql-Sdk.slnf -c Release`), not `Debug` — RS0016/RS0017 from the PublicAPI analyzer are suppressed in `Debug`.
- Item 2 edits a `.cs` file, so the full CI pipeline will run rather than being skipped by the docs-only `ignorePatterns` filter. That is expected.
- Do not touch the copyright header, the `DeepCopy` fix, the quarantine test, or the XML summaries on `Add`/`Load` — all four are verified correct and final.

Then reply on this PR saying what you changed per item, so both threads can be resolved against the new SHA.

Cql/Cql.Firely/ValueSetSource.cs
baseTwo: Closing this thread, with the decision recorded rather than lost.

**Applied:** the disclosure half. `e1ab98d75` added the residual's cost to the release-note fragment, a follow-up tightens that wording, and the class remarks are getting the same sentence so the mechanism and its consequence sit together. `IncludedExpansionlessValueSet_IsStillExpandedInPlace_ByTheExpander` pins the behaviour, so the upstream fix landing fails a test instead of quietly staling two documents.

**Declined for now:** the copy-on-resolve decorator. It is verified to close the residual entirely (8 lines, `inner.HasExpansion=False`, zero PropertyChanged events, membership unchanged), and it remains available whenever you want it — the trade-off is a deep copy per resolved dependency on the cold expansion path, plus `ToValueSetDictionaryAsync` re-expanding shared includes for grouping-heavy content, which is unmeasured. If it is adopted later, the disclosure sentences in the fragment and the remarks come back out and the quarantine test flips to `Assert.IsFalse(inner.HasExpansion, ...)`.

Resolving so the thread list is clean, not because the question is settled: the residual is disclosed to consumers, pinned by a test, documented in the class, and tracked upstream at FirelyTeam/firely-net-sdk#3582 (adjacent defect B). Nothing about it blocks this merge.

docs/releases/vnext/1592-valuesetsource-no-mutation.md
baseTwo: Addressed in 6c21e1532. The `## Fixes` bullet now leads with the defect and its consequence ("A valueset that arrives without an expansion is now expanded on a private copy, so a host cache serving that instance ... is no longer corrupted by the SDK"), then states the scope, then the residual - so the duplicated opening clause is gone and the section no longer reads as claiming a fix and retracting it. `## Potentially Breaking` is untouched, as asked.

Verified at that head: the two bullets' opening sentences are now distinct, the file's widest line is 101 characters, `Cql.Firely` builds clean in `Release`, and `ValueSetSourceFacadeMemoTests` is 13/13 on `net10.0`.

Two deliberate wording variances from my suggested text, both improvements: "still loses its expansion if that expansion fails" instead of "still has its expansion cleared", which now matches the sentence added to the class remarks; and "A source still expands a given canonical at most once", keeping the original's "still".

Resolving.

Cql/Cql.Firely/ValueSetSource.cs
baseTwo: Addressed in 6c21e1532: the remarks now carry the consequence as well as the mechanism - "A cache serving an expansion-less valueset that another valueset includes is therefore still written to through that path, and still loses its expansion if that expansion fails." - inserted after the upstream citation and before "The price of the copy is", with the paragraph re-flowed.

Verified at that head: lines 82-99 run 90-103 characters against a file that sits at 98-104, no orphan lines, and the wording elsewhere in the paragraph is unchanged. The phrasing matches the release-note fragment's, so the two documents now say the same thing in the same words.

Resolving.

baseTwo (approved):

PR #1596, opened 2026-09-02
Expand included valuesets on private copies

### LINKED ISSUE #1597: ValueSetExpander mutates resolver-owned valuesets reached through compose.include
## Summary

#1592 stopped `ValueSetSource` from mutating the valueset handed to `Add(ValueSet)` or resolved by `Load(canonical)`: the expansion is computed on a `DeepCopy`. Its review verified that the same hazard remains reachable one include-level down. Expansion is recursive — the FHIR SDK's `ValueSetExpander` resolves `compose.include` dependencies through `ValueSetExpanderSettings.ValueSetSource`, which `ValueSetSource.BuildExpander()` populates with the caller-supplied resolver — and an expansion-less *included* valueset resolved that way is expanded **in place, on the instance the resolver returned** (the expander's own behavior, adjacent defect B of [firely-net-sdk#3582](https://github.com/FirelyTeam/firely-net-sdk/issues/3582)).

## Impact

For any host whose resolver serves shared instances (a server-side conformance cache such as Firely Server's), this is the same corruption #1591 describes, one level down:

- a cached, expansion-less valueset that some other valueset includes is written to during the fill (half-built expansion observable by unrelated consumers);
- a **failed** expansion clears the included instance's expansion for every consumer until cache eviction or restart.

## Verified repro

outer → `compose.include`(inner, expansion-less) → `compose.include`(leaf, expanded). After `Add(outer)`, `inner.HasExpansion == true` with a `PropertyChanged` event on the caller's inner instance. (This was pinned by the quarantine test `IncludedExpansionlessValueSet_IsStillExpandedInPlace_ByTheExpander`.)

## Expected behavior

The no-mutation contract of #1591/#1592 covers every valueset obtained through the caller's resolver, transitively resolved includes included: the expander's recursive writes land on private copies, and the resolver-owned instances stay byte-for-byte untouched on success and failure alike.

## Fix

A copy-on-resolve decorator installed only into the resolver `BuildExpander()` hands the expander: every resolved `ValueSet` is returned as a private `DeepCopy` (both interface dispatch paths forwarded; `CodeSystem` resources deliberately pass through uncopied). Accepted trade-off: a canonical reached only as another valueset's include is expanded once per outer valueset instead of being found already expanded — that incidental reuse was the mutation itself; hosts wanting reuse ship static expansions or seed via `Add(string, IEnumerable<CqlCode>)`.

Implemented in PR #1596. The decorator becomes removable if the upstream expander stops mutating resolved instances (firely-net-sdk#3582).

alexzautke:
## Description

**Primary Work**

PR #1592 stopped `ValueSetSource` from mutating the valueset handed to `Add` or resolved by `Load` — the expansion is computed on a `DeepCopy`. Its review verified a residual one include-level down, now tracked as #1597: the `ValueSetExpander` resolves `compose.include` dependencies through the resolver this class configures (`ValueSetExpanderSettings.ValueSetSource`, set in `BuildExpander()`), and expands an expansion-less *included* valueset in place — writing the expansion into, and on failure clearing it from, the very instance the resolver returned. On the `Load` path that instance is on loan from the caller's resolver, which a host conformance cache (e.g. Firely Server's) may serve to every consumer in the process, so the include path carried the same corruption the top-level copy prevents (adjacent defect B of [firely-net-sdk#3582](https://github.com/FirelyTeam/firely-net-sdk/issues/3582)).

`BuildExpander()` now hands the expander a `CopyOnResolve` decorator around the resolver. It copies **exactly what the expander writes to, and nothing else**: a resolved `ValueSet` that has no expansion yet is returned as a private copy, so the expander's recursive in-place writes land on that copy. A resolved valueset that already carries an expansion passes through as the resolver's own instance — the expander only reads it, and the expansion is the bulk of a valueset, so copying it to read it once would cost far more than the compose-only copies the decorator does make. `CodeSystem` resources pass through uncopied for the same reason. The copies are confined to the cold path: the expander runs only for a valueset that misses the per-canonical facade layer and arrives without an expansion, and content that ships static expansions never reaches it. The decorator deliberately keeps no memo of its own, so a canonical reached solely as another valueset's include is expanded anew for every resolver lookup that pulls it in; hosts wanting reuse ship static expansions or seed via `Add(string, IEnumerable)`.

The decorator forwards **both dispatch paths** of `IAsyncResourceResolver` — the modern `TryResolve*Async` members and the obsolete ones — to the wrapped resolver's own implementation of that same path, since the `TryResolve*` default implementations fall back to the obsolete members of the instance they run on; a resolver supporting only the modern members keeps working.

The not-yet-released #1592 release-note fragment no longer describes the residual (both fragments ship in the same release, so it must not claim a gap this PR closes), and a new fragment records the fix, the narrowed copy rule and its cost bound, together with its `Potentially Breaking` consequence: a host reading a computed expansion back from a resolver-owned *included* instance no longer finds one, mirroring #1592's entry for the top-level instance. The decorator becomes removable if the upstream expander stops mutating resolved instances (firely-net-sdk#3582).

**Auxiliary Work**

None.

## Related issues

Fixes #1597 — the transitive residual of #1591, verified in the review of #1592.

## Testing

- New `TransitivelyIncludedValueSet_LeavesTheResolvedInstanceUntouched`: outer → expansion-less inner → expanded leaf; after `Add(outer)` the inner instance stays `!HasExpansion` with zero `PropertyChanged` events, while the facade answers membership correctly. Fails on unfixed develop (`inner.HasExpansion == true`).
- New `FailedTransitiveExpansion_LeavesTheIncludedInstanceUntouched`: a failing include chain throws without altering the included instance. Fails on unfixed develop (two `Expansion` change events: write, then clear).
- New `TransitiveResolution_ReachesTheResolverThroughItsModernMembers`: the include chain resolves through a resolver whose obsolete members throw, pinning that the decorator forwards the modern `TryResolve*` path. Fails against a decorator implementing only the obsolete members.
- New `ResolvedValueSets_AreCopiedOnlyWhenTheyLackAnExpansion`: a `ValueSet` subclass overriding `DeepCopyInternal` counts its copies; the expansion-less include is copied exactly once and the already-expanded include not at all, with the included instance untouched and membership correct. Fails against the copy-everything decorator (leaf copied once).
- The quarantine test `IncludedExpansionlessValueSet_IsStillExpandedInPlace_ByTheExpander`, which pinned the residual and whose comment named this exact change as its replacement trigger, is replaced by its inverse.
- Full `CoreTests` suite: 1378 passed, 0 failed, 8 pre-existing skips (net10.0, Release).

## FirelyTeam Checklist

https://claude.ai/code/session_01ESiK1197pEWExuZJJGsBAd


alexzautke:
All four Copilot findings addressed in db794d9; each was verified before fixing:

1. **Dispatch through the obsolete members** — confirmed by reflection (the `TryResolve*Async` members are default implementations falling back to the instance's obsolete members) and by test: `CopyOnResolve` now forwards each dispatch path to the wrapped resolver's own implementation of that path, replacing only a carried `ValueSet` in the `ResolverResult` and passing failures and other resource types through untouched. The new regression test `TransitiveResolution_ReachesTheResolverThroughItsModernMembers` resolves the include chain through a resolver whose obsolete members throw — it fails with that resolver's `NotSupportedException` on the single-path decorator (verified by stashing the fix) and passes with it.
2. **Once-per-canonical in the class remarks** — qualified: the per-canonical facade layer memoizes only valuesets added or loaded directly; a canonical reached solely as another valueset's include is expanded once per outer valueset that pulls it in.
3. **Same claim in the #1592 fragment** — qualified the same way, so the two fragments agree.
4. **Missing compatibility entry** — the #1596 fragment gains a `## Potentially Breaking` entry mirroring #1592's (a host reading a computed expansion back from a resolver-owned included instance no longer finds one), and the PR now carries the `breaking change` label with the checklist ticked.

Full `CoreTests`: 1377 passed / 0 failed / 8 pre-existing skips; `Cql.Firely` builds with 0 warnings.

---

ewoutkramer:
I am getting a bit nervous about all this copying. This is exactly why in the SDK we allow the expansion to be in place, with proper multi-threaded guards around it. Are we sure we really need to go the keep-on-duplicating-everything way here? It feels wasteful. And now this is designed as a wrapper around an IAsynResolver, it will make copies *every time*, not just on the first Add(). So it is really easy to make a mistake and keep copying and expanding copies of unexpanded valuesets.

alexzautke:
@ewoutkramer We could / should take another look at the locking, based on the race-condition findings from testing. Only the LocalTerminologyService locks the write of the expansion, through its static semaphore. The CQL SDK's ValueSetSource and our the FS terminology plugin's $expand both call ValueSetExpander directly, which skips the locking completely.

ewoutkramer:
> @ewoutkramer We could / should take another look at the locking, based on the race-condition findings from testing. Only the LocalTerminologyService locks the write of the expansion, through its static semaphore. The CQL SDK's ValueSetSource and our the FS terminology plugin's $expand both call ValueSetExpander directly, which skips the locking completely.

Yes - I agree this is actually a bigger architectural issue - I also realized another reason not to cache "internally" like this: Unless we explicitly guard for it, if Vonk decides to invalidate its caches, the CQL plugin will just continue with its old expansions.

Of course, if we switch to just calling $validate-code on a new, shiny, new-admindb based central TX then everything would be solved, but we aren't there yet.

Shall we ask an AI to a) verify the locking in the LocalTS, b) use the same pattern here and the $expand operation?

@cknaap - what do you think?

alexzautke:
> Yes - I agree this is actually a bigger architectural issue - I also realized another reason not to cache "internally" like this: Unless we explicitly guard for it, if Vonk decides to invalidate its caches, the CQL plugin will just continue with its old expansions.

Tthe ValueSetSourceCaching is built explicitly in a way such that it invalidates itself if underlying ValueSet changes. So it's not as bad as it sounds.

ewoutkramer:
@alexzautke @ewoutkramer I traced the full path — this branch, the pinned `ValueSetExpander` (6.3.0), upstream `develop`, `LocalTerminologyService`'s semaphore, and how Vonk actually constructs `ValueSetSource`. Summary below; **we're preparing a follow-up issue** (details at the end), so nothing here asks for a change to this branch yet.

## a) What `LocalTerminologyService` actually locks

`SEMAPHORE` is `private static readonly SemaphoreSlim(1, 1)` — process-global, single permit — and it wraps exactly one thing: the top-level `_expander.ExpandAsync(vs)` in `getExpandedValueSet`, under a `!vs.HasExpansion` pre-check. So:

- **Writer/writer, within `LocalTerminologyService`: covered**, and because the permit is global the expander's internal recursion into `compose.include` is covered too (a second LocalTS caller can't be inside `ExpandAsync` at all).
- **Writer/writer, across entry points: not covered** — exactly Alex's point. `ValueSetSource` and the FS `$expand` plugin call `ValueSetExpander` directly and never touch this semaphore. Confirmed.
- **Writer/reader: not covered, and there is nowhere to put the lock.** `expandAsync` starts with `source.Expansion = ExpansionComponent.Create()` and fills `Expansion.Contains` incrementally per include-group. Any consumer holding the same cached POCO — a REST `$validate-code` walking `FindInExpansion`, another CQL scope's `Load` testing `HasExpansion` — reads it bare, taking no lock. Closing that means a reader lock at *every* consumer of the conformance cache, i.e. changing the cache's contract. That's the architectural issue Ewout named, and it's the reason I don't think locking substitutes here.
- **Failure path:** `Expansion = null` on failure is serialized by a lock, not prevented. The `!HasExpansion` pre-checks mean neither LocalTS nor `ValueSetSource` nor the include path strips a *static* expansion — but a direct `ExpandAsync` caller without that pre-check (the FS `$expand` plugin) can, and every guarded caller still leaves `Expansion == null` where a previously *computed* one stood.

So: a lock fixes tearing between writers. The defect is that the SDK writes into — and on failure nulls from — an instance the host owns and shares. Serialization leaves that write in place.

## b) How wasteful is the copying, really?

Measured on the integration corpus: **662/662 valuesets carry a static expansion** (658 also carry `compose`; 0 have hierarchical expansions). `Add` therefore takes the expansion memo path, the expander never runs, and **neither `DeepCopy` site fires** on real content. Today the copying costs nothing measurable.

The caveat that makes it worth caring about anyway: the expensive path is *self-arming*. It arms as soon as something strips an expansion from a shared instance — and `Vonk.Plugins.Terminology/R{3,4,5}/SdkTerminologyProvider*.cs:110` still sets `ValueSetSource = _resourceResolver` raw, so the FS `$expand` path can still do that. Once it does, every CQL scope's `Load` sees `!HasExpansion` and copy-and-expand becomes hot.

Ewout's "it will copy *every time*, not just on the first `Add()`" is factually right, and it compounds with the qualification Alex already made in point 2 above: the per-canonical facade layer only memoizes canonicals added or loaded *directly*, so a canonical reached only as an include is expanded — and now copied — once per outer valueset that pulls it in. Cheap fix if we keep the wrapper: memoize inside `CopyOnResolve` (`ConcurrentDictionary<string, ValueSet>`, per instance) so a canonical included from two places is copied once, restoring diamond-include reuse. Concurrent for now — `processValueSetGroup`'s `Task.WhenAll` only went sequential in `5c49a407e`, after our pinned 6.3.0.

## c) The invalidation worry is a lifetime question, not a caching one

On Ewout's point about Vonk invalidating its caches: worth separating, because the CQL layer caches **derived code sets, never terminology resources**. `BuildFromExpansion` materializes `CqlCode`s and drops the `ValueSet` — the facade holds no reference to the resource, the expansion, or the source. Three layers, one of which can go stale:

| | what | lifetime | stale after a resolver flush? |
|---|---|---|---|
| L0 | `_internHash` (`CqlCode` intern table) | per source | n/a — value-equal codes |
| L1 | `_valueSets`: canonical → `InMemoryValueSet` | per source, strong, never invalidated | **yes** |
| L2 | `FacadesByExpansion`, weak-keyed on `ValueSet.ExpansionComponent` | process-wide, weak | no — dies with the resolver's object |

L2 already inherits the resolver's invalidation exactly: the entry lives precisely as long as the host keeps that expansion instance alive, so an evicted or replaced valueset takes its facade with it.

L1 is the only hole — and in Vonk it's already closed by *scoping*: `ResolvedLibraryArtifacts.CreateValueSetSource` builds a fresh `ValueSetSource` per DI scope, re-seeded from an immutable `CqlCode` snapshot, and `ResolvedLibraryArtifactsCache` is per measure run. Vonk owns the cross-request caching; the SDK's L1 is effectively per-scope there. The residual exposure is a third-party host keeping one `ValueSetSource` process-wide over a live resolver.

Two things argue against handing more of this to the resolver rather than less: every cold `IsCodeInValueSet` reaches the resolver through `TaskHelper.Await(Load)` — sync-over-async — so L1 is what keeps a blocking resolver call off the membership path inside a retrieve; and L1 is what makes a canonical answer *identically for every subject and group in one run*, which a resolver free to flush mid-run would not. So I'd keep L1 and treat the gap as documentation (`ValueSetSource` is evaluation-scoped) plus an explicit `Clear()`/`Invalidate(canonical)` for hosts that keep one longer — not as a cache to remove.

## d) One mutation vector neither #3582 nor this branch's remarks name

`getExpansionForValueSet` returns the imported valueset's **live** `Expansion.Contains` list; `addCapped` re-lists it but keeps the element references, so the outer expansion ends up holding the *included* valueset's `ContainsComponent` objects. `handleExclude` → `ContainsSetExtensions.Remove` then recurses `component.Contains.Remove(system, code)` over those shared components ([still on upstream `develop`](https://github.com/FirelyTeam/firely-net-sdk/blob/develop/src/Hl7.Fhir.Shims.Base/Specification/Terminology/ValueSetExpander.cs#L337)). So `compose.exclude` on the outer valueset can mutate the expansion of a resolved included valueset **that already had one and was never expanded by us**. #3582's adjacent defect B covers only the `Expansion` assignment/null, not this. It only bites on hierarchical expansions (0 in the corpus), but it means `CopyOnResolve` is covering more than its own remarks claim.

## Follow-up we're preparing

`5c49a407e` closed #3582 with *"Not addressed here: `ValueSetExpander` mutates the resolver-owned ValueSet's Expansion in place… That's a separate, more invasive fix and can be tracked in its own issue"* — and that issue doesn't exist yet. We're drafting it, with three asks:

1. Build the expansion in a local and assign it once — no partially-built expansion ever visible, no null-on-failure.
2. Expand an expansion-less import without writing into the resolved instance.
3. Never share imported `ContainsComponent` trees into the importer's expansion (or make exclude copy-on-write) — point (d).

With those, **both `DeepCopy` sites in `ValueSetSource` become deletable** at the version bump, and L2 starts hitting across sources again instead of each source expanding its own copy. That, rather than CQL-side tuning, is what closes the loop — and it's Firely code on both sides with Vonk pinning the version.

Suggested split, for discussion: keep this branch as the bridge under 6.3.0 (correct today, ~zero cost on real content), optionally add the `CopyOnResolve` memo for the include reuse Ewout flagged, and put the weight on the upstream issue. Happy to go the locking route instead if you'd rather — but it needs a reader-side story for the conformance cache, and a process-global single-permit semaphore serializes every measure run's expansions against each other, which may cost more than the copies do.

---

alexzautke:
> The per-canonical facade layer only memoizes canonicals added or loaded directly, so a canonical reached only as an include is expanded — and now copied — once per outer valueset that pulls it in.

I don't think that's worth doing, it's mainly a handful of ValueSet expansions at most. Don't think there's much overlap practically on the ValueSet.includes across different ValueSets.

ewoutkramer (approved):

ewoutkramer:
Follow-up filed: **FirelyTeam/firely-net-sdk#3596** — the split-out of #3582's adjacent defect B that `5c49a407e` deferred.

It asks for the three things from my analysis above: assemble the expansion in a local and assign it once (no partially built expansion visible, no null-on-failure), expand an expansion-less import without writing to the resolved instance, and stop sharing imported `ContainsComponent` trees into the importer's expansion so `compose.exclude` cannot edit an import's own expansion.

Worth noting for planning: all three writes are present in 6.3.0, 6.4.0 **and** `develop` — `v6.4.0` predates `5c49a407e`, so even the concurrency fix is still unreleased. Once #3596 lands and `FirelyNetVersion` moves past it, both `DeepCopy` sites in `ValueSetSource` (including this branch's `CopyOnResolve`) become deletable, and the expansion memo starts hitting across sources again instead of each source expanding its own copy.

---

PR #1599, opened 2026-09-07
Return null from scalar after/before on null operands

### LINKED ISSUE #1598: Scalar before/after return a boolean instead of null when an operand is null
**Describe the bug**

`before` and `after` on Date/DateTime/Time values return a boolean when an operand is null, where the spec requires null. The four sibling comparison operators get this right, so it looks like a missed guard rather than a deliberate choice.

Spec, [After](https://cql.hl7.org/09-b-cqlreference.html#after) and [Before](https://cql.hl7.org/09-b-cqlreference.html#before), identical wording in both:

> If either or both arguments are null, the result is null.

**To Reproduce**

```cql
library NullCompare version '1.0.0'
using FHIR version '4.0.1'
context Patient

define "null after value": (null as DateTime) after @2024-01-01T00:00:00.000Z
define "value after null": @2024-01-01T00:00:00.000Z after (null as DateTime)
define "null before value": (null as DateTime) before @2024-01-01T00:00:00.000Z
define "value before null": @2024-01-01T00:00:00.000Z before (null as DateTime)
define "null after null": (null as DateTime) after (null as DateTime)
define "null before null": (null as DateTime) before (null as DateTime)
```

**Observed vs expected**

| expression | observed | expected |
| --- | --- | --- |
| `null after value` | `true` | `null` |
| `value after null` | `false` | `null` |
| `null after null` | `false` | `null` |
| `null before value` | `false` | `null` |
| `value before null` | `true` | `null` |
| `null before null` | `false` | `null` |

`Date` behaves the same way. `same or after` and `same or before` are correct, as are `>`, `>=`, `<` and `<=`.

**Where it comes from**

1. There is no typed scalar overload. [`ICqlOperators`](https://github.com/FirelyTeam/firely-cql-sdk/blob/v2.14.0/Cql/Cql.Runtime/Operators/ICqlOperators.cs#L45-L66) declares 22 `After` overloads and all but one take an interval, so a point comparison binds the catch-all `After(object?, object?, string?)`.

2. That catch-all has no null guard. [`CqlOperators.DateTimeOperators.cs#L68-L90`](https://github.com/FirelyTeam/firely-cql-sdk/blob/v2.14.0/Cql/Cql.Runtime/Operators/CqlOperators.DateTimeOperators.cs#L68-L90) passes `left!, right!` and checks only whether the *result* is null.

3. The result is never null for a null operand. [`ICqlComparer.Compare`](https://github.com/FirelyTeam/firely-cql-sdk/blob/v2.14.0/Cql/Cql.Abstractions/Comparers/ICqlComparer.cs#L47-L61) documents its null return as meaning the comparison is uncertain because of `precision`, and `CqlComparerSharedMethods` notes that "the comparers reach here with operands already known non-null".

4. Instead, [`CqlComparer.cs#L192-L201`](https://github.com/FirelyTeam/firely-cql-sdk/blob/v2.14.0/Cql/Cql.Abstractions/Comparers/CqlComparer.cs#L192-L201) gives null a total ordering for sorting: `null` vs value is `1`, value vs `null` is `-1`, `null` vs `null` is `0`. So `null after value` reaches `> 0` and returns `true`, and the rest of the table follows.

**Suggested fix**

The guard already exists in the sibling operators, with the spec cited above it, e.g. [`Greater`](https://github.com/FirelyTeam/firely-cql-sdk/blob/v2.14.0/Cql/Cql.Runtime/Operators/CqlOperators.ComparisonOperators.cs#L99-L113):

```csharp
// https://cql.hl7.org/09-b-cqlreference.html#greater
// Spec:If either argument is null, the result is null.
if (left is null || right is null)
    return null;
```

Adding the same two lines to `After` and `Before` should cover it. Happy to open a PR if that is useful.

**Version used:**
 - Version: 2.14.0

**Additional context**

Found by differential testing against the reference implementation (`@cqframework/cql` 5.2.0) over 278 generated expressions, covering each binary operator crossed with each operand type it accepts and null in every position. These six cells were the only divergences; everything else agreed, including the 77 control expressions with no nulls.

Related but distinct: #1352 covered null *interval boundaries* being read as min/max in the interval operators. This is the scalar path.

alexzautke:
Fixes #1598

## What

Scalar (point) `after` and `before` on `Date`, `DateTime` and `Time` returned a boolean when an operand was null. The CQL spec (§9.B After / Before) says "If either or both arguments are null, the result is null." All six null cells from the issue now evaluate to `null`.

## Why it happened

Point comparisons bind the catch-all `After(object?, object?, string?)` / `Before(object?, object?, string?)` overloads in `CqlOperators.DateTimeOperators.cs`. Those handed both operands straight to `ICqlComparer.Compare`, which deliberately gives null a total ordering so sorting is deterministic (null vs value = 1, value vs null = -1, null vs null = 0). The result was therefore never null for a null operand, so `(null as DateTime) after @2024-01-01T00:00:00.000Z` came out `true`.

This is not related to the SDK's missing support for CQL *uncertainty* (the interval-valued result of `days between` on partially specified dates). Precision-based comparison of partial dates already returns null correctly; a null operand is a separate spec rule.

## Change

- Add the same two-line null guard the sibling `Greater` operator already carries to `After` and `Before` (`Cql/Cql.Runtime/Operators/CqlOperators.DateTimeOperators.cs`). The comparer's null ordering is left alone since sorting depends on it.
- New `Cql/CoreTests/ScalarAfterBeforeNullOperandTests.cs`: the null-operand matrix (null left / null right / both null, `after` and `before`) for all three scalar types, plus controls pinning the unchanged non-null results and the unchanged precision-uncertainty result (`DateTime(2014) after DateTime(2014, 2, 15)` is still null).
- Release-note fragment under `docs/releases/vnext/` flagging that CQL evaluation results move.

No `ICqlOperators` signature change and no `GeneratorToolVersion` bump: the emitted C# is identical, the fix is entirely in the runtime, so already-packaged libraries pick it up on the new runtime.

## Verification

- New tests before the fix: 18 of 23 failed (exactly the null-operand cells); after: 23 of 23 pass.
- Full `CoreTests`: 1401 passed, 8 pre-existing skips, 0 failed on both net8.0 and net10.0.
- Full `CqlToElmTests` (DQIC CQL-level suite): 2505 passed, 33 pre-existing skips, 0 failed.

## Not changed

Other operators that route through the comparer's null ordering (`Between`, interval boundary logic) were not audited here; the issue only reports the scalar `after`/`before` path.

https://claude.ai/code/session_01UEtbVW88NwioXahP17i6ey

---
_Generated by [Claude Code](https://claude.ai/code/session_01UEtbVW88NwioXahP17i6ey)_


Cql/CoreTests/ScalarAfterBeforeNullOperandTests.cs
```diff
@@ -0,0 +1,187 @@
...
+using Hl7.Cql.Operators;
+using Hl7.Cql.Primitives;
+
+namespace CoreTests;
+
+/// <summary>
+/// Covers the scalar (point) <c>after</c>/<c>before</c> operators on Date, DateTime and Time.
+/// Per the CQL spec (§9.B After / Before) "If either or both arguments are null, the result is
+/// null" — see https://github.com/FirelyTeam/firely-cql-sdk/issues/1598, where these returned a
+/// boolean because the point comparison fell through to the comparer's total ordering of null.
```
baseTwo: **Backstory in a source comment.** `CLAUDE.md` (and `.github/copilot-instructions/`) is explicit that the only place backstory goes is the PR description: *"Why a thing was once done, which PR or issue caused it, what used to be pinned, broken or worked around ... none of it belongs in source ... State the rule or the constraint as it is now and stop."*

The clause *"— see .../issues/1598, where these returned a boolean because the point comparison fell through to the comparer's total ordering of null"* is exactly that history.

Concretely it goes stale the moment #1598 closes: the comment then describes behaviour the code no longer has, and the next reader spends a round trip on a closed issue to discover it is irrelevant. The spec rule and the overload-binding note below it are the parts that stay true, so keep those:

```suggestion
/// Per the CQL spec (§9.B After / Before) "If either or both arguments are null, the result is
/// null".
```

(The same applies to the `OverlapsAfterNullBoundaryTests` precedent, but that one predates the rule — no need to touch it here.)

Cql/CoreTests/ScalarAfterBeforeNullOperandTests.cs
```diff
@@ -0,0 +1,187 @@
...
+    [TestMethod]
+    public void DateTime_After_NullLeft_IsNull() =>
+        Assert.IsNull(Operators.After(Boxed((CqlDateTime?)null), Boxed(DateTime(2024, 1, 1)), null));
+
+    [TestMethod]
+    public void DateTime_After_NullRight_IsNull() =>
+        Assert.IsNull(Operators.After(Boxed(DateTime(2024, 1, 1)), Boxed((CqlDateTime?)null), null));
+
+    [TestMethod]
+    public void DateTime_After_BothNull_IsNull() =>
```
baseTwo: **Six of the eighteen null-operand tests are two distinct assertions.** `Boxed` returns `object?`, so `Boxed((CqlDateTime?)null)`, `Boxed((CqlDate?)null)` and `Boxed((CqlTime?)null)` all evaluate to a plain `null` reference — the cast is erased at the call boundary. That makes `DateTime_After_BothNull_IsNull`, `Date_After_BothNull_IsNull` and `Time_After_BothNull_IsNull` the *identical* call `Operators.After(null, null, null)`, and likewise for the three `*_Before_BothNull_IsNull`.

Concretely: the suite reads as if the both-null cell is covered per type for Date, DateTime and Time, but a type-specific regression in that cell cannot be expressed here and would never be caught — the new guard short-circuits before any type is inspected. Anyone later reading a green run of these six as "all three scalar types are covered for both-null" is reading a coverage claim the tests do not make.

Collapse them to one `After_BothNull_IsNull` and one `Before_BothNull_IsNull` (no type prefix), so the method names match what is actually exercised. The `NullLeft`/`NullRight` variants do at least carry a typed non-null operand and are worth keeping per type.

Cql/Cql.Runtime/Operators/CqlOperators.DateTimeOperators.cs
```diff
@@ -67,7 +67,12 @@ internal partial class CqlOperators
 
         public bool? After(object? left, object? right, string? precision)
         {
-            var result = Comparer.Compare(left!, right!, precision);
+            // https://cql.hl7.org/09-b-cqlreference.html#after
+            // Spec: If either or both arguments are null, the result is null.
+            if (left is null || right is null)
```
baseTwo, line 72: **The invariant this fix depends on is now untested.** The PR rightly leaves `CqlComparer<T>.Compare`'s total ordering of null alone (`null` vs value = `1`, value vs `null` = `-1`, `null` vs `null` = `0`) because sorting needs it to be deterministic — but `After`/`Before` were the only thing making that ordering *observable*, and this guard removes that. `Cql/CoreTests/CqlComparersTests.cs` has no assertion on `Compare` with a null operand, and neither does the new file.

Concrete failure: a later contributor reads `After`/`Before`/`Greater`/`Less` — every one of which now returns `null` for a null operand — concludes `CqlComparer<T>.Compare` (`Cql/Cql.Abstractions/Comparers/CqlComparer.cs`, the `(x is null || IsNull(x), y is null || IsNull(y))` switch at ~line 178) should do the same, and changes it to return `null`. `Sort`/`OrderBy` over a list containing nulls then becomes non-deterministic (a `null?` comparison result collapses to "not greater", so relative order depends on input order), and the entire test suite stays green — including all 23 tests added here, because they never reach the comparer.

Please pin it, e.g. in `CqlComparersTests`:

```csharp
Assert.AreEqual(1,  comparers.Compare(null, Dt(2024), null));
Assert.AreEqual(-1, comparers.Compare(Dt(2024), null, null));
Assert.AreEqual(0,  comparers.Compare(null, null, null));
```

docs/releases/vnext/1599-scalar-after-before-null-operands.md
```diff
@@ -0,0 +1,18 @@
+## Fixes
+
+- Scalar (point) `after` and `before` on Date, DateTime and Time now return `null` when either or
```
baseTwo: **Fragment does not name the SDK component.** House style is to lead each release-note bullet with the aspect of the SDK that changed — translator, code generator, runtime, Packager CLI. The only other fragment currently in this folder (`1615-central-package-management.md`) opens its bullets `**Build tooling:**` and `**Documentation:**`, and the versioned notes under `docs/releases/` do the same.

Concretely: at release-cut time every fragment here is folded into one document and these bullets end up interleaved with entries from other PRs. A reader scanning `## Fixes` for "does this affect libraries I already packaged?" has to read the whole paragraph to learn that this one is runtime-only — which is precisely the fact the last sentence of the fragment is at pains to establish.

```suggestion
- **Runtime:** scalar (point) `after` and `before` on Date, DateTime and Time now return `null` when either or
```

baseTwo:
@claude[agent] please fix all review comments, reply and resolve them too

baseTwo (approved):

PR #1603, opened 2026-09-08
Walk qualified paths in LateBoundProperty

alexzautke:
## Description
`CqlOperators.LateBoundProperty` treated its `propertyName` as a single property name, but an ELM `Property` element's `path` may be qualified (`medication.reference.value`) and the code generator passes it through verbatim. Reflection resolves one property name at a time, so the whole dotted string matched nothing and every qualified late-bound access silently evaluated to `null`.

The compiler already walks such paths segment by segment when the types are known at design time (`CodeBuilderContext.Query.cs`, "support paths like birthDate.value on Patient"); only the late-bound fallback lacked the walk. This PR mirrors it in the runtime operator:

- intermediate segments resolve as `object`, only the final segment is converted to the requested type;
- a null or unresolvable segment short-circuits to `null`, so an unresolvable path still degrades rather than throws.

Found while tracing the MADiE integration corpus: in CMS165 the content resolves `MedicationRequest.medication` given as a `Reference` by comparing the bundled `Medication`'s `id.value` against the last segment of `medication.reference.value`. Both sides were qualified late-bound paths, both returned `null`, and the reference-following branch of the retrieve never matched. Case `45e01fed` now passes (Denominator Exclusions `true`).

No generated C# changes, so no `GeneratorToolVersion` bump. Release fragment added under `docs/releases/vnext/`.

## Related issues
None filed; traced from the CQL SDK Integration Test Runner corpus.

## Testing
- 4 new tests in `Cql/CoreTests/LateBoundPropertyTests.cs`, including a bundle fixture reproducing the `MedicationRequest.medication` → bundled `Medication` shape.
- Full `CoreTests` suite passes (net10.0).
- Integration corpus in `ReadAndRunAllValidatePreviousPasses` mode: all baseline passes still pass, `45e01fed` newly passes, zero regressions.

## FirelyTeam Checklist

https://claude.ai/code/session_01AoPNvHfeXidST7F5ypteFd

---
_Generated by [Claude Code](https://claude.ai/code/session_01AoPNvHfeXidST7F5ypteFd)_


ewoutkramer:
Superseded by #1638

Issue #1427, opened 2026-07-27
Drop .NET 8 support ahead of its 10 November 2026 end-of-support date


baseTwo:
## Summary

Microsoft has announced that .NET 8 reaches End of Support on **10 November 2026**. Per our own announcement in [discussion #1426](https://github.com/FirelyTeam/firely-cql-sdk/discussions/1426), CQL SDK 2.x support for .NET 8 ends on the same date. This ticket tracks the work needed to actually drop .NET 8 from the supported target matrix ahead of that deadline, rather than letting the date pass with no corresponding change to the SDK. As part of this cleanup, also migrate the repo off the legacy `.sln` solution format to `.slnx`.

## Impact

- Consumers building against CQL SDK 2.x on .NET 8 will be running on an unsupported .NET runtime after 10 November 2026, with no further compatibility validation or issue resolution from us for that target (per the announcement).
- Without an explicit change on our side, nothing actually stops CQL SDK 2.x from continuing to target/build against .NET 8 past the deadline — the announcement is a policy statement, not an enforced constraint.

## Expected Behavior

Ahead of 10 November 2026:

- CQL SDK 2.x's supported target framework list no longer includes .NET 8 (project `TargetFramework(s)` updated across the relevant `.csproj` files).
- Packaging/CI reflects the updated target matrix.
- Docs/release notes call out the dropped target explicitly.
- The old `.sln` solution files are removed in favor of `.slnx` (the newer XML-based solution format), so the solution tooling itself is modernized alongside the target-framework cleanup.

## Acceptance Criteria

- [ ] Decide and confirm the release (version) in which .NET 8 support is actually removed.
- [ ] Remove `net8.0` (and any net8.0-specific conditionals) from the affected project files.
- [ ] Update CI pipeline definitions to stop building/testing against .NET 8.
- [ ] Convert the repo's `.sln` files to `.slnx` and remove the old `.sln` files (including any `.slnf` filter files, which will need `.slnx`-compatible equivalents), keeping build scripts working on both Windows and Unix.
- [ ] Update any documentation, CI pipeline references, and tooling that point at the removed `.sln`/`.slnf` files to use the new `.slnx` files.
- [ ] Add an entry to the relevant release notes documenting the dropped target and the solution-file format change, referencing [discussion #1426](https://github.com/FirelyTeam/firely-cql-sdk/discussions/1426).
[...]

PR #1613, opened 2026-09-16
Cut 2.15.0 release notes and version bump

### LINKED ISSUE #1593: Cut the next CQL SDK release (after 2.14.0)
## Summary

Cut the next CQL SDK release. Current version is `2.14.0` (`v2.14.0` tag, cut 2026-08-20); this covers everything accumulated on `develop` since then, up to and including the version-bump PR. Creating the git tag, the GitHub Release, and approving the NuGet/GitHub Packages deploy are separate, permission-gated release-manager steps and are **not** part of this ticket.

Full authoritative process: [Creating Tags and Releases](https://github.com/FirelyTeam/firely-cql-sdk/wiki/Creating-Tags-and-Releases) wiki page. This ticket tracks working through the [`cut-release-notes`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/.claude/skills/cut-release-notes/SKILL.md) skill.

**One requirement is new this cycle:** the release notes must open with a standing warning that .NET 8 support ends on 10 November 2026, pointing consumers at [discussion #1426](https://github.com/FirelyTeam/firely-cql-sdk/discussions/1426), and `release-notes-template.md` must carry a slot for it so the next cut does not drop it. See [.NET 8 end-of-support warning](#net-8-end-of-support-warning) below.

> **Evidence refreshed 2026-09-16** against `origin/develop` at `8860b1332`. This ticket was filed on 2026-08-27 when the release window was empty, and its original version-level reasoning (*"zero fragments, zero pending public API entries, documentation only"*) has since been overtaken: four `Potentially Breaking` fragments have landed. The superseded counts are kept out of this body deliberately — see the edit history if you need what it said at filing.

`develop` is **38 commits ahead** of `v2.14.0` (head `8860b1332`), with **seven** merged PRs in the window.

## Version number — decided: MESO, `2.15.0`

The level is **MESO** under [`versioning.md`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/versioning.md), forced independently by two triggers:

1. **Fragments declaring `## Potentially Breaking`** — all four in this window carry one (#1592, #1595, #1596, #1606).
2. **CQL evaluation results change** — #1595 makes `if`/`else if` branches return their own values where they previously evaluated to `null`, so measure output moves. `versioning.md` counts this "even when the old output was a defect".

Triggers that did **not** fire:

- **`GeneratorToolVersion` unchanged at `5.2.1.0`**, invoker range still `[5.1.0.0, 5.3.0.0)`. #1595 and #1606 do change the emitted C#, but by changing the ELM the *translator* produces, not the ELM-to-C# generator — so the generator version correctly stays put, and the release notes say so explicitly rather than leaving readers to infer that generated code is unchanged.
- **No public API movement** — all 13 `PublicAPI.Unshipped.txt` files are header-only, so the promotion step is a verified no-op.
- **Dependency major crossing — a near miss.** `Microsoft.SourceLink.GitHub` went `8.0.0` → `10.0.401` (#1609), crossing two majors, but the reference is `PrivateAssets="All"` and so never enters a consumer's dependency graph. The trigger exists for the case where "the consumer's own graph moves with it", which this is not.

**MACRO was considered and rejected.** The measured trigger wants a re-architecture, a whole surface area removed, or several MESO-level migrations together. Four `Potentially Breaking` entries look like the third, but each is narrow — the two translator entries touch only defines containing one specific branch-type pattern, the two valueset entries only hosts that read a computed `Expansion` back off an instance they supplied. No declared MACRO has been taken either.

## Pending content to consolidate

Four fragments under `docs/releases/vnext/`, all carrying a `## Potentially Breaking` section as well as `## Fixes`:

| Fragment | PR | Closes |
|---|---|---|
| `1592-valuesetsource-no-mutation.md` | #1592 | #1591 |
| `1595-conditional-branch-type-narrowing.md` | #1595 | #1594 |
| `1596-copy-on-resolve-included-valuesets.md` | #1596 | #1597 |
| `1606-if-any-typed-branch-widening.md` | #1606 | #1601 |

`docs/releases/vnext-release-notes.md` is confirmed still the static pointer doc it became in #1445 — one sentence pointing at `vnext/`, nothing to fold in.

**One template gap to note:** the fragments carry `## Fixes` content, but `release-notes-template.md`'s structure has no `#### Fixes` heading under CQL SDK (only New Public API, Improvements, Dependency Updates, Potentially Breaking). The cut adds one rather than dropping four substantial fix entries or mislabelling them as Improvements. Worth folding back into the template as a follow-up.

### #1589 — decision recorded

#1589 ("Document EffVer as the SDK versioning scheme", closing #1588) added [`docs/versioning.md`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/versioning.md) and touched instruction files. `versioning.md` puts contributor-facing documentation under *What never affects the version*, so it contributes **nothing** to the level.

On the judgement call the original filing flagged: the versioning policy **is** consumer-relevant — a consumer reading our version numbers as SemVer has been reading them wrong — so it earns a one-line pointer under `#### Improvements` rather than only a PR-table row. Both are included.

### Merged PRs in the window

| PR | Bearing on the release |
|---|---|
| #1589 | Documentation only; pointer under Improvements (above). |
| #1592 | Fragment: Potentially Breaking + Fixes. |
| #1595 | Fragment: Potentially Breaking + Fixes. **CQL results move.** |
| #1596 | Fragment: Potentially Breaking + Fixes. |
| #1606 | Fragment: Potentially Breaking + Fixes. |
| #1609 | `Microsoft.SourceLink.GitHub` `8.0.0` → `10.0.401` (NU1902). Root README table already updated by that PR. |
| #1610 | Dependabot config; CI only, outside the version scales. |

### Still open — re-check before the version-bump PR merges

| PR | State | If it lands first |
|---|---|---|
| #1611 — "Update Firely .NET SDK to 6.5.0" | **Open, non-draft** | Bumps `FirelyNetVersion` `6.3.0` → `6.5.0`. Needs the root README External Dependencies table updated, a fresh Vonk `FhirNetApiVersion` check, and a Dependency Updates entry. Does not change the level (no major crossing). |
| #1580 — "Emit nullable-aware generated C#" | Draft | Changes emitted C#, so it *would* bump `GeneratorToolVersion`, needs the invoker range re-checked, and needs every checked-in `*.g.cs` plus the integration-runner resources regenerated. |
| #1272 — "Align Matches operator with CQL regex spec" | Draft | Spec-conformance behaviour change; needs a fragment. |

## .NET 8 end-of-support warning

The release notes must **open** with a warning that .NET 8 support ends on 10 November 2026, so that consumers still on `net8.0` meet it before anything else in the document. Source of truth for the wording and dates is [discussion #1426](https://github.com/FirelyTeam/firely-cql-sdk/discussions/1426); the work that actually removes the target is [#1427](https://github.com/FirelyTeam/firely-cql-sdk/issues/1427), still open. The SDK still targets `net8.0;net10.0`, so this remains a notice rather than a shipped change.

Three things make this more than a paste:

1. **Placement fights the template.** [`release-notes-template.md`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/releases/release-notes-template.md) states that "The `tl;dr` section must be the very first section after the version heading". A notice above the `tl;dr` therefore needs the template amended to carry a slot for it, not just a one-off insertion — otherwise the next cut silently drops the warning while the deadline gets closer.
2. **It must not leak into the version level.** It does not belong in the `tl;dr`'s **Breaking changes** bullet, and must not appear in the `### Version Level` reasoning. Getting this wrong manufactures a MESO out of a documentation addition.
3. **It has an expiry.** The release that actually drops `net8.0` (#1427) replaces the warning with the accomplished fact. The template slot says so explicitly.

The same notice belongs in the GitHub Release body and the release announcement discussion (the `vX.Y.Z` Announcements post, e.g. [#1590](https://github.com/FirelyTeam/firely-cql-sdk/discussions/1590)) — those are release-manager steps outside this ticket, so they are flagged in the release PR description rather than done here.

## Acceptance Criteria

- [x] Version bump level decided and justified in the PR description, citing the rule in [`versioning.md`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/versioning.md) rather than what previous releases did.
- [x] New `docs/releases/release-notes-<version>.md` written using [release-notes-template.md](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/releases/release-notes-template.md)'s `tl;dr` format, product-area grouping, and "Common Items To Check" list, based on every commit since `v2.14.0`.
- [x] **The new release-notes document opens with the .NET 8 end-of-support warning**, above the `tl;dr`, linking [discussion #1426](https://github.com/FirelyTeam/firely-cql-sdk/discussions/1426) and [#1427](https://github.com/FirelyTeam/firely-cql-sdk/issues/1427), and naming 10 November 2026 and .NET 10 as the recommended target.
- [x] **`release-notes-template.md` carries a slot for that notice**, so it survives the next cut: placed above the `tl;dr`, with the existing "the `tl;dr` section must be the very first section after the version heading" guidance amended to admit it, and with an explicit instruction to remove the notice in the release that drops `net8.0` (#1427).
- [x] The warning is **not** counted as a breaking change: it does not appear in the `tl;dr`'s Breaking changes bullet and does not feature in the `### Version Level` reasoning.
- [x] All four fragment files under `docs/releases/vnext/*.md` folded into the matching section of the new document, with headings normalised against the template, then **deleted**.
- [x] A decision recorded on #1589's missing fragment, including whether `docs/versioning.md` earns a consumer-facing pointer.
- [x] `docs/releases/vnext-release-notes.md` confirmed to still be the static pointer doc, with any stray directly-authored bullets folded in.
- [x] Exact public type/member/property/method names called out for API changes; dependency version bumps name the version property and old/new value.
- [x] `VersionPrefix` bumped in every `*.props` file that declares it — `cql-sdk.props` and `Demo/cql-demo.props`, both at `2.14.0` (add `VersionSuffix` only for a pre-release).
- [x] `README.md`'s "This is release version 2.14.0 of the engine." line updated to match.
- [x] `FirelyNetVersion` confirmed against Vonk's `FhirNetApiVersion` in [Vonk's `Directory.Packages.props`](https://github.com/FirelyTeam/Vonk/blob/develop/Directory.Packages.props), and the root `README.md` External Dependencies table updated if it moved. **Checked, and the result is drift, not parity:** #1611 landed after this ticket was refreshed and moved ours `6.3.0` → `6.5.0` (README table updated with it), while Vonk's `develop` is still `6.3.0`. `cql-base.props` requires the two to be equal, so that invariant does not hold as of 2026-09-16. The check itself is done and the outcome is recorded in the release notes' Common Items To Check; **resolving the drift is a Vonk-side decision and is deliberately out of scope for this ticket.**
- [x] Pending `PublicAPI.Unshipped.txt` entries promoted to each project's `PublicAPI.Shipped.txt` — expected to be a no-op (all 13 header-only); verify the full set rather than assuming.
- [x] `GeneratorToolVersion`'s final value recorded in the release notes' required `### Generated Code Compatibility` section, and the invoker's supported range confirmed to cover it. **State explicitly that emitted C# changes anyway** for libraries hitting the #1595/#1606 patterns, so consumers do not read an unchanged generator version as "nothing to regenerate".
- [x] `Cql-Sdk.slnf` builds clean **in `Release`** (the PublicAPI analyzer's RS0016/RS0017 are suppressed in `Debug`, so a green `Debug` build proves nothing here) and `CoreTests` pass on both target frameworks.

---

_Filed 2026-08-27. Evidence refreshed 2026-09-16 against `origin/develop` at `8860b1332`: window grown from 7 commits to 38, from one merged PR to seven, and from zero fragments to four — all four `Potentially Breaking`, which settles the level at MESO. Follows the shape of #1586 (2.14.0), #1557 (2.13.0) and #1503 (2.12.0)._

baseTwo:
Fixes #1593.

Cuts the `2.15.0` release notes and bumps the version. Creating the git tag, the GitHub Release, and approving the NuGet/GitHub Packages deploy remain separate release-manager steps and are not part of this PR.

### Version level: MESO → `2.15.0`

Decided against [`docs/versioning.md`](https://github.com/FirelyTeam/firely-cql-sdk/blob/develop/docs/versioning.md), not against precedent. Two triggers fired independently, either sufficient on its own:

1. **Fragments declaring `## Potentially Breaking`** — four of the five in this window carry one (#1592, #1595, #1596, #1606).
2. **CQL evaluation results change** — #1595 makes `if`/`else if` branches return their own values where they previously evaluated to `null`, so measure output moves. `versioning.md` counts this "even when the old output was a defect".

**MACRO was considered and rejected.** The measured trigger wants a re-architecture, a whole surface area removed, or several MESO-level migrations landing together. Four `Potentially Breaking` entries resemble the third, but each is narrow — the two translator entries affect only defines containing one specific branch-type pattern and generate exactly as before otherwise; the two valueset entries affect only hosts that read a computed `Expansion` back off an instance they supplied. Nothing was removed or re-architected, and no **declared** MACRO (a new engine generation) has been taken.

Triggers that did **not** fire are recorded in the notes too, so the next cut does not have to re-derive them.

### Release notes

New [`docs/releases/release-notes-2.15.0.md`](../blob/1593-cut-next-cql-sdk-release/docs/releases/release-notes-2.15.0.md), consolidating all five `docs/releases/vnext/` fragments, which are deleted in this PR. `vnext/` is back to just its reserved `README.md`. `vnext-release-notes.md` was confirmed to still be the static pointer doc from #1445 — nothing stranded there.

Two calls in the notes are worth a reviewer's attention:

- **`GeneratorToolVersion` stays at `5.2.1.0`, but generated C# changes anyway.** #1595 and #1606 alter the emitted C# by changing the ELM the *translator* produces, not the ELM-to-C# generator — the generator emits what it always did for a given input, so its version correctly does not move. Because an unchanged generator version would otherwise read as "nothing to regenerate", the `### Generated Code Compatibility` section states the distinction explicitly and the Upgrade Checklist leads with regeneration.
- **`Microsoft.SourceLink.GitHub` `8.0.0` → `10.0.401` does not fire the dependency-major trigger.** It crosses two of that package's majors, but the reference is `PrivateAssets="All"` and never enters a consumer's dependency graph, which is the case that trigger exists for. Recorded as an explicit near miss so a future cut does not apply it mechanically — a bump of a package *without* `PrivateAssets` across its major does fire it.

### Version bump

- `VersionPrefix` `2.14.0` → `2.15.0` in `cql-sdk.props` and `Demo/cql-demo.props` (no `VersionSuffix`; this is a final release).
- `README.md`'s "This is release version X of the engine." line updated to match.

### .NET 8 end-of-support notice, and a template slot for it

New this cycle, per #1593. The notes open with a warning that .NET 8 support ends on 10 November 2026, above the `tl;dr`, linking [discussion #1426](https://github.com/FirelyTeam/firely-cql-sdk/discussions/1426) and #1427.

`release-notes-template.md` now carries a slot for it rather than this being a one-off insertion — otherwise the next cut silently drops the warning as the deadline approaches. The template's existing "the `tl;dr` section must be the very first section after the version heading" guidance is amended to admit the notice, and the slot carries an explicit instruction to delete it in the release that actually drops `net8.0` (#1427), which replaces the warning with the accomplished fact.

The notice is deliberately kept out of the `tl;dr`'s **Breaking changes** bullet and out of the `### Version Level` reasoning: the SDK still ships `net8.0` in this release, so it is a forward-looking statement about a date, not a change being shipped. Counting it would manufacture a MESO out of a documentation addition.

---

### `FirelyNetVersion` moves ahead of Vonk — expected, and worth stating

#1611 landed while this PR was open, bumping `FirelyNetVersion` `6.3.0` → `6.5.0`. This branch merged `develop`, folded #1611's fragment into the Dependency Updates section, and added it to the PR table. The bump stays inside the Firely .NET SDK's own `6.x`, so it is **not** a MESO trigger and the level is unchanged.

[`cql-base.props:13-14`](../blob/develop/cql-base.props#L13-L14) asks `FirelyNetVersion` to match `FhirNetApiVersion` in [Vonk's `Directory.Packages.props`](https://github.com/FirelyTeam/Vonk/blob/develop/Directory.Packages.props), and Vonk's `develop` is still `6.3.0`. **This is the expected order, not a defect:** the SDK moves first, and Vonk moves to `6.5.0` as part of upgrading to the CQL SDK release that carries it, so the two converge at adoption.

The reason it is still worth a line in the notes is the consumer-facing consequence: `Hl7.Cql.Fhir` `2.15.0` declares `Hl7.Fhir.Base >= 6.5.0`, so any host that pins the Firely .NET SDK directly — Vonk included — has to move that pin in the same step as the SDK upgrade, not afterwards. The `Common Items To Check` entry says exactly that, rather than asserting a parity that does not exist at cut time.

Nothing in this release depends on Vonk, and the full suite is green at `6.5.0`.

### A `#### Fixes` heading the template does not define

The fragments carry `## Fixes` content, but `release-notes-template.md`'s CQL SDK structure has only New Public API, Improvements, Dependency Updates and Potentially Breaking. This PR adds a `#### Fixes` heading to the 2.15.0 document rather than dropping four substantial fix entries or mislabelling them as Improvements. **The template itself is not changed for this** — folding the heading back into the template is flagged on #1593 as a follow-up, so it can be decided on its own merits rather than inside a release cut.

### #1589 recorded rather than passed over

#1589 ("Document EffVer as the SDK versioning scheme") landed in this window with no fragment. That is correct — `versioning.md` puts contributor-facing documentation under *What never affects the version* — but the conclusion is recorded explicitly rather than left unaddressed.

It does earn a consumer-facing line: a consumer who has been reading our version numbers as SemVer has been reading them wrong, so `docs/versioning.md` gets a one-line pointer under `#### Improvements` as well as its PR-table row.

## Verification

Re-run after merging `develop` (so these numbers are against `FirelyNetVersion` `6.5.0`, not `6.3.0`):

- `dotnet build Cql-Sdk.slnf -c Release` — **0 warnings, 0 errors**, both `net8.0` and `net10.0`. Release specifically, since the PublicAPI analyzer's RS0016/RS0017 are suppressed in Debug.
- `CoreTests` — 1378 passed, 0 failed, on both target frameworks.
- `CqlToElmTests` — 2509 passed, 0 failed, on both target frameworks.
- Counts are identical to the pre-merge run, so the Firely .NET SDK bump changed no observable behaviour here.
- All 13 `PublicAPI.Unshipped.txt` files verified header-only, so the promotion step is a confirmed no-op rather than an assumed one.


docs/releases/release-notes-2.15.0.md
```diff
@@ -0,0 +1,140 @@
...
+> - **Highlights:** An `if`/`else if` chain returning lists of different tuple types no longer silently returns `null` for every branch but the first — a defect that produced wrong measure scores with no error and no warning.
+
+---
+
+### Version Level
+
+- **MESO** (`2.15.0`) — the second digit, meaning some small effort may be required. Two triggers from [versioning.md](../versioning.md) fired independently, either of which forces the level on its own:
+
+  1. **Fragments declaring `## Potentially Breaking`.** All four release-note fragments in this window carry one: #1592, #1595, #1596 and #1606.
+  2. **CQL evaluation results change.** #1595's fix makes branches of an `if`/`else if` chain return their own values where they previously evaluated to `null`, so a measure containing that pattern scores differently than it did on `2.14.0`. versioning.md counts this even though the old output was a defect: "A measure whose output moves is a migration even when the old output was a defect."
```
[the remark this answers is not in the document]
baseTwo: Fixed in 3256c01ab.

Both later mentions of `versioning.md` in the `### Version Level` section (the "CQL evaluation results change" trigger and the "MACRO was considered and rejected" paragraph) are now `[versioning.md](../versioning.md)` links. The first mention in that section was already linked, which is what made the other two easy to miss.

Swept the rest of the document for the same defect — no bare `.md` cross-references remain.

docs/releases/release-notes-2.15.0.md
```diff
@@ -0,0 +1,140 @@
...
+
+  Triggers that did **not** fire, recorded so the next cut does not have to re-derive them:
+
+  - `LibrarySetCSharpCodeGenerator.GeneratorToolVersion` is unchanged at `5.2.1.0`. #1595 and #1606 change the C# that comes out of the pipeline, but they do so by changing the **ELM the translator produces**, not the ELM-to-C# generator — the generator emits the same C# it always did for a given input. The version marks the generator, so it does not move here.
```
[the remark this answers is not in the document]
baseTwo: Not applying this one — I believe it misreads what `GeneratorToolVersion` marks, and the question was already settled on the PRs that introduced the change.

**`GeneratorToolVersion` marks the ELM-to-C# generator, not the pipeline end-to-end.** #1595 and #1606 live in `Cql.CqlToElm`. They change the **ELM** the translator produces; the generator then emits, for that new ELM, exactly the C# it always would have. No generator behaviour moved, which is why the golden C# generation tests in `CoreTests` are green and unmodified (1378 passed, both TFMs). The version tells a consumer *which generator* produced their artifacts and lets `LibraryInstanceInvoker_5_0` decide whether it can invoke them — and the shape of generated code, which that range governs, is unchanged here. `05-build-and-test.md:5.3.1` reads broadly, but its three enumerated cases are all generator-side (`CodeGeneration.NET`, `CqlOperatorsBinder` emitting different `ICqlOperators` calls, `ICqlOperators` signature changes). A translator fix is none of them.

**It was decided on #1595, where it belonged.** That PR merged deliberately without a bump, and so did #1606. Re-litigating it in the release-notes PR would retroactively change two already-merged changes.

**The consumer-facing concern behind the comment is real, and is addressed — just not via the version.** Affected libraries genuinely do need regenerating. That is why the `### Generated Code Compatibility` section states explicitly that generated C# changes even though the generator version did not, and why the Upgrade Checklist leads with regeneration rather than leaving it implied. An unchanged version number reading as "nothing to do" was the exact failure mode worth guarding against.

For completeness on scope: applying this would mean bumping the version, re-checking the invoker range, regenerating every checked-in `*.g.cs`, regenerating the integration runner's vendored resources, re-running the integration suite and bumping the submodule pointer — inside a release cut, for two changes that already shipped.

Leaving this thread open rather than resolving it, since it is a judgement call and a human should have the final word.

baseTwo:
Filed #1614 for the structural half of the Vonk-sync problem this PR flags.

To be clear about the split, since they are easy to conflate:

- **This PR** records that `FirelyNetVersion` is `6.5.0` here while Vonk's `develop` is `6.3.0`, so the invariant in `cql-base.props` does not currently hold. It does not fix that.
- **#1614** proposes a root `Directory.Packages.props` matching Vonk's, so versions live in one authoritative file and the cross-repo obligation is expressed somewhere a check can reach. It explicitly does **not** pick a version either — it makes the divergence visible instead of leaving it to whoever next reads Vonk's file by hand, which is how this one surfaced.
- **Resolving 6.5.0 vs 6.3.0** remains a Vonk-side decision and is out of scope for both.

docs/releases/release-notes-2.15.0.md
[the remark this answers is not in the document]
baseTwo: Closing this thread: the decision stands, and the release ships at `5.2.1.0`.

Confirmed against the branch at 66400b4dd — `GeneratorToolVersion` is unchanged, `LibraryInstanceInvoker_5_0`'s range `[5.1.0.0, 5.3.0.0)` still covers it, and the golden C# generation tests in `CoreTests` are green and unmodified (1378 passed on both target frameworks), which is the evidence that no generator behaviour moved.

The substantive point was settled when #1595 and #1606 merged without a bump; this release-notes PR is not the place to reverse that for two changes that already shipped. The consumer-facing obligation the comment was really pointing at — that affected libraries need regenerating — is stated explicitly in `### Generated Code Compatibility` and leads the Upgrade Checklist, so it is not lost by leaving the version alone.

No code change made, by design.

docs/releases/release-notes-2.15.0.md
```diff
@@ -0,0 +1,142 @@
...
+
+> **Upgrading?** Here is the short version:
+>
+> - **Breaking changes:** Two CQL-to-ELM translator typing corrections change the C# generated for conditional expressions — an `if`/`else if` chain whose branches share no common type is now typed as the *choice* of those types (generating `object`), while an `if` with one `Any`-typed branch now takes the other branch's concrete type. Separately, expanding a valueset no longer writes the computed expansion back into the `ValueSet` instance the host supplied or the resolver owns.
+> - **Required migrations:** Regenerate packaged artifacts and checked-in `*.g.cs` for any library containing those conditional patterns, and re-verify measure output — results move where an `if`/`else` chain previously evaluated branches to `null`. If your host reads `ValueSet.Expansion` back after `Add`/`Load` or after expanding a valueset with `compose.include`, expand for yourself instead.
+> - **Highlights:** An `if`/`else if` chain returning lists of different tuple types no longer silently returns `null` for every branch but the first — a defect that produced wrong measure scores with no error and no warning.
+
+---
+
+### Version Level
```
alexzautke: Suggest to leave the reasoning out of the release notes, too verbose.

alexzautke (changes requested):

docs/releases/release-notes-2.15.0.md
```diff
@@ -0,0 +1,142 @@
...
+3. **Update any code compiled against an affected define's previous signature** — this is a source-level change where the signature moved to or from `object`.
+4. **If your host reads `ValueSet.Expansion` back** off an instance it passed to `ValueSetSource.Add`/`Load`, or off a valueset reached through another's `compose.include`, expand it yourself — the SDK now expands on a private copy and no longer writes the result back.
+5. **If you relied on two `ValueSetSource` instances with different resolvers sharing an expansion** computed by whichever ran first, note that each now answers from its own resolver. Serve valuesets with static expansions, or seed sources via `Add(string, IEnumerable<CqlCode>)`, if you need sharing.
+6. **If you are on `net8.0`, start planning the move to `net10.0`** — see the warning at the top of this document.
+
+---
+
+### Common Items To Check
+
+- Version level decided against [versioning.md](../versioning.md): **MESO**, triggered by four `Potentially Breaking` fragments and independently by changed CQL evaluation results. Recorded in full under `### Version Level` above.
```
alexzautke: No need to carry the reasoning

alexzautke (changes requested):

docs/releases/release-notes-2.15.0.md
baseTwo: @claude[agent] we do not need to reason about selecting a version number in the release notes, that is already explained elsewhere. Please update your instructions, templates, pr description and the release notes.

docs/releases/release-notes-2.15.0.md
baseTwo: @claude[agent] see  my earlier comment on reasoning

docs/releases/release-notes-2.15.0.md
```diff
@@ -0,0 +1,130 @@
...
+
+> **Upgrading?** Here is the short version:
+>
+> - **Breaking changes:** Two CQL-to-ELM translator typing corrections change the C# generated for conditional expressions — an `if`/`else if` chain whose branches share no common type is now typed as the *choice* of those types (generating `object`), while an `if` with one `Any`-typed branch now takes the other branch's concrete type. Separately, expanding a valueset no longer writes the computed expansion back into the `ValueSet` instance the host supplied or the resolver owns.
+> - **Required migrations:** Regenerate packaged artifacts and checked-in `*.g.cs` for any library containing those conditional patterns, and re-verify measure output — results move where an `if`/`else` chain previously evaluated branches to `null`. If your host reads `ValueSet.Expansion` back after `Add`/`Load` or after expanding a valueset with `compose.include`, expand for yourself instead.
+> - **Highlights:** An `if`/`else if` chain returning lists of different tuple types no longer silently returns `null` for every branch but the first — a defect that produced wrong measure scores with no error and no warning.
+
+---
+
+### Version Level
```
baseTwo: @claude[agent] just remove the Version Level from the release notes (and template) altogether

baseTwo:
@claude[agent] please fix review comments

alexzautke (approved):

PR #1615, opened 2026-09-16
Adopt central package management

### LINKED ISSUE #1614: Adopt central package management: add a Directory.Packages.props matching Vonk's
## Summary

Adopt NuGet Central Package Management (CPM) in `firely-cql-sdk` by introducing a root `Directory.Packages.props`, mirroring the arrangement [Vonk already uses](https://github.com/FirelyTeam/Vonk/blob/develop/Directory.Packages.props): `ManagePackageVersionsCentrally=true`, a `PropertyGroup` of version variables for families of related packages, and one `PackageVersion` item per package, with every `.csproj` carrying a bare `<PackageReference Include="..." />` and no `Version` attribute.

Today this repository has **no** `Directory.Packages.props`. Package versions are literal strings spread across individual `.csproj` files, with a single property-driven exception (`FirelyNetVersion`) declared in [`cql-base.props:16`](https://github.com/FirelyTeam/firely-cql-sdk/blob/0a0576c9a679895d79fb94917522481eaed224c4/cql-base.props#L16) and again in `Demo/cql-demo.props:16`.

## Impact

**The motivating incident is live as of 2026-09-16.** [`cql-base.props:13-14`](https://github.com/FirelyTeam/firely-cql-sdk/blob/0a0576c9a679895d79fb94917522481eaed224c4/cql-base.props#L13-L14) states the invariant in a comment:

> Keep FirelyNetVersion in sync between cql-base.props and cql-demo.props and README.md
> The version must be the same as FhirNetApiVersion in https://github.com/FirelyTeam/Vonk/blob/develop/Directory.Packages.props

#1611 moved our `FirelyNetVersion` from `6.3.0` to `6.5.0`. Vonk's `FhirNetApiVersion` on `develop` is still `6.3.0`.

**That divergence is expected, and correcting an earlier framing of this ticket: it is not a defect.** The SDK moves first, and Vonk moves to `6.5.0` as part of upgrading to the CQL SDK release that carries it, so the two converge at adoption. The problem is not that the versions differ today — it is that **the comment states a rule ("must be the same") that does not describe the actual lifecycle (the SDK legitimately leads for a while), and nothing anywhere expresses or checks the constraint that really has to hold.**

The real constraint is a *floor tied to adoption*: `Hl7.Cql.Fhir` `2.15.0` declares `Hl7.Fhir.Base >= 6.5.0`, and Vonk pins `Hl7.Fhir.Base` directly, so at the moment Vonk bumps `FirelyCqlVersion` its `FhirNetApiVersion` must already be at or above ours or NuGet reports a downgrade. Nothing encodes that. It was noticed only because a release cut (#1613) happened to re-read Vonk's file by hand — which is fine for a divergence that is expected, and no good at all for one that is not.

There is a second, quieter instance of the same pattern: Vonk's file carries `<!-- Keep MicrosoftCodeAnalysisVersion in sync with the CQL SDK's -->` against `MicrosoftCodeAnalysisVersion` `5.3.0`. Our `Microsoft.CodeAnalysis.CSharp` is also `5.3.0`, so that one holds today — but it is enforced by exactly the same mechanism, which is to say none, and it is asymmetric: their file names the obligation, ours does not.

**Within this repository, the duplication is real but has not yet bitten.** Measured on `0a0576c9a`, excluding `submodules/`, `bin/`, `obj/` and commented-out references:

- 32 distinct packages referenced.
- **12 of them are declared in more than one file.**
- **0 currently disagree.** This is a prevention ticket for the in-repo half, not a bug report — the versions are consistent today and the work should not be justified as fixing in-repo drift that does not exist.

| Package | Version | Declared in |
| --- | --- | --- |
| `Hl7.Fhir.R4` | `$(FirelyNetVersion)` | 8 files |
| `Microsoft.Extensions.Logging.Console` | `10.0.3` | 5 files |
| `Microsoft.Extensions.Logging.Debug` | `10.0.3` | 4 files |
| `Fhir.Metrics` | `1.4.0` | 3 files |
| `MSTest.TestAdapter` | `3.8.3` | 3 files |
| `MSTest.TestFramework` | `3.8.3` | 3 files |
| `Microsoft.Extensions.Configuration` | `10.0.3` | 3 files |
| `Microsoft.NET.Test.Sdk` | `17.13.0` | 3 files |
| `coverlet.collector` | `6.0.2` | 3 files |
| `BenchmarkDotNet` | `0.14.0` | 2 files |
| `FluentAssertions` | `7.2.0` | 2 files |
| `Microsoft.CodeAnalysis.CSharp` | `5.3.0` | 2 files |

Only `Hl7.Fhir.R4`/`Hl7.Fhir.Base` is protected by a property; the other 11 repeat a literal version string in two to five places, so a Dependabot PR or a manual bump that updates one site and misses another produces a silent NU1605-class inconsistency rather than a build error.

A third recurring cost: the root `README.md` "External Dependencies" table restates these versions by hand, and [CLAUDE.md](https://github.com/FirelyTeam/firely-cql-sdk/blob/0a0576c9a679895d79fb94917522481eaed224c4/CLAUDE.md) has a dedicated rule reminding contributors to update it after every bump, precisely because it silently drifts. With one authoritative file, that table becomes checkable against a single source.

## Observed Behavior

- No `Directory.Packages.props` exists; `ManagePackageVersionsCentrally` is unset, so NuGet resolves per-project `Version` attributes.
- Cross-repo alignment with Vonk on two packages is asserted in a comment and verified only when a human remembers to look.
- The `FirelyNetVersion` ↔ Vonk `FhirNetApiVersion` invariant is broken right now and no build, test or CI job reports it.

## Expected Behavior

- A root `Directory.Packages.props` is the single source of truth for every package version in the repository.
- Every `.csproj` references packages without a `Version` attribute.
- Package families that must move together (the `Microsoft.Extensions.*` set, the MSTest pair, the Firely .NET SDK packages) are driven by named version properties, as Vonk does, so one edit moves the family.
- The structure is recognisable to anyone who works in both repositories.

## Out of scope

- **Changing any package version.** This is a pure restructuring: the resolved version of every package must be byte-for-byte what it is today. A version change hiding inside this diff would be invisible.
- **Resolving the current `6.5.0` vs `6.3.0` divergence with Vonk.** That is a Vonk-side decision, tracked separately. This ticket makes the divergence *visible and checkable*; it does not pick a version.
- Adopting central management inside the private `Firely.Cql.Sdk.Integration.Runner` submodule.

## Acceptance Criteria

- [x] Root `Directory.Packages.props` added with `<ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>` and one `PackageVersion` entry per package.
- [x] Version variables introduced for package families that move together, following Vonk's `PropertyGroup` convention, with a comment on each cross-repo-synced property naming the counterpart it must match.
- [x] Every `PackageReference` in every `.csproj` and `.props` loses its `Version` attribute; no literal version string for a NuGet package remains outside `Directory.Packages.props`.
- [x] `FirelyNetVersion` is declared once. The duplicate declaration in `Demo/cql-demo.props` is removed, and the `cql-base.props` comment is updated to point at the new single location.
- [x] **Resolved versions are provably unchanged.** Capture the restore graph before and after (e.g. compare the generated `project.assets.json` files, or `dotnet list package --include-transitive` output) and attach the diff — it must show no version change for any package on any target framework. This is the criterion that makes the PR reviewable.
- [x] `Cql-Sdk.slnf` builds clean in **`Release`** (the PublicAPI analyzer's RS0016/RS0017 are suppressed in `Debug`) on both `net8.0` and `net10.0`, and `CoreTests` and `CqlToElmTests` pass on both.
- [x] The `Demo/` projects and `Examples/` projects build, and the commented-out `Hl7.Cql.Invocation` references in `Examples/CqlSdkExamples*` are resolved. **Decision: the example projects stay under central package management**, and the alternatives keep their place as a documented menu but lose their `Version` attributes (which would now be NU1008 if uncommented) along with the two stale pins, `2.1.0-alpha.18` and a June 2025 local build stamp. A short note in each project says where the version comes from and that a consumer may instead put `Version="..."` on the reference.
- [x] Dependabot still functions against the new layout — it supports CPM, but the `.github/dependabot.yml` configuration should be confirmed rather than assumed, since #1610 recently touched it.
- [x] Root `README.md` "External Dependencies" table verified against `Directory.Packages.props` and corrected if it has drifted.
- [x] `CLAUDE.md` and `.github/copilot-instructions/` updated so the dependency-bump rule points at `Directory.Packages.props`. **Done inline in this PR at the maintainer's explicit request, rather than dispatched as its own issue for `@copilot` as this ticket originally specified** - recorded here so the departure from that convention is visible. Also covered in the same pass: the `cut-release-notes` skill, and the [Creating Tags and Releases](https://github.com/FirelyTeam/firely-cql-sdk/wiki/Creating-Tags-and-Releases) wiki page, which additionally told the reader to copy Vonk's `FhirNetApiVersion` into ours - a step that would have lowered our version.
- [x] A note in `docs/releases/vnext/` **only if** consumer-visible behaviour changes. A pure build-file restructuring with identical resolved versions changes nothing a consumer adopts, so the expected outcome is **no fragment**; record that conclusion explicitly rather than leaving it unaddressed.

## Notes for whoever picks this up

- Vonk's file is the reference for shape and naming: [`Directory.Packages.props`](https://github.com/FirelyTeam/Vonk/blob/develop/Directory.Packages.props). Note that it keeps version *variables* in the same file as the `PackageVersion` items, with a comment explaining why — `Directory.Packages.props` is imported before the project body, so the items can reference the properties. Our `cql-base.props` cannot serve that role for the same reason.
- Vonk also pins `FirelyCqlVersion` (`2.14.0` at time of writing), i.e. they consume us. Worth keeping in mind that the sync obligation runs in both directions.

baseTwo:
Fixes #1614.

> **Was stacked on #1613, now standalone.** #1613 has merged, this branch has been rebased onto `develop`, and the base is retargeted — so the diff shown is the central-package-management work against `develop`, and there is no longer an ordering constraint.

## Primary Work

Adds a root `Directory.Packages.props` as the single source of truth for every package version, mirroring the arrangement [Vonk uses](https://github.com/FirelyTeam/Vonk/blob/develop/Directory.Packages.props): version variables for families that move together, one `PackageVersion` per package, and bare `PackageReference` items in every project.

- **32 packages** centralised.
- **61 `Version` attributes** removed across **22** project files.
- **`FirelyNetVersion` reduced from two declarations to one**, in the file where versions now live. Its Vonk sync obligation is documented there, stated as the *floor at adoption* it actually is rather than as continuous equality: `Hl7.Cql.Fhir` declares `Hl7.Fhir.Base >= this version`, and Vonk pins `Hl7.Fhir.Base` directly, so Vonk must be at or above us by the time it upgrades the SDK. The SDK legitimately leading between releases is expected, and the comment now says so.

### The verification that matters

The ticket's central requirement is that **no resolved version changes** — a restructuring this mechanical could otherwise hide a version bump in 26 files of churn.

I captured the full restore graph before and after, across **all 29 projects** (not just `Cql-Sdk.slnf` — `Demo/`, `Examples/` and `tools/` are all affected by a root file):

```
baseline: assets=29 combos=46 rows=1792
after:    assets=29 combos=46 rows=1792
diff:     IDENTICAL
```

**1792 resolved package rows over 46 project/target-framework combinations, zero differences.** The capture script reads every `obj/project.assets.json` and records each resolved `name/version` per project per TFM, keyed on the full project path (keying on directory name alone silently collides `Cql/Cql` with `Demo/Cql`).

---

## Auxiliary Work

Both of the following are required to make CPM work here. Neither is incidental cleanup, and I'd rather they were seen than buried.

### Restore was machine-dependent, and CPM exposed it

The first restore sweep failed with **NU1507**: *"There are 2 package sources defined ... nuget.org, Github"*. Our [`NuGet.config`](../blob/develop/NuGet.config) declares **only** nuget.org — the second source came from machine-level configuration. So package resolution in this repository has been depending on whose machine the restore runs on, and CPM is simply the first thing to complain about it.

Fixed by adding `<clear />` (drops sources inherited from user/machine config) plus a `packageSourceMapping` of `*` → `nuget.org`.

**This is the one change in this PR with effects beyond the restructuring, so it deserves an explicit decision:** a developer who relies on a personal private feed to restore this repository would now be cut off. Every package here is public and comes from nuget.org, so I believe this is correct and makes builds reproducible — but say the word and I'll split it out.

### The private submodule would have broken

MSBuild finds the *nearest* `Directory.Packages.props` by walking up from each project. `submodules/Firely.Cql.Sdk.Integration.Runner` has none, so its **7 projects** would have walked up to our new root file, inherited `ManagePackageVersionsCentrally=true`, and failed restore with **NU1008** — they declare versions on their own `PackageReference` items. `Cql-Sdk.slnf` excludes them, so this would not have shown up until someone built `Cql-Sdk-All.sln` or ran the integration runner.

Fixed with an opt-out at `submodules/Directory.Packages.props` setting the property back to `false`. That path is in *this* repository, not inside the submodule, so the boundary is untouched and adopting CPM there stays the submodule's own decision.

Verified by evaluation rather than by assumption:

```
submodule IntegrationRunner.csproj  ManagePackageVersionsCentrally = false
our       Cql.Firely.csproj         ManagePackageVersionsCentrally = true
```

### Instruction files and the wiki updated in this PR, by maintainer override

`CLAUDE.md` still said package versions live in the `.props` files, which this PR makes false, so leaving it would have shipped instructions that contradict the repository.

The repository convention is that instruction-file changes are filed as their own issue and applied by `@copilot` rather than edited in a working copy. **The maintainer explicitly overrode that convention for this change**, so the updates are in this PR. The convention itself is unchanged and this is a one-off exception, recorded here and on #1614 so it does not read as an oversight or set a precedent.

What changed:

- `CLAUDE.md` — the dependency-bump section now points at `Directory.Packages.props`, notes that a `Version` attribute on a `PackageReference` is an error (`NU1008`), and names the two deliberate exclusions.
- `.github/copilot-instructions/04-development-guidelines.md` 4.5.2.1 — the same rule, per the sync requirement between the two instruction sets.
- `.github/copilot-instructions.md` — bumped to 3.18.0 with a version-history entry. The 3.17.0 entry describing the old props comments is deliberately left alone: it is a historical record, not current guidance.
- `.claude/skills/cut-release-notes/SKILL.md` — reads `FirelyNetVersion` from its new home, and its Vonk step becomes a **record** step rather than a reconcile step.
- The [Creating Tags and Releases](https://github.com/FirelyTeam/firely-cql-sdk/wiki/Creating-Tags-and-Releases) wiki page (wiki commit `02d25bc`, outside this diff). Besides the moved location it told the reader to copy Vonk's `FhirNetApiVersion` into ours — a step that would have **lowered** our version and undone #1611.

### `Microsoft.Extensions.Logging.Abstractions` unpinned

`Microsoft.Extensions.Logging.Abstractions` was pinned separately at `10.0.7` while the rest of the `Microsoft.Extensions.*` family sat at `10.0.3`. It is now folded back into `$(MicrosoftExtensionsVersion)`.

The pin existed to satisfy a transitive floor from an older Firely .NET SDK release, which required `Logging.Abstractions >= 10.0.7` and turned our `10.0.3` into an `NU1605` package downgrade. The current Firely .NET SDK no longer imposes that floor, so the pin did nothing except hold three packages above the family. Verified by removing it and restoring: no `NU1605`, clean across all 29 projects.

**This is the one place this PR does not leave resolved versions untouched, and it is deliberate:**

| Package | Change | Rows |
| --- | --- | --- |
| `Microsoft.Extensions.Logging.Abstractions` | `10.0.7` → `10.0.3` | 26 |
| `Microsoft.Extensions.DependencyInjection.Abstractions` | `10.0.7` → `10.0.3` | 26 |
| `System.Diagnostics.DiagnosticSource` | `10.0.7` → `10.0.3` | 10 |

The latter two were being lifted transitively by the pin. For consumers this **lowers** a dependency floor rather than raising one, so it is not breaking: the floor the affected `Hl7.Cql.*` packages declare drops from `10.0.7` to `10.0.3`, and an application already resolving `10.0.7` or newer keeps doing so.

Recorded as a release-note fragment (`docs/releases/vnext/1615-central-package-management.md`) — the earlier "no fragment needed" reasoning rested on resolved versions being identical, which no longer holds.

## Verification

- Restore graph across all 29 projects is identical to the pre-change baseline **except** the three packages listed above — 1792 rows over 46 project/target-framework combinations, with no package added or removed. Every other version is byte-identical.
- `dotnet build Cql-Sdk.slnf -c Release` — **0 warnings, 0 errors**, both `net8.0` and `net10.0`.
- `CoreTests` — 1378 passed, 0 failed, both TFMs.
- `CqlToElmTests` — 2509 passed, 0 failed, both TFMs.
- All counts identical to the pre-change runs.
- `Demo/CLI`, `Demo/Measures.Demo`, `Demo/Test.Measures.Demo`, both `Examples/` projects and `tools/XsdToCSharpConverter` all build in Release.
- No `Version=` remains on any live (non-commented) `PackageReference` anywhere in the repository.
- Root `README.md` "External Dependencies" table checked package-by-package against the new file — all 9 entries match, no drift to correct.
- `.github/dependabot.yml` already targets `directory: "/"` and excludes `submodules/**`, so it picks up the central file; NuGet CPM is supported by Dependabot.

## Not done here


baseTwo:
@claude[agent] please fix and resolve Copilots review comments

CLAUDE.md
```diff
@@ -76,9 +76,13 @@ For FHIR primitives that checker is `Hl7.Fhir.Model.<Type>.IsValidValue(value)`
...
 
-## Dependency version bumps — update the root README table
+## Dependency version bumps — change `Directory.Packages.props`, then the root README table
 
-When any external package version is bumped in a `.csproj` or `.props` file, also update the version number in the **"External Dependencies" table in the root `README.md`** — not just the sub-project READMEs (which don't state version numbers). The root README's versioned table is the only place the canonical pinned versions are documented for consumers, and it silently drifts after bumps that only touch `csproj` files. Likewise, check `docs/` for any design or assessment doc that mentions a version number for the same package and update it too.
+External package versions are centrally managed. **Every version lives in the root [`Directory.Packages.props`](Directory.Packages.props)** — that file explains the arrangement in full and is the one place to change a version. Projects carry a bare `<PackageReference Include="Foo" />` with no `Version` attribute; adding one is an error (`NU1008`). Don't go looking in `cql-base.props` or a `.csproj` for a package version, and don't reintroduce one there.
```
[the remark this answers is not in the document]
baseTwo: Fixed — the PR description no longer contradicts the diff.

The `Not done here` bullet claiming `CLAUDE.md` and `.github/copilot-instructions/` were intentionally unchanged is removed, and Auxiliary Work now carries a section recording what actually happened: the maintainer explicitly overrode the dispatch-to-`@copilot` convention for this change, so the instruction updates are in this PR. The convention itself is unchanged and the exception is noted as one-off, here and on #1614, so it does not read as an oversight or set a precedent.

The section lists the full scope, which was wider than the two files this thread names: `CLAUDE.md`, `04-development-guidelines.md` 4.5.2.1, `copilot-instructions.md` (3.18.0 plus a history entry), the `cut-release-notes` skill, and the Creating Tags and Releases wiki page (wiki commit 02d25bc, outside this diff) — that last one additionally told the reader to copy Vonk's `FhirNetApiVersion` into ours, which would have lowered our version and undone #1611.

Also corrected while in there: the banner still said this PR was stacked on #1613 and must not be merged first. #1613 has since merged, the branch is rebased onto `develop` (`6b37d723f`) and the base retargeted, so that constraint is gone.

No code change was needed for this thread — the instruction files were already updated in the branch, the description had simply not caught up.

Directory.Packages.props
```diff
@@ -0,0 +1,99 @@
...
+		<PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" />
+		<PackageVersion Include="Microsoft.Extensions.Configuration" Version="$(MicrosoftExtensionsVersion)" />
+		<PackageVersion Include="Microsoft.Extensions.Configuration.Binder" Version="$(MicrosoftExtensionsVersion)" />
+		<PackageVersion Include="Microsoft.Extensions.Configuration.CommandLine" Version="$(MicrosoftExtensionsVersion)" />
+		<PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsVersion)" />
+		<PackageVersion Include="Microsoft.Extensions.FileSystemGlobbing" Version="$(MicrosoftExtensionsVersion)" />
+		<PackageVersion Include="Microsoft.Extensions.Hosting" Version="$(MicrosoftExtensionsVersion)" />
+		<PackageVersion Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsVersion)" />
+		<!--
+			Deliberately NOT $(MicrosoftExtensionsVersion): this package is pinned at 10.0.7 while the
```
alexzautke: @baseTwo I don’t understand why this is needed

alexzautke (changes requested):

Directory.Packages.props
baseTwo: It's been like this for some time, I'll check what broke originally and if it is still necessary to pin it outside the group

baseTwo: Good question — the answer is that it *was* needed and no longer is, so it is now removed (757650caa).

**Why it existed:** Firely .NET SDK `6.2.0` changed its transitive dependency `Microsoft.Extensions.Caching.Memory` from `10.0.3` to `10.0.7`, and `Caching.Memory 10.0.7` requires `Microsoft.Extensions.Logging.Abstractions >= 10.0.7`. Our family pin at `10.0.3` therefore became an `NU1605` package downgrade. #1281 fixed it with a single targeted pin rather than bumping the whole `Microsoft.Extensions.*` family, because a repo-wide bump produced `NU1605` in downstream and submodule builds instead — the integration-runner submodule pins that family at `10.0.3`.

**Why it no longer is:** Firely reverted that dependency in `6.3.0`. `Hl7.Fhir.Base` `6.3.0` and `6.5.0` both pull `Caching.Memory 8.0.1`, which requires only `Logging.Abstractions 8.0.2` — well below our `10.0.3`. The pin has been dead weight since we adopted `6.3.0`.

Verified rather than assumed: with the pin removed, all 29 projects restore clean with no `NU1605`, Release builds 0 warnings / 0 errors, and `CoreTests` (1378) and `CqlToElmTests` (2509) pass on both target frameworks with counts identical to the pinned build.

One consequence worth flagging, since it makes this PR no longer a pure no-op on resolved versions: the pin was transitively lifting two siblings, so three packages move.

| Package | Change |
| --- | --- |
| `Microsoft.Extensions.Logging.Abstractions` | `10.0.7` → `10.0.3` |
| `Microsoft.Extensions.DependencyInjection.Abstractions` | `10.0.7` → `10.0.3` |
| `System.Diagnostics.DiagnosticSource` | `10.0.7` → `10.0.3` |

For consumers this lowers a dependency floor rather than raising one, so it is not breaking. It is written up in the PR description and in a release-note fragment.

Also removed the long explanatory comment that sat above the pin: per the convention now recorded in `CLAUDE.md`, that backstory belongs here in the PR, not in the file.

alexzautke (approved):

PR #1617, opened 2026-09-16
#1616 Fix Interval highClosedExpression index

### LINKED ISSUE #1616: Interval: highClosedExpression populated from the lowClosed argument
`ElmFactory.Populate(Interval, Expression[])` checks `arguments[3]` to decide whether the high boundary's closed flag is a literal, but in the non-literal branch assigns `arguments[2]`, so `highClosedExpression` is populated from the low boundary's argument (`Cql/Cql.CqlToElm/ElmFactory.cs`, the `else` after the `arguments[3]` literal check).

A call such as `"Interval"(1, 10, 1 > 2, 1 < 2)` translates with both closed expressions equal to `1 > 2` and evaluates to `[2, 9]` instead of `[2, 10]`, with no error or warning.

Only the quoted-identifier call reaches this branch. The `Interval[...]` selector and every internal caller of `SystemLibrary.Interval` pass Boolean literals for those two positions, so shipped selector syntax is unaffected.

Fix is the one-character index correction plus a test that pins each closed expression to its own argument. PR to follow.

shawn-vincent-well:
## Description

`Populate(Interval, Expression[])` tested `arguments[3]` for the high boundary's closed argument but assigned `arguments[2]` in the non-literal branch, so `highClosedExpression` came from the low boundary's argument. The index is corrected.

Per review: CQL defines no `Interval` function and the reference translator rejects `"Interval"(low, high, lowClosed, highClosed)` with non-literal closed arguments, so this translator now rejects it too, as a translation error, rather than advertising the form as supported. The validator hangs off `SystemLibrary.Interval` the way `DifferenceBetween`'s does. A quoted-name call resolves through `InvocationBuilder.Invoke(IHasSignature, string?, ...)`, which did not run validators, so that overload now validates system functions; `"DifferenceBetween"(...)` and `"DurationBetween"(...)` called by quoted name were skipping theirs as well.

Scope: only the quoted identifier can supply non-literal closed arguments. The `Interval[...]` selector (`ExpressionVisitor.VisitIntervalSelector`) and every internal caller of `SystemLibrary.Interval` (`TimingExpressionVisitor`) pass `ElmFactory.Literal` for those positions, and `CoercionProvider.ChangeIntervalType` builds the node directly, outside the validated path. None of the 139 `.cql` files under `LibrarySets` and `Demo` use the quoted form.

No `GeneratorToolVersion` bump: the change is in the translator, not the generator, same ruling as #1606.

## Related issues

Closes #1616.

## Testing

`IntervalTest.Interval_NonLiteral_Closed_Arguments` translates `"Interval"(1, 10, 1 > 2, 1 < 2)` and asserts the error `The closed boundary arguments of Interval must be Boolean literals.` It fails on `develop` with `Expected errors not to be empty.`, and fails the same way with either half of the change reverted alone (the validator registration, or the `Validate` call in the quoted-name invoke path). `CqlToElmTests` (2510) and `CoreTests` (1378) pass on net8.0 and net10.0.

Release-note fragment: `docs/releases/vnext/1617-interval-high-closed-argument.md`, under Potentially Breaking.

## FirelyTeam Checklist

- [x] Title under 50 characters
- [x] Potentially breaking: the quoted call with non-literal closed arguments no longer translates


alexzautke:
## Argument against merging this as a fix

The index change is correct — `arguments[3]` is the right operand, and the investigation behind #1616 is solid. My concern is with the framing: this fixes the semantics of a construct I don't think we should accept at all, and shipping it as a behaviour fix costs us more than rejecting the construct would.

### The construct is not CQL

Three independent checks:

1. **Grammar.** `intervalSelector` is `'Interval' ('['|'(') expression ',' expression (']'|')')` — exactly two expressions, with closedness carried by the bracket characters. `Interval` is a `reservedWord`, and `referentialIdentifier` is `identifier | keywordIdentifier` where `keywordIdentifier` excludes reserved words, so an unquoted `Interval(…)` can never be a function invocation. The quoted identifier is the only way in.
2. **Appendix B** defines no `Interval(…)` operator. Per §Function Resolution, an unqualified invocation "must refer to a locally defined function, or a CQL system function" — and there is no such system function.
3. **The reference translator rejects it.** `org.cqframework:cql-to-elm-cli:5.3.0`:

```
define private QuotedInterval: "Interval"(1, 10, 1 > 2, 1 < 2)

Error:[3:32, 3:62] Could not resolve call to operator Interval
  with signature (System.Integer, System.Integer, System.Boolean, System.Boolean).
```

The two-argument form fails the same way. In `cql-to-elm`, interval construction only exists as `LibraryBuilder.createInterval(low, lowClosed: Boolean, high, highClosed: Boolean)` — Kotlin booleans — so the reference translator **cannot** produce a non-literal closed boundary from CQL at all. We resolve it only because `SystemLibrary` registers the ELM node as a callable 4-ary function.

### Why that argues against merging as-is

**1. It publishes an endorsement.** The release-note fragment tells consumers that `"Interval"(low, high, lowClosed, highClosed)` now populates `highClosedExpression` correctly. That reads as "supported, and now correct". If we later reject the form, we retract it one release after advertising it, and anyone who adopts it in between gets broken.

**2. The loud remedy is the safer one for exactly the affected population.** The fragment's own framing is that evaluation results change silently. A library using this form has been computing the wrong interval; this PR makes it silently compute a *different* interval. A translation error makes the author look at the line instead. For measure content, failing loudly beats quietly changing an answer.

**3. Rejection has a smaller blast radius than the fix.** Every internal caller passes `ElmFactory.Literal` — `ExpressionVisitor.cs:75-79`, and `TimingExpressionVisitor.cs:63` (`rhsClosed`), `:92`, `:105`, `:114`, `:372-376` — and `CoercionProvider.ChangeIntervalType` builds the node directly rather than through `Populate`. So rejecting non-literal closed arguments cannot affect anything the SDK itself generates, whereas the index fix changes results for the population that has the problem. None of the 139 `.cql` files under `LibrarySets`/`Demo` use the quoted form either.

### What is *not* in question

`lowClosedExpression`/`highClosedExpression` are a real, shipped ELM feature: `cql-to-elm`'s `LibraryBuilder.convertIntervalExpression` emits them as `Property` reads of `lowClosed`/`highClosed` for every non-literal interval type conversion, mirroring our own `CoercionProvider.ChangeIntervalType`. That ELM is in this repo (`Cql/CoreTests/Input/ELM/HL7/RR23.json`, the `dqm-content-qicore-2025` set — all `translatorVersion 5.0.0`), and our compiler handles it correctly today:

```csharp
// Cql/CoreTests/CSharp/RR23-1.0.0.g.cs:117-120
bool? l_ = ((CqlInterval<CqlDate>)g_)?.lowClosed;
bool? m_ = ((CqlInterval<CqlDate>)g_)?.highClosed;
CqlInterval<CqlDateTime> n_ = context.Operators.Interval(i_, k_, l_, m_);
```

Two distinct slots, correctly wired. Neither option here touches that path — which also means no shipped content was ever mistranslated by #1616, consistent with the green HEDIS parity run.

### Concrete proposal — same size as this PR

Keep the index fix (the node is still built before validation, so it stops being user-visible rather than being wasted), and add the rejection:

- `Cql/Cql.CqlToElm/Builtin/SystemLibrary.cs:222` — chain a validator before `.For(…)`, the way `DifferenceBetween` and `DurationBetween` already do:

```csharp
public static OverloadedFunctionDef Interval = nary<Interval>(new TypeSpecifier[] { T, T, BooleanType, BooleanType, }, 4, T.ToIntervalType())
    .ValidateWith(Validators.Validate)
    .For(T, IntegerType, LongType, DecimalType, QuantityType, DateType, DateTimeType, TimeType);
```

- `SystemLibrary.cs`, in `Validators` (around :330):

```csharp
internal static void Validate(Elm.Interval i)
{
    if (i.lowClosedExpression is not null || i.highClosedExpression is not null)
        i.AddError("The closed boundary arguments of Interval must be Boolean literals.");
}
```

`InvocationBuilder` already calls `systemFunction.Validate(expression)` on every system-function invocation, so this produces a proper translator diagnostic with a locator rather than an exception — and the selector path passes it, since those arguments are always literals.

- `Interval_NonLiteral_Closed_Arguments` becomes an error-assertion test in the style of `Interval_Quantity_Incompatible_Units`, rather than asserting the evaluated value.
- The fragment changes from "results change" to "this call is now a translation error", which reads as `## Potentially Breaking`.

### Separate, larger question — not for this PR

Whether `"Interval"(…)` should resolve **at all** is a different decision. `SystemLibrary`'s constructor reflects over every public static `IDefinitionElement` into the symbol table that serves user function resolution (`SystemLibrary.cs:22-34`), so there is no per-operator choice today: if it is a public static, it is callable by quoted name. Diffing our 161 symbols against `cql-to-elm`'s 158 registered system operators, the other names in the same position as `Interval` are `Case`, `ToList`, `MinValue`, `MaxValue`, `Descendants`, `TimeComponent`, `NotEqual`, `Between`, `ProperBetween`. That needs its own issue, an opt-out mechanism, and a deprecation story — I only verified `Interval` against the reference translator empirically; the rest is a name diff and each would need checking.


---

shawn-vincent-well:
Good point, you're right. The reference translator rejects it, so we should too. Fixed as you suggested in 101a9f87: validator on `SystemLibrary.Interval`, the test now asserts the translation error, and the fragment is under Potentially Breaking. The index correction stays since the node is built before validation.

One wrinkle: a quoted-name call resolves through `InvocationBuilder.Invoke(IHasSignature, string?, ...)`, which did not run validators (only the operator and overload paths did), so the validator on its own had no effect. That overload now validates system functions too. The same gap applied to `"DifferenceBetween"(...)` and `"DurationBetween"(...)`, which were skipping their validators when called by quoted name.

Happy to open the issue on quoted names resolving to system operators separately; I have later fixes that touch the same symbol table.

shawn-vincent-well:
Opened #1633 with the verified list: eight names, not nine, checked against the reference for every overload. Descendants is registered as fluent `descendents` and the reference accepts it; TimeComponent is not a symbol here.

alexzautke:
Closed in favour of https://github.com/FirelyTeam/firely-cql-sdk/pull/1634

PR #1621, opened 2026-09-17
Fix Combine({}) and string/decimal Equivalent per spec

### LINKED ISSUE #1619: Combine({}) returns '' and string/decimal Equivalent (~) ignore the spec's case, whitespace and precision rules
## Summary

Three string/decimal operators return results that contradict the CQL 1.5.3 specification (`spec/cql/condensed/09-b-cqlreference.md`), and each one shows up as a failure in the cqframework [cql-tests](https://github.com/cqframework/cql-tests) conformance suite when run against Firely Server's `$cql` operation:

| cql-tests case | Expression | Expected | Actual |
|---|---|---|---|
| `CqlStringOperatorsTest / Combine / CombineEmptyList` | `Combine({})` | `null` | `''` |
| `CqlComparisonOperatorsTest / Equivalent / EquivFloatTrailingZero` | `1.001 ~ 1.000` | `true` | `false` |
| `CqlComparisonOperatorsTest / Equivalent / EquivStringIgnoreCase` | `'Abel' ~ 'abel'` | `true` | `false` |

This issue covers exactly these three fixes. `ReplaceMatches` substitution escaping, date/time quantity normalisation and the interval null-boundary fixes from the same triage are deliberately out of scope and will be filed separately (interval: #1620; Combine null elements: #1622).

## Impact

- `Combine` over an empty list yields an empty string that then flows into `Length`, `Concatenate`, equality etc. as a real value instead of null.
- `~` on decimals ignores the spec's "trailing zeroes are ignored in determining precision" rule and rounds half-to-even, so equivalence of values written with different scale is wrong in both directions.
- `~` on strings is a plain ordinal equality, so any measure logic relying on case-insensitive or whitespace-insensitive equivalence (the whole point of `~` over `=` for strings) gets `false`.

All three change CQL evaluation results, so per [docs/versioning.md](https://github.com/FirelyTeam/firely-cql-sdk/blob/adc54138e81d32080da30c7ff791cff0e27373b4/docs/versioning.md) the release that ships them is a MESO bump.

## Observed Behavior

Observed in cql-tests runs against Firely Server `$cql` on 2026-07-31 and again on 2026-09-17 (Firely Server on SDK 2.14.0), and confirmed by reading the current `develop` implementation. The expectations for these three cases are identical in cql-tests `main` as of 2026-09-14.

Note for verification: cql-tests `main` also changed `EquivFloat1Float1WithPrecision` (`1.5 ~ 1.55`) and `EquivFloat1Float1WithPrecisionAndZ` (`1.50 ~ 1.55`) to expect `false`. Older cql-tests checkouts expect `true` there; the fix below follows `main` (round away from zero), so a runner on an old checkout will report those two as failing.

## Root Cause

1. **`Combine`** — [`CqlOperators.StringOperators.cs#L15-L21`](``https://github.com/FirelyTeam/firely-cql-sdk/blob/adc54138e81d32080da30c7ff791cff0e27373b4/Cql/Cql.Runtime/Operators/CqlOperators.StringOperators.cs#L15-L21``) only checks `source == null`; `string.Join` over an empty sequence returns `""`. Spec (Combine): *"If the source argument is null, or the source list is empty, the result is null."*

2. **Decimal equivalence** — [`CqlComparers.DecimalCqlComparer.cs#L26-L40`](``https://github.com/FirelyTeam/firely-cql-sdk/blob/adc54138e81d32080da30c7ff791cff0e27373b4/Cql/Cql.Runtime/Comparers/CqlComparers.DecimalCqlComparer.cs#L26-L40)`` uses `decimal.Scale` as the precision, which counts trailing zeros (`1.000` has scale 3, so no rounding happens and `1.001 ~ 1.000` is false), and uses `decimal.Round` with the default `MidpointRounding.ToEven`. Spec (Equivalent): *"For decimals, equivalent means the values are the same with the comparison done on values rounded to the precision of the least precise operand; trailing zeroes after the decimal are ignored in determining precision for equivalent comparison."*

3. **String equivalence** — [`StringCqlComparer.cs#L17-L45`](https://github.com/FirelyTeam/firely-cql-sdk/blob/adc54138e81d32080da30c7ff791cff0e27373b4/Cql/Cql.Runtime/Comparers/StringCqlComparer.cs#L17-L45) is constructed with `CqlComparerEquivalentImplementation.Equals`, so `~` routes to `EqualsValues`, which does `StringComparer.Ordinal.Equals(x.Normalize(), y.Normalize())` (registration at [`CqlComparers.cs#L40`](https://github.com/FirelyTeam/firely-cql-sdk/blob/adc54138e81d32080da30c7ff791cff0e27373b4/Cql/Cql.Runtime/Comparers/CqlComparers.cs#L40)). Spec (Equivalent): *"For string values, equivalence returns true if the strings are the same value while ignoring case and locale, and normalizing whitespace. Normalizing whitespace means that all whitespace characters are treated as equivalent, with whitespace characters as defined in the whitespace lexical category. Note that null is not equivalent to the empty string ('')."*

## Expected Behavior

- `Combine({})` and `Combine({}, ',')` return `null`; `Combine(null)` stays `null`; non-empty lists are unchanged.
- Decimal `~`: precision of an operand is its scale after stripping trailing zeros; the more precise operand is rounded to that precision with `MidpointRounding.AwayFromZero`; then compared. `1.001 ~ 1.000` → true, `1.5 ~ 1.55` → false, `1.50 ~ 1.55` → false, `1.0 ~ 1` → true, `1.0 ~ 1.0` → true. Decimal `=` is untouched.
- String `~`: case-insensitive, with every whitespace character of the CQL whitespace lexical category treated as equivalent to any other, one-for-one; `'Abel' ~ 'abel'` → true, `'a\tb' ~ 'a b'` → true, `'a  b' ~ 'a b'` → false (no run collapsing), `'' ~ null` → not equivalent. String `=`, `<`, `>` and hash-based operators (`distinct`, `union`, …) stay ordinal.

## Acceptance Criteria

- [x] `Combine` returns `null` for an empty source list, with unit tests in `Cql/CoreTests`. (#1621)
- [x] `DecimalCqlComparer.EquivalentValues` ignores trailing zeros when determining precision and rounds away from zero, with unit tests covering the five cases above. (#1621)
- [x] `StringCqlComparer` gets a dedicated equivalence implementation (case-insensitive, whitespace-normalising); equality and comparison stay ordinal, with unit tests for case, mixed whitespace, and null vs empty string. (#1621)
- [x] `CombineEmptyList` passes in `Cql/CqlToElmTests` (`XmlTest`) and no other case in that suite regresses; `EquivFloatTrailingZero` and `EquivStringIgnoreCase` do not exist in the vendored DQIC copy (it predates them) and are covered by the unit tests instead. (#1621)
- [x] No `ICqlOperators` signature change and no `GeneratorToolVersion` bump needed. (#1621)
- [x] A `docs/releases/vnext/<PR>-<slug>.md` fragment under `## Fixes` records the three behaviour changes and states that they change CQL evaluation results (MESO). (#1621)

### LINKED ISSUE #1622: Combine treats null list elements as empty strings instead of ignoring them
## Summary

`Combine` joins a null element of its source list as if it were an empty string, so `Combine({'A', null, 'B'}, ',')` yields `'A,,B'`. The CQL 1.5.3 specification (`spec/cql/condensed/09-b-cqlreference.md`, section Combine) says: *"For consistency with aggregate operator behavior, null elements in the input list are ignored."* and gives the example `Combine({ 'A', 'B', 'C', null }) // 'ABC'`.

Follow-up to #1619 / #1621, which made `Combine` return `null` for an empty source list but deliberately left non-empty behaviour byte-identical.

## Impact

Any `Combine` over a list built from nullable data (for example `Combine(Encounters E return E.status.value, ', ')` where some elements are null) produces doubled separators or a leading/trailing separator. Fixing it changes CQL evaluation results, so per [docs/versioning.md](https://github.com/FirelyTeam/firely-cql-sdk/blob/adc54138e81d32080da30c7ff791cff0e27373b4/docs/versioning.md) the release that ships it is a MESO bump.

## Root Cause

[`CqlOperators.StringOperators.cs`](https://github.com/FirelyTeam/firely-cql-sdk/blob/58983f5d8/Cql/Cql.Runtime/Operators/CqlOperators.StringOperators.cs) (`Combine`) appends every element without a null check; `StringBuilder.Append(null)` is a no-op, so a null element contributes nothing but still gets a separator on each side.

## Expected Behavior

- `Combine({'A', null, 'B'}, ',')` → `'A,B'`
- `Combine({'A', 'B', 'C', null})` → `'ABC'` (spec example)
- `Combine({null, 'A'}, '-')` → `'A'`
- `Combine({null, null})` and `Combine({null}, ',')` → `null`, because after ignoring null elements nothing remains and the spec ties the behaviour to aggregate operators, which return `null` when no non-null elements exist (compare `Sum`, `Min`, `Max`).
- `Combine({})`, `Combine(null)` → `null` (already the case since #1621).
- Non-null elements, including empty strings, are unchanged: `Combine({'A', '', 'B'}, ',')` → `'A,,B'`.

## Acceptance Criteria

- [x] `Combine` skips null elements and returns `null` when no non-null elements remain, with unit tests in `Cql/CoreTests/StringOperatorTests.cs` for every case above, including the empty-string case that must keep its separator. (#1621)
- [x] `ICqlOperators` signature unchanged, no `GeneratorToolVersion` bump. (#1621)
- [x] `Cql/CqlToElmTests` `XmlTest` has no regressions; the spec's `CombineWithNulls` example is added to `Cql/CqlToElmTests/Input/DQIC/CqlStringOperatorsTest.xml` and passes. (#1621)
- [x] Release-note fragment under `docs/releases/vnext/` (`## Fixes`) stating the evaluation-result change (MESO); folded into the #1621 fragment since both land in the same PR. (#1621)

alexzautke:
## Description

**Primary Work**

Four string/decimal operator behaviours contradicted the CQL 1.5.3 specification and fail in the cqframework cql-tests conformance suite against Firely Server `$cql`:

- Decimal equivalence (`~`): each operand's precision is now its scale after stripping trailing zeros, and the more precise operand is rounded to that precision with `MidpointRounding.AwayFromZero` before comparing (`1.001 ~ 1.000` → true, `1.5 ~ 1.55` → false). Decimal equality, ordering and hashing are unchanged. (`CqlComparers.DecimalCqlComparer.cs`)
- `Combine` returns `null` for an empty source list instead of `''`, and ignores null elements (a null contributes neither text nor a separator; an all-null list yields `null`, consistent with the aggregate operators). Empty-string elements keep their separators. (`CqlOperators.StringOperators.cs`)
- String equivalence (`~`) ignores case and treats every whitespace character of the CQL whitespace lexical category (`[ \r\n\t]`) as equivalent, one-for-one (no run collapsing). `=`, ordering and hash-based operators stay ordinal. The equivalence mode is a constructor parameter of `StringCqlComparer` defaulting to the previous behaviour, so the non-CQL uses in `BundleDataSource` (value-set code/system matching) and `CompareResourcesById` keep case-sensitive matching; only the CQL comparer registration in `CqlComparers.cs` switches to spec equivalence. (`StringCqlComparer.cs`, `CqlComparers.cs`)

No `ICqlOperators` signature changes and no `GeneratorToolVersion` bump. These fixes change CQL evaluation results, so the release that ships them is a MESO bump per `docs/versioning.md`.

**Auxiliary Work**

- Release-note fragment `docs/releases/vnext/1621-combine-null-and-equivalence-precision.md` (`## Fixes`, one bullet per fix).
- Vendored DQIC suite: `CombineEmptyList` expectation corrected from `''` to `null` (matches both the spec and cql-tests `main`); the spec's `CombineWithNulls` example added as a new case.

Follow-ups noticed but out of scope: FHIR primitive string elements (`FhirString`, `Code`, …) are compared through the FHIR value comparers, not `StringCqlComparer`, so `~` on them is still ordinal.

## Related issues

Fixes #1619
Fixes #1622

## Testing

- New unit tests: `Cql/CoreTests/StringOperatorTests.cs` (11 Combine cases) and `Cql/CoreTests/CqlComparersTests.cs` (decimal: 15 data rows plus equality-unchanged assertions; string: case, whitespace, no-collapse, null vs empty, `=` unchanged, `Distinct` unchanged), all exercised through `ICqlOperators` and/or `CqlComparers`.
- `dotnet test Cql/CoreTests/CoreTests.csproj -f net10.0` on the merged branch: 1413 passed, 8 skipped, 0 failed (before the #1622 commit; that commit's own run: StringOperatorTests 11/11).
- `dotnet test Cql/CqlToElmTests/CqlToElmTests.csproj -f net10.0 --filter XmlTest`: 1507 passed / 30 skipped / 0 failed before, 1508 passed / 30 skipped / 0 failed after (the +1 is `CombineWithNulls`); the skipped list is unchanged.

## FirelyTeam Checklist

https://claude.ai/code/session_01HrgCec45seYogQkmqRpSUi


alexzautke:
@copilot Fix the merge conflicts

ewoutkramer:
@copilot:

* On the decimal equivalence: something tells me Evan did this explicitly for HEDIS, and I seem to remember this behaviour being discussed in comittee. Can you check the new version of the CQL spec (https://cql.hl7.org/R2/01-introduction.html) to see if this behaviour is still what we need to do?
* On string equivalence: if BundleDataSource is using string equivalence on codes, but that no longer works, maybe bundledatasource should no longer use string equivalence. In other works: we should look a bit deeper into whether the code matching on a retrieve should actually use string equivalence, if its semantics for retrieve are different from the default string equivalence semantics. adding a parameter to the equivalence operator feels like we're having two kinds of string equivalence, and that cannot be right.

ewoutkramer:
@copilot - you have budget again, pick up the issue above.

ewoutkramer (approved):

PR #1642, opened 2026-09-19
cql-test prs we are already passing

richfirely:


richfirely:
@alexzautke ignore these copilot items. It's suggesting xml changes, Ill evaluate them separately for PRs in the cql-tests repo.

